PR #8, opened 2024-04-18
chore: update `annotate-snippets` to `0.11`

DaniPopes:


crates/interface/src/diagnostics/emitter/human.rs
```diff
@@ -132,91 +135,82 @@ impl HumanEmitter {
...
-impl OwnedSourceAnnotation {
-    fn as_ref(&self) -> SourceAnnotation<'_> {
-        SourceAnnotation {
-            range: self.range,
-            label: &self.label,
-            annotation_type: self.annotation_type,
-        }
+impl OwnedAnnotation {
+    fn as_ref(&self) -> Annotation<'_> {
+        self.annotation_type.span(self.range.clone()).label(&self.label)
```
DaniPopes: .level

PR #46, opened 2024-08-15
feat: reduce sizeof::<TokenKind> from 12 to 8

DaniPopes:
Closes https://github.com/paradigmxyz/sulk/issues/28

### DIFF
--- a/crates/ast/src/token.rs
+++ b/crates/ast/src/token.rs
@@ -248,7 +248,7 @@ pub enum TokenKind {
     /// A literal token.
     ///
     /// Note that this does not include boolean literals.
-    Literal(TokenLit),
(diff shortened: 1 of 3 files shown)


crates/ast/src/token.rs
```diff
@@ -248,7 +248,7 @@ pub enum TokenKind {
     /// A literal token.
     ///
     /// Note that this does not include boolean literals.
-    Literal(TokenLit),
+    Literal(TokenLitKind, Symbol),
```
DaniPopes, line 251: Bit unfortunate, would've liked to keep this but the other solution would be making TokenLit repr(packed) to remove the alignment, but that adds unnecessary unsafe code

PR #89, opened 2024-10-23
feat: implement duplicate parameter type check

DaniPopes:
part 1 of 100000 of type checking

### DIFF
--- a/crates/sema/src/ty/mod.rs
+++ b/crates/sema/src/ty/mod.rs
@@ -261,6 +261,10 @@ impl<'gcx> Gcx<'gcx> {
     /// Returns the parameter variable declarations of the given function-like item.
     ///
     /// Also accepts structs.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the item is not a function-like item or a struct.
     pub fn item_parameters(self, id: hir::ItemId) -> &'gcx [hir::VariableId] {
         self.item_parameters_opt(id)
             .unwrap_or_else(|| panic!("item_parameters: invalid item {id:?}"))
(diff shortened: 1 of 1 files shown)


crates/sema/src/ty/mod.rs
```diff
@@ -238,28 +248,44 @@ impl<'gcx> Gcx<'gcx> {
...
-            _ => return None,
-        })
+        let params = self.item_parameters(id);
+        debug_assert_eq!(tys.len(), params.len());
+        std::iter::zip(tys.iter().copied(), params.iter().copied())
+    }
+
+    /// Returns the parameter variable declarations of the given function-like item.
+    ///
+    /// Also accepts structs.
```
onbjerg, line 263: nit: document panic

crates/sema/src/ty/mod.rs
```diff
@@ -238,28 +248,44 @@ impl<'gcx> Gcx<'gcx> {
...
+    pub fn item_parameters_opt(self, id: hir::ItemId) -> Option<&'gcx [hir::VariableId]> {
+        self.hir.item(id).parameters()
+    }
+
+    /// Returns the return variable declarations of the given function-like item.
```
onbjerg, line 280: nit: document panic

crates/sema/src/typeck/mod.rs
```diff
@@ -1,22 +1,83 @@
...
+                return false;
+            }
+        }
+        if !same_external_params(gcx, gcx.type_of_item(a), gcx.type_of_item(b)) {
+            return false;
+        }
+        true
+    };
+
+    reported.clear();
```
onbjerg: no need to clear here its already empty

tests/ui/typeck/duplicate_overloaded_items.sol
```diff
@@ -0,0 +1,49 @@
+contract C {
```
onbjerg, line 50: we should add tests for top-level events and fns with duplicate names

crates/sema/src/typeck/mod.rs
```diff
@@ -1,22 +1,83 @@
...
+    for (_name, decls) in &scope.declarations {
+        let decls = &decls[..];
+        if decls.len() <= 1 {
+            continue;
+        }
+        for (i, &decl) in decls.iter().enumerate() {
```
onbjerg, line 52: we should clear in here before this loop though

PR #110, opened 2024-11-08
chore: use unimplemented! instead of todo! in eval.rs

DoTheBestToGetTheBest:


DaniPopes:
Thanks for the PR! However these match arms are not actually to-do, we want to fail regardless because they're not supported. Would you mind changing this PR to replace `todo!()` with `unimplemented!()`?

crates/sema/src/eval.rs
```diff
@@ -62,36 +62,38 @@ impl<'gcx> ConstantEvaluator<'gcx> {
...
             hir::ExprKind::Ternary(cond, t, f) => {
                 let cond = self.try_eval(cond)?;
                 Ok(if cond.to_bool() { self.try_eval(t)? } else { self.try_eval(f)? })
             }
-            // hir::ExprKind::Tuple(_) => todo!(),
-            // hir::ExprKind::TypeCall(_) => todo!(),
-            // hir::ExprKind::Type(_) => todo!(),
+            hir::ExprKind::Tuple(_) => unimplemented!(),
+            hir::ExprKind::TypeCall(_) => unimplemented!(),
+            hir::ExprKind::Type(_) => unimplemented!(),
```
DaniPopes: Please keep them commented, we don't want to panic

DoTheBestToGetTheBest: > Please keep them commented, we don't want to panic

done :)

DaniPopes (approved):

PR #111, opened 2024-11-11
AST validate check statement only in while/for loops

ferranbt:
This PR extends the `AstValidator` to check that `break` statements can only happen while in `for` or `while` statements.

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -1,6 +1,10 @@
 //! AST-related passes.
 
-use solar_ast::{ast, visit::Visit};
+use solar_ast::{
+    ast,
+    ast::{Stmt, StmtKind},
+    visit::Visit,
+};
 use solar_interface::{diagnostics::DiagCtxt, sym, Session, Span};
 
 #[instrument(name = "ast_passes", level = "debug", skip_all)]
@@ -19,17 +23,22 @@ pub fn validate(sess: &Session, ast: &ast::SourceUnit<'_>) {
 struct AstValidator<'sess> {
     span: Span,
     dcx: &'sess DiagCtxt,
+    in_loop_depth: u64,
 }
 
 impl<'sess> AstValidator<'sess> {
     fn new(sess: &'sess Session) -> Self {
-        Self { span: Span::DUMMY, dcx: &sess.dcx }
+        Self { span: Span::DUMMY, dcx: &sess.dcx, in_loop_depth: 0 }
     }
 
     /// Returns the diagnostics context.
     #[inline]
     fn dcx(&self) -> &'sess DiagCtxt {
         self.dcx
+    }
+
+    fn in_loop(&self) -> bool {
+        self.in_loop_depth != 0
     }
 }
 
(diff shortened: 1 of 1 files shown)


tests/ui/parser/break_outside_loop.sol
```diff
@@ -0,0 +1,3 @@
+function f() {
+    break; //~^ERROR: `break` outside of a loop
```
DaniPopes:
```suggestion
    break; //~ ERROR: `break` outside of a loop

    { break; } //~ ERROR: `break` outside of a loop
```

Please also add the same function inside of a `contract C`

tests/ui/parser/break_outside_loop.sol
```diff
@@ -0,0 +1,3 @@
...
+}
```
DaniPopes: should also test that it doesn't error when inside of loops and nested loops

crates/sema/src/ast_passes.rs
```diff
@@ -69,11 +78,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+                self.in_loop_depth += 1;
+                self.visit_stmt(stmt);
+                self.in_loop_depth -= 1;
+            }
+            StmtKind::For { body, .. } => {
+                self.in_loop_depth += 1;
+                self.visit_stmt(body);
+                self.in_loop_depth -= 1;
+            }
+            StmtKind::Break => {
```
DaniPopes, line 92: same for `continue`

crates/sema/src/ast_passes.rs
```diff
@@ -69,11 +78,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+                if !self.in_loop() {
+                    self.dcx().err("`break` outside of a loop").span(self.span).emit();
```
```suggestion
                    self.dcx().err("`break` outside of a loop").span(stmt.span).emit();
```

crates/sema/src/ast_passes.rs
```diff
@@ -69,11 +78,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+            StmtKind::DoWhile(stmt, ..) => {
+                self.in_loop_depth += 1;
+                self.visit_stmt(stmt);
+                self.in_loop_depth -= 1;
+            }
+            StmtKind::For { body, .. } => {
+                self.in_loop_depth += 1;
+                self.visit_stmt(body);
+                self.in_loop_depth -= 1;
+            }
```
DaniPopes: you can check all of these in one branch by doing `While(_, block, ..) | DoWhile(block, ..) | For { body, .. } => {}`

crates/sema/src/ast_passes.rs
```diff
@@ -69,11 +78,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
         }
     }
 
+    fn visit_stmt(&mut self, stmt: &'ast ast::Stmt<'ast>) {
+        let Stmt { kind, .. } = stmt;
+
+        match kind {
+            StmtKind::DoWhile(stmt, ..) => {
+                self.in_loop_depth += 1;
+                self.visit_stmt(stmt);
```
DaniPopes: these should use `self.walk_stmt` instead of `self.visit_stmt` to recurse on all statements correctly

DaniPopes (changes requested):

tests/ui/parser/break_outside_loop.sol
ferranbt: Check the update smart contract and let me know if that looks okay.

DaniPopes (approved):
LGTM, thanks!

PR #112, opened 2024-11-11
Add syntax test pragma solidity

ferranbt:
This PR adds a syntax unit test to check the case where the pragma has a version but not the `solidity` pragma.


DaniPopes (approved):
Thanks!

PR #115, opened 2024-11-12
feat: return ControlFlow from AST visitor methods

### LINKED ISSUE #14: Return `ControlFlow` from AST visitor methods
Currently AST visitors do not return anything, whereas it would be beneficial to return a `ControlFlow` value, with the associated break type defined by implementors. By default, all functions should always return `Continue(())`, and implementors may choose to break early.

https://github.com/paradigmxyz/sulk/blob/5f827dd5a0fed12f8e57c48ac60e371b98d921a3/crates/ast/src/visit.rs#L7-L10

mahmudsudo:
closes #14

### DIFF
--- a/crates/ast/src/visit.rs
+++ b/crates/ast/src/visit.rs
@@ -9,19 +9,22 @@ use std::ops::ControlFlow;
 declare_visitors! {
     /// AST traversal.
     pub trait Visit VisitMut <'ast> {
-        fn visit_source_unit(&mut self, source_unit: &#mut SourceUnit<'ast>) -> ControlFlow<()> {
+        /// The value returned when breaking from the traversal
+        type BreakValue;
+
+        fn visit_source_unit(&mut self, source_unit: &'ast #mut SourceUnit<'ast>) -> ControlFlow<Self::BreakValue> {
             // TODO: SAFETY: Idk
             let source_unit = unsafe { trustme::decouple_lt #_mut(source_unit) };
             let SourceUnit { items } = source_unit;
             for item in items.iter #_mut() {
-                if let ControlFlow::Break(_) = self.visit_item #_mut(item) {
-                    return ControlFlow::Break(());
+                if let ControlFlow::Break(val) = self.visit_item #_mut(item) {
+                    return ControlFlow::Break(val);
                 }
             }
             ControlFlow::Continue(())
         }
 
-        fn visit_item(&mut self, item: &'ast #mut Item<'ast>) -> ControlFlow<()> {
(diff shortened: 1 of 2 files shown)


crates/ast/src/visit.rs
```diff
@@ -4,139 +4,154 @@ use crate::ast::*;
...
-        fn visit_source_unit(&mut self, source_unit: &#mut SourceUnit<'ast>) {
+        fn visit_source_unit(&mut self, source_unit: &#mut SourceUnit<'ast>) -> ControlFlow<()> {
             // TODO: SAFETY: Idk
             let source_unit = unsafe { trustme::decouple_lt #_mut(source_unit) };
             let SourceUnit { items } = source_unit;
             for item in items.iter #_mut() {
-                self.visit_item #_mut(item);
+                if let ControlFlow::Break(_) = self.visit_item #_mut(item) {
+                    return ControlFlow::Break(());
+                }
```
DaniPopes: This can be `?`

crates/sema/src/ast_passes.rs
```diff
@@ -107,11 +109,14 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
             }
             _ => {}
         }
+        ControlFlow::Continue(())
     }
 
-    // Intentionally override unused default implementations to reduce bloat.
```
DaniPopes, line 112: Please keep this comment

crates/ast/src/visit.rs
```diff
@@ -4,139 +4,154 @@ use crate::ast::*;
 use solar_data_structures::trustme;
 use solar_interface::{Ident, Span};
 use solar_macros::declare_visitors;
+use std::ops::ControlFlow;
 
 declare_visitors! {
     /// AST traversal.
     pub trait Visit VisitMut <'ast> {
-        fn visit_source_unit(&mut self, source_unit: &#mut SourceUnit<'ast>) {
+        fn visit_source_unit(&mut self, source_unit: &#mut SourceUnit<'ast>) -> ControlFlow<()> {
```
DaniPopes: We should let implementors define the `Break` type; it should be possible to add `type BreakValue;` in the trait and these return types to be `ControlFlow<Self::BreakValue>`.

One can specify `Never` as the `BreakValue`, this should be in the docs for the type

DaniPopes (changes requested):

crates/ast/src/visit.rs
mahmudsudo: putting a ``?`` would prompt an error in the if let statement

DaniPopes (approved):
Thanks!

PR #116, opened 2024-11-13
AST validate unchecked nested blocks

ferranbt:
This PR adds an AST validation to check for nested `unchecked` blocks.

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -93,7 +93,7 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
             StmtKind::Break => {
                 if !self.in_loop() {
                     self.dcx()
-                        .err("\"break\" has to be in a \"for\" or \"while\" loop.")
+                        .err("'break' has to be in a 'for' or 'while' loop")
                         .span(stmt.span)
                         .emit();
                 }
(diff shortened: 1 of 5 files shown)


crates/sema/src/ast_passes.rs
```diff
@@ -105,6 +106,15 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
                 }
             }
+            StmtKind::UncheckedBlock(block) => {
+                if self.in_unchecked_block {
+                    self.dcx().err("\"unchecked\" blocks cannot be nested.").span(stmt.span).emit();
+                }
+
+                self.in_unchecked_block = true;
+                self.walk_block(block);
+                self.in_unchecked_block = false;
```
```suggestion
                let prev = self.in_unchecked_block;
                self.in_unchecked_block = true;
                self.walk_block(block);
                self.in_unchecked_block = prev;
```

tests/ui/typeck/unchecked_block.sol
```diff
@@ -0,0 +1,9 @@
+contract C {
+    function a() public {
+        unchecked {}
+
+        unchecked {
+            unchecked {} //~ ERROR: "unchecked" blocks cannot be nested.
+        }
```
DaniPopes: Let's add one more to test the previous suggestion

```suggestion
            unchecked {} //~ ERROR: "unchecked" blocks cannot be nested.
            unchecked {} //~ ERROR: "unchecked" blocks cannot be nested.
        }
```

DaniPopes (commented):
Thanks! One nit

crates/sema/src/ast_passes.rs
```diff
@@ -105,6 +106,15 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
                         .emit();
                 }
             }
+            StmtKind::UncheckedBlock(block) => {
+                if self.in_unchecked_block {
+                    self.dcx().err("\"unchecked\" blocks cannot be nested.").span(stmt.span).emit();
```
DaniPopes: we should use ` (backticks) for keywords and code in messages

ferranbt: Two things:
- I am using the [exact](https://github.com/ethereum/solidity/blob/b0d991bd4b3bc2d48d40029c3303f67f604d7cc5/libsolidity/analysis/SyntaxChecker.cpp#L237) error message from Solidity in case we have to keep compatibility.
- I have to change the error message from the other 'break' PR too (I did it yesterday). Do you want me to change it in this PR too or as a followup PR?

DaniPopes: There's no need for compatibility in this case, and yes please also change that one too in this PR

also remove the `.` at the end, I prefer the rustc style of error message than the solc ones

ferranbt: Done

DaniPopes (approved):

PR #117, opened 2024-11-13
AST validate that a contract does not have a function with contract name

ferranbt:
This PR adds an AST validation check similar to [this](https://github.com/ethereum/solidity/blob/df01dcc7b759f53ba45811979d3286562eab7846/libsolidity/analysis/SyntaxChecker.cpp#L385) which verifies that a function does not have the same name as the contract.

Note, eventually this can be merged into a specific AST contract level checker.

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -120,10 +120,8 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
                 if let Some(func_name) = header.name {
                     if func_name.as_str() == contract_name {
                         self.dcx()
-                            .err(
-                                "Functions are not allowed to have the same name as the contract.
-        If you intend this to be a constructor, use \"constructor(...) { ... }\" to define it.",
-                            )
+                            .err("functions are not allowed to have the same name as the contract")
+                            .note("if you intend this to be a constructor, use \"constructor(...) { ... }\" to define it")
                             .span(func_name.span)
                             .emit();
                     }
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -123,7 +123,10 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
         self.walk_stmt(stmt)
     }
 
-    fn visit_item_contract(&mut self, contract: &'ast ast::ItemContract<'ast>) {
+    fn visit_item_contract(
+        &mut self,
+        contract: &'ast ast::ItemContract<'ast>,
+    ) -> ControlFlow<Self::BreakValue> {
         let ast::ItemContract { kind: _, name, bases: _, body } = contract;
         let contract_name = name.as_str();
 
@@ -143,7 +146,7 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
             }
         }
 
-        self.walk_item_contract(contract);
+        self.walk_item_contract(contract)
     }
 
     // Intentionally override unused default implementations to reduce bloat.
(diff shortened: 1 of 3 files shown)


crates/sema/src/ast_passes.rs
```diff
@@ -109,6 +109,31 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+        for item in body.iter() {
+            if let ast::ItemKind::Function(ast::ItemFunction { kind: _, header, body: _ }) =
+                &item.kind
+            {
+                if let Some(func_name) = header.name {
+                    if func_name.as_str() == contract_name {
+                        self.dcx()
+                            .err(
+                                "Functions are not allowed to have the same name as the contract.
+        If you intend this to be a constructor, use \"constructor(...) { ... }\" to define it.",
```
onbjerg: we should lowercase the start of the error and move the "if you intend..." into a `note(..)` call instead

crates/sema/src/ast_passes.rs
```diff
@@ -123,6 +123,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+
+        for item in body.iter() {
+            if let ast::ItemKind::Function(ast::ItemFunction { kind: _, header, body: _ }) =
+                &item.kind
+            {
+                if let Some(func_name) = header.name {
+                    if func_name.as_str() == contract_name {
+                        self.dcx()
+                            .err("functions are not allowed to have the same name as the contract")
+                            .note("if you intend this to be a constructor, use \"constructor(...) { ... }\" to define it")
```
```suggestion
                            .note("if you intend this to be a constructor, use `constructor(...) { ... }` to define it")
```

onbjerg (approved):
lgtm

crates/sema/src/ast_passes.rs
```diff
@@ -123,6 +123,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+    ) -> ControlFlow<Self::BreakValue> {
+        let ast::ItemContract { kind: _, name, bases: _, body } = contract;
+        let contract_name = name.as_str();
+
+        for item in body.iter() {
+            if let ast::ItemKind::Function(ast::ItemFunction { kind: _, header, body: _ }) =
+                &item.kind
+            {
+                if let Some(func_name) = header.name {
+                    if func_name.as_str() == contract_name {
```
```suggestion
                    if func_name == contract.name {
```

Compare Symbols not strings, it has the same result but faster, which is the point of Symbol in the first place

crates/sema/src/ast_passes.rs
```diff
@@ -123,6 +123,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+        contract: &'ast ast::ItemContract<'ast>,
+    ) -> ControlFlow<Self::BreakValue> {
+        let ast::ItemContract { kind: _, name, bases: _, body } = contract;
+        let contract_name = name.as_str();
+
+        for item in body.iter() {
+            if let ast::ItemKind::Function(ast::ItemFunction { kind: _, header, body: _ }) =
+                &item.kind
+            {
+                if let Some(func_name) = header.name {
```
DaniPopes: we need to also check that kind is Function

crates/sema/src/ast_passes.rs
```diff
@@ -123,6 +123,32 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+    fn visit_item_contract(
+        &mut self,
+        contract: &'ast ast::ItemContract<'ast>,
+    ) -> ControlFlow<Self::BreakValue> {
+        let ast::ItemContract { kind: _, name: _, bases: _, body } = contract;
+
+        for item in body.iter() {
+            if let ast::ItemKind::Function(ast::ItemFunction { kind, header, body: _ }) = &item.kind
+            {
+                if *kind == ast::FunctionKind::Function {
```
ferranbt, line 135: Unsure if there is a more idiomatic way to chain this if statements.

DaniPopes: There will be when let-chains are stabilized :)

PR #119, opened 2024-11-13
Initial AST validation for using-for

ferranbt:
This PR adds the initial AST validation for the `using` statement. This PR only has tests for the 3 parsing scenarios while the Solidity repo has [dozens of them](https://github.com/ethereum/solidity/tree/b0d991bd4b3bc2d48d40029c3303f67f604d7cc5/test/libsolidity/syntaxTests/using). I was wondering how do you want to approach that on `solar`. Should I create sub-directories like they do? Merge them on the same file? Should I copy/paste the whole syntax e2e tests from Solidity? Is it enough with basic tests and we do not need as many as Solidity has?

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -24,11 +24,12 @@ struct AstValidator<'sess> {
     span: Span,
     dcx: &'sess DiagCtxt,
     in_loop_depth: u64,
+    contract_kind: Option<ast::ContractKind>,
 }
 
 impl<'sess> AstValidator<'sess> {
     fn new(sess: &'sess Session) -> Self {
-        Self { span: Span::DUMMY, dcx: &sess.dcx, in_loop_depth: 0 }
+        Self { span: Span::DUMMY, dcx: &sess.dcx, in_loop_depth: 0, contract_kind: None }
     }
 
     /// Returns the diagnostics context.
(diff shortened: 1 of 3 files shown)


crates/sema/src/ast_passes.rs
```diff
@@ -109,6 +110,40 @@ impl<'ast> Visit<'ast> for AstValidator<'_> {
...
+        self.walk_item_contract(contract);
+        self.contract_kind = None;
+    }
+
+    fn visit_using_directive(&mut self, using: &'ast ast::UsingDirective<'ast>) {
+        let ast::UsingDirective { list: _, ty, global } = using;
+        let with_typ = ty.is_some();
+        if self.contract_kind.is_none() && !with_typ {
+            self.dcx()
+                .err("The type has to be specified explicitly at file level (cannot use '*')")
```
DaniPopes: please use `` ` `` backticks and don't capitalize the start of the message

DaniPopes (commented):
As long as there is at least one test per error message it's fine; if you want to add more syntax tests in the same file, otherwise the other ones in that directory also include type checking which we'll add as we implement them

ferranbt:
> otherwise the other ones in that directory also include type checking which we'll add as we implement them

Would you include those in the same file as well?

DaniPopes:
sure yeah, but you might need to split it into multiple files if a parsing error occurs because it won't continue checking the next lines

ferranbt:
> sure yeah, but you might need to split it into multiple files if a parsing error occurs because it won't continue checking the next lines

Ok, will keep in mind for next time. This PR is ready for review again.

DaniPopes (approved):
LGTM, needs rebase

Issue #29, opened 2024-07-06
Implement AST validation


DaniPopes:
We must verify correctness of the parsed AST.

A few examples:
- [x] enums must have at least 1 variant and no more than 256
- [x] `continue` and `break` are only allowed inside of a loop
- [x] `unchecked {}` blocks can only be used inside of regular blocks
- [x] placeholder statements (`_;`) are only allowed in modifiers
- [x] variable declaration statements are not allowed as the body of a loop (for, while, do while), meaning they must be inside of a block

Solc is more strict in the parser, so some of these checks can be found in the parser, while others are found in visitors, like [`SyntaxChecker`](https://github.com/ethereum/solidity/blob/df01dcc7b759f53ba45811979d3286562eab7846/libsolidity/analysis/SyntaxChecker.cpp) and [`ContractLevelChecker`](https://github.com/ethereum/solidity/blob/d42f92bd68d76db01b0dc17477cd4c7716059d93/libsolidity/analysis/ContractLevelChecker.cpp#L84-L102).

0xalpharush:
To implement "placeholder statements (_;) are only allowed in modifiers" and maybe also https://github.com/paradigmxyz/solar/pull/116, it'd be nice to have something like [Rowan's](https://github.com/rust-analyzer/rowan) red-green trees to get the nearest enclosing lexical scope. For example, one can see whether a given AST node is in an assembly block, block scope `{}`, function, contract, or file by recursively looking up its parent.

DaniPopes:
`SyntaxChecker` has been ported over (except for the parts we don't care about, mainly pragmas at the moment).

Thanks everyone who contributed!

PR #120, opened 2024-11-13
validate num. variants in enum declaration

0xalpharush:
implements "enums must have at least 1 variant and no more than 256" from https://github.com/paradigmxyz/solar/issues/29


0xalpharush:
The failure is due to not failing at parsing and during AST validation. It's odd as https://github.com/paradigmxyz/solar/pull/121 is passing and solc succeeds parsing that case. Will need to decide whether to allow error resilient parsing or try to maintain parity with solc. IMO keeping all of this in the semantic stage makes sense for IDE, formatting, etc

DaniPopes (approved):
Thanks!

Issue #118, opened 2024-11-13
Struct with empty fields fails with incorrect validation


ferranbt:
The next Solidity snippet should work because the AST validation does not check yet whether the struct is empty or not.
```
contract test {
    struct A {}
}
```

However, right now it fails with this error:
```
+++ <stderr output>
+error: expected one of `function`, `mapping`, elementary type name, or path, found `}`
+  --> ROOT/tests/ui/parser/empty_struct.sol:LL:CC
+   |
+LL |     struct A {}
+   |               ^ expected one of `function`, `mapping`, elementary type name, or path
+   |
+
+error: aborting due to 1 previous error
```

PS: I did not see any template for bug reports.

PR #121, opened 2024-11-13
better error for struct without any fields

0xalpharush:
closes https://github.com/paradigmxyz/solar/issues/118

### DIFF
--- a/crates/parse/src/parser/mod.rs
+++ b/crates/parse/src/parser/mod.rs
@@ -697,7 +697,8 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
         }
 
         if let Some(sep_kind) = &sep.sep {
-            if sep.trailing_sep_required && !trailing {
+            let open_close_delim = first && allow_empty;
+            if !open_close_delim && sep.trailing_sep_required && !trailing {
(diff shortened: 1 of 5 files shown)


crates/parse/src/parser/mod.rs
```diff
@@ -697,7 +697,8 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
         }
 
         if let Some(sep_kind) = &sep.sep {
-            if sep.trailing_sep_required && !trailing {
+            let open_close_delim = first && allow_empty;
```
0xalpharush, line 700: This just means we saw a pattern like `{}` and allows better error messages. Otherwise, the missing trailing comma distracts from the missing field error

tests/ui/typeck/empty_struct.sol
```diff
@@ -0,0 +1 @@
+struct A {} //~ERROR: structs must have at least one field
```
DaniPopes:
```suggestion
struct A {} //~ ERROR: structs must have at least one field

contract C {
    struct A {} //~ ERROR: structs must have at least one field
}

```

DaniPopes (approved):

PR #123, opened 2024-11-14
fix: install rayon pool in Session::enter

DaniPopes:
Install the rayon pool together with session globals in Session::enter instead of in CLI only; this enables library users to use `sema` and any other `rayon` usage

cc @ferranbt https://t.me/paradigm_solar/100


ferranbt:
I can confirm it works.

PR #125, opened 2024-11-15
feat: print AST statistics with -Zast-stats flag

### LINKED ISSUE #25: Print AST statistics
### Describe the feature

Add a `-Z ast-stats` flag that prints the size of the AST and of each node for each input file to stderr.

### Additional context

See rustc:
- implementation: <https://github.com/rust-lang/rust/blob/8a8ad3433e4168f54ac1364f85da694be9eeca7c/compiler/rustc_passes/src/hir_stats.rs#L81-L88>
- output file: <https://github.com/rust-lang/rust/blob/8a8ad3433e4168f54ac1364f85da694be9eeca7c/tests/ui/stats/hir-stats.stderr>

Chriscbr:
Closes #25

For example, running `cargo run -- -Z ast-stats examples/src/Counter.sol` from the project root emits:

```
ast-stats AST STATS
ast-stats Name                Accumulated Size         Count     Item Size
ast-stats ----------------------------------------------------------------
ast-stats Block                     32 ( 2.7%)             2            16
ast-stats Type                      64 ( 5.4%)             2            32
ast-stats - Elementary                64 ( 5.4%)             2
ast-stats Stmt                     160 (13.6%)             2            80
ast-stats - Expr                     160 (13.6%)             2
ast-stats Expr                     240 (20.4%)             5            48
ast-stats - Assign                    48 ( 4.1%)             1
ast-stats - Unary                     48 ( 4.1%)             1
ast-stats - Ident                    144 (12.2%)             3
ast-stats Item                     680 (57.8%)             5           136
ast-stats - Variable                 136 (11.6%)             1
ast-stats - Contract                 136 (11.6%)             1
ast-stats - Pragma                   136 (11.6%)             1
ast-stats - Function                 272 (23.1%)             2
ast-stats ----------------------------------------------------------------
ast-stats Total                  1_176
ast-stats
```


crates/sema/src/lib.rs
```diff
@@ -58,6 +60,12 @@ pub fn parse_and_resolve(pcx: ParsingContext<'_>) -> Result<()> {
         }
     }
 
+    if sess.ast_stats {
+        for source in sources.asts() {
+            stats::print_ast_stats(source, "AST STATS", "ast-stats");
+        }
+    }
```
Chriscbr, line 67: Should I add a test for this somewhere in the test suite? If so should it be testing from the CLI perspective, or just testing the `print_ast_stats` function?

crates/sema/src/stats.rs
```diff
@@ -0,0 +1,243 @@
...
+            $(
+                $mod::$tykind::$variant { .. } => {
+                    $self.record_variant(stringify!($ty), stringify!($variant), $id, $val)
+                }
+            )*
+        }
+    };
+}
+
+impl<'ast> Visit<'ast> for StatCollector {
```
Chriscbr, line 150: I'm probably missing some cases here... I can go through this with a more fine tooth comb later this week, but wanted to just get the PR out first for initial feedback.

If I understand correctly, I think I need to recurse in the cases where an AST element is being stored within a parent node, and not in the cases where it's just a reference (i.e. not taking up additional memory).

Chriscbr (commented):
This is just a first pass at adding a form of AST stats. The approach is heavily based on that of rustc; looking for any feedback!

crates/sema/src/lib.rs
DaniPopes: Yes, you can add a tests/ui/stats/ast.sol with a simple program and just pass `//@ compile-flags: -Zast-stats`

crates/sema/src/stats.rs
```diff
@@ -0,0 +1,243 @@
...
+                prefix,
+                label,
+                to_readable_str(size),
+                percent(size, total_size),
+                to_readable_str(node.stats.count),
+                to_readable_str(node.stats.size)
+            );
+            if !node.subnodes.is_empty() {
+                // We will soon sort, so the initial order does not matter.
+                #[allow(rustc::potential_query_instability)]
```
DaniPopes: This `allow` + comment is not needed

crates/sema/src/stats.rs
DaniPopes: Yes I believe that's right

crates/sema/src/lib.rs
Chriscbr: Added.

crates/sema/src/stats.rs
Chriscbr: I did some hand tuning -- basically when an AST struct contains another visitable thing if it's a Boxed reference to that thing, then we visit it (so we can uncover the size of that heap-allocated structure), but if it's inlined (e.g. Ident, Span, etc.), I have the code skip the field since it that Ident or Span or other data structure is already going to be accounted for when `std::mem::size_of_val` is called on the parent value.

crates/sema/src/stats.rs
Chriscbr: Removed

crates/sema/src/stats.rs
```diff
@@ -0,0 +1,570 @@
...
+        &mut self,
+        doc_comment: &'ast ast::DocComment,
+    ) -> ControlFlow<Self::BreakValue> {
+        self.record("DocComment", None, doc_comment);
+        // Don't visit span field since it isn't boxed
+        ControlFlow::Continue(())
+    }
+
+    fn visit_path(&mut self, path: &'ast ast::PathSlice) -> ControlFlow<Self::BreakValue> {
+        // self.record("PathSlice", None, path); // TODO: can't pass a slice since its size is unknown at compile time
```
DaniPopes: You can make `T: ?Sized` in `record` and this will work

Chriscbr: good call! fixed

DaniPopes (approved):
LGTM, thanks!

PR #126, opened 2024-11-15
Basic fuzzing harness

jmcph4:
This PR adds a basic fuzzing harness to Solar using [`cargo-fuzz`](https://rust-fuzz.github.io/book/cargo-fuzz.html), a popular frontend to the `libfuzzer` fuzzing engine.

Specifically, this PR contains the following cases:

 - `run_compiler_args`, a (reasonably close to) end-to-end case for the entire compiler
 - `parsing_context_parse_and_resolve`, specifically targeting just parsing and semantic analysis
 - `canonicalize`, targeting path canonicalisation (low-hanging and likely uninteresting, but added for posterity)

The intent (and expectation) is that this PR will allow for easy extension and expansion of fuzzing cases as development of the compiler progresses.


jmcph4:
Would appreciate a double check of the `parsing_context_parse_and_resolve` case to make sure that this is an acceptable usage of that API (so that this fuzzing case actually hits an acceptable scenario).

fuzz/fuzz_targets/canonicalize.rs
DaniPopes, line 1: This target is redundant since the `canonicalize` function is implemented using the standard library, and it's also coming from `dunce`, not implemented here.

fuzz/fuzz_targets/parsing_context_parse_and_resolve.rs
```diff
@@ -0,0 +1,24 @@
...
+    file.write_all(data).expect("Failed to write to temporary file");
+
+    let path = file.into_temp_path();
+    let path = path.keep().expect("Failed to persist temporary file");
+
+    let sess = Session::builder().with_buffer_emitter(solar_interface::ColorChoice::Auto).build();
+
+    let _ = sess.enter(|| -> solar_interface::Result<()> {
+        let mut pcx = solar_sema::ParsingContext::new(&sess);
+        pcx.load_file(&path).unwrap();
```
DaniPopes, line 21: We can avoid the temp file by using `add_file` with a `SourceFile` created in memory, like so:
```rust
let Ok(data) = std::str::::from_utf8(data) else {
    return Ok(());
};
let Ok(file) = sess.source_map().new_dummy_source_file("fuzz".into(), data.to_string()) else {
    return;
};
pcx.add_file(file);
```

fuzz/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+solar-parse = { path = "../crates/parse" }
+solar-interface = { path = "../crates/interface" }
+solar-sema = { path = "../crates/sema" }
+solar-cli = { path = "../crates/cli" }
```
DaniPopes, line 17: Use `.workspace = true` instead for these dependencies

DaniPopes (commented):
Thanks a lot! A few comments

Issue #129, opened 2024-11-18
[ICE]: nested struct with two members as return value for an interface function


samooyo:
### Code

```Rust
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

struct SimpleStruct {
    uint256 a;
}

struct NestedTestStruct {
    SimpleStruct simpleStruct;
    SimpleStruct simpleStruct2;
}

interface TestContract {
    function nestedStructFunction() external pure returns (NestedTestStruct memory);
}
```

### Rust Version

```Shell
solar Version: 0.1.0
Commit SHA: fff9d9f8e652bbfb2405211f544047ab8a221b84
Build Timestamp: 2024-11-08T04:36:07.494153185Z
Build Features: asm,default,jemalloc
Build Profile: dist
```

### Current error output

```Shell
error: recursive types cannot be parameter or return types of public functions
  --> src/Counter.sol:14:60
   |
14 |     function nestedStructFunction() external pure returns (NestedTestStruct memory);
   |                                                            ^^^^^^^^^^^^^^^^^^^^^^^
   |

error: aborting due to 1 previous error
```

### Anything else?

The following code compiles with solc v0.8.26, but fails with Solar.

Note that removing one of the fields of `NestedTestStruct` makes the compilation succeed.

PR #133, opened 2024-11-19
fix: properly handle recursive types

DaniPopes:
Fixes https://github.com/paradigmxyz/solar/issues/129

### DIFF
--- a/crates/data-structures/src/cycle.rs
+++ b/crates/data-structures/src/cycle.rs
@@ -1,0 +1,140 @@
+use crate::map::FxHashSet;
+use std::ops::ControlFlow;
+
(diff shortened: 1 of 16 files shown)


crates/data-structures/src/cycle.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn to_controlflow(self) -> ControlFlow<Self> {
+        match self {
+            Self::Continue => ControlFlow::Continue(()),
+            _ => ControlFlow::Break(self),
+        }
+    }
+
+    /// Returns `true` if the result is `Continue`.
+    #[inline]
+    pub fn is_ok(&self) -> bool {
```
onbjerg: nit: let's name it `is_continue`?

Issue #27, opened 2024-07-06
Implement a simple AST pretty-printer


DaniPopes:
### Describe the feature

We will need a pretty-printer for use in:
- lints/code suggestions
- displaying items in auto-generated documentation
- re-use in a formatter

0x0elliot:
Hey! I would love to pick this issue up to help out. Funnily enough, I have no idea where to start.

I would love to become a frequent contributor to solar. It would be kind of you to nudge me so that I have a clear idea :)

PR #153, opened 2024-12-01
[wip] feat: AST pretty-printer

klkvr:
Closes https://github.com/paradigmxyz/solar/issues/27

Incomplete for now, opening draft for sanity review on direction.

lmk if it makes more sense to implement this as an `Visitor` impl. I've took an approach with standalone methods for now to avoid boilerplate code converting `fmt::Result`s to `ControlFlow`s


Cargo.toml
```diff
@@ -81,6 +81,7 @@ codegen-units = 1
 # compiler crates
 solar = { version = "0.1.0", path = "crates/solar", package = "solar-compiler" }
 solar-ast = { version = "0.1.0", path = "crates/ast" }
+solar-ast-pretty = { version = "0.1.0", path = "crates/ast-pretty" }
```
DaniPopes: i think we should have this in ast crate

crates/ast-pretty/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
+use core::fmt::{self, Write};
+use solar_ast as ast;
+
+/// AST pretty-printer.
+#[derive(Debug)]
+pub struct Printer<W> {
+    writer: W,
+    ident: usize,
```
DaniPopes: typo

crates/ast-pretty/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl<W> Printer<W> {
+    pub fn new(writer: W) -> Self {
+        Self { writer, ident: 0 }
+    }
+}
+
+impl<W: Write> Printer<W> {
+    fn write_indent(&mut self) -> fmt::Result {
```
DaniPopes: should be called print_

DaniPopes (commented):
approach is fine, we can also make individual methods public, eg for expr

klkvr:
@DaniPopes this should be complete, though no tests yet — unsure if we need those and what'd be the best approach to write them

should we add some helpers to check whether two ASTs are identical?

TilakMaddy:
I was thinking if there's a method on ast nodes to "peek" into the relevant portion of the source code that would be great too

Bisht13:
Is this PR pending because of lack of tests? Wanted to help get this PR cross the line

DaniPopes:
I'm currently in the process of rewriting forge fmt using solar, I believe we will be able to upstream that implementation so closing for now

wtdcode:
> I'm currently in the process of rewriting forge fmt using solar, I believe we will be able to upstream that implementation so closing for now

Hi! Any progress for the new implementation? Or is it open-sourced?

DaniPopes:
@wtdcode work is being done here
- https://github.com/foundry-rs/foundry/pull/10350
- https://github.com/foundry-rs/foundry/pull/10907

PR #158, opened 2024-12-08
feat: variable declaration statements are not allowed as the body of loop

TilakMaddy:
[AST Validation improvement](https://github.com/paradigmxyz/solar/issues/29)


crates/sema/src/ast_passes.rs
```diff
@@ -119,14 +121,39 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
+                if let Some(init) = init {
+                    self.x_depth = 0;
+                    self.visit_stmt(init)?;
+                }
+                if let Some(cond) = cond {
+                    self.walk_expr(cond)?;
+                }
+                if let Some(next) = next {
+                    self.walk_expr(next)?;
+                }
```
DaniPopes: Should be `visit_expr` and remove the two empty `fn visit_*` at the end

tests/ui/resolve/loops.sol
```diff
@@ -8,25 +8,33 @@ function funky() {
 
     for (;;) break;
     for (; i < 40; i++) continue;
-    for (; i++ < 50;) continue;
+    for (; i++ < 50; ) continue;
 
     // ---
     // TODO: `Error: Variable declarations can only be used inside blocks.`
```
DaniPopes: Can you remove these TODOs and just add the error annotations instead of fixing this file? I would like to keep it the same

crates/sema/src/ast_passes.rs
```diff
@@ -25,6 +25,7 @@ struct AstValidator<'sess, 'ast> {
     function_kind: Option<ast::FunctionKind>,
     in_unchecked_block: bool,
     in_loop_depth: u64,
+    x_depth: u64, // How far away are you from the nearest loop
```
DaniPopes: I think it's simpler to just have a separate function for loop bodies that check if they're a declaration statement and emit the error there

tests/ui/resolve/loops.sol
TilakMaddy: Can we add multiple annotations to the same line ? 🤔 `//~ERROR: a1 a2` Like that ? I'm not sure the old error is emitted because I think it is shadowed by this new error. (I should check again)

DaniPopes: See CONTRIBUTING.md. You should use `//~^ ERROR:` on the line below, and then `//~| ERROR:` on the line below that.

TilakMaddy: Thank you, I did that but now, the the old errors of unresolved symbol is no longer emitted in places where this variable decl error is emitted. So what might we do here?

DaniPopes: Ah right, the original fix was fine then

crates/sema/src/ast_passes.rs
```diff
@@ -246,12 +273,24 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
-    fn visit_ty(&mut self, _ty: &'ast ast::Type<'ast>) -> ControlFlow<Self::BreakValue> {
-        ControlFlow::Continue(())
+    fn visit_variable_definition(
+        &mut self,
+        var: &'ast solar_ast::VariableDefinition<'ast>,
+    ) -> ControlFlow<Self::BreakValue> {
+        if self.in_loop() && self.x_depth == 1 {
+            self.dcx()
+                .err("variable declaration statements are not allowed as the body of a loop (for, while, do while), meaning they must be inside of a block")
+                .span(var.span)
```
```suggestion
                .err("variable declarations are not allowed as the body of a loop")
                .span(var.span)
                .help("wrap the statement in a block (`{ ... }`)")
```

crates/sema/src/ast_passes.rs

TilakMaddy: I have removed `x_depth` because I have moved it to a new function in `utils.rs`

crates/sema/src/ast_passes.rs

crates/sema/src/ast_passes/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+            | solar_ast::StmtKind::For { .. } => {
+                return ControlFlow::Continue(());
+            }
+            _ => {}
+        };
+        self.walk_stmt(stmt)
+    }
+}
+
+pub(crate) fn check_if_loop_body_is_a_variable_declaration<'ast>(
```
DaniPopes: This is much easier than you think, you can simply make this a method on `AstValidator` that checks that a statement is one of the variable declaration statements.

See the equivalent code in solc https://github.com/ethereum/solidity/blob/a62a455039120da834fd5faac2e0db8fc222c10d/libsolidity/analysis/SyntaxChecker.cpp#L190-L222
`if (dynamic_cast<VariableDeclarationStatement const*>)` is equivalent to `matches!(stmt, ast::StmtKind::DeclSingle(..) | ast::StmtKind::DeclMulti(..))`; you can also add this check as a method on `StmtKind`.

DaniPopes: This also applies to if statements

TilakMaddy: ✅

DaniPopes (approved):
LGTM

TilakMaddy:
Merged main + I have also cleaned some changes that happened in `loop.sol` due to auto-formatting. Now it's a lot more what changes this PR introduces : )

DaniPopes (approved):

PR #160, opened 2024-12-08
feat: function visibility check for contract functions [Need Help]

TilakMaddy:
I have implemented the function visibility checks. In this PR I have targeted those functions that live inside of contracts.
All my test cases worked, however the problem is there is large number of test files that need to be fixed because of this and I'm not sure I'm going in the right direction. Help appreciated 🙏


tests/ui/resolve/func_visibility.sol
```diff
@@ -0,0 +1,15 @@
...
+    function c() {} //~ERROR: No visibility specified. Did you intend to add external?
+}
+
+contract U3 { //~ERROR: No visibility specified for fallback. Did you intend to add external?
+    fallback() {}
+}
+
+contract U4 {//~ERROR: No visibility specified for receive. Did you intend to add external?
+    receive() {}
+}
```
TilakMaddy: My target cases were these and I have solved them successfully.

TilakMaddy:
@DaniPopes any suggestion you might have for me here ?

crates/sema/src/ast_passes.rs
```diff
@@ -179,16 +179,41 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
+                        let error_stmt = format!(
+                            "No visibility specified. Did you intend to add {suggested_visibility}?",
+                        );
+                        self.dcx().err(error_stmt).span(func_name.span).emit();
+                    }
+                }
+            }
+        }
+
+        if func.kind.is_fallback() || func.kind.is_receive() {
```
DaniPopes: you can group the two branches (function, fallback|receive) into one

crates/sema/src/ast_passes.rs
```diff
@@ -179,16 +179,41 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
+            if let Some(contract) = self.contract {
+                if let Some(func_type_name) = if func.kind.is_fallback() {
+                    Some("fallback")
+                } else if func.kind.is_receive() {
+                    Some("receive")
+                } else {
+                    None
+                } {
+                    let error_stmt = format!(
+                        "No visibility specified for {func_type_name}. Did you intend to add external?",
```
DaniPopes: please make the error message lowercase and move the second part to a `.help` call

crates/sema/src/ast_passes.rs
```diff
@@ -179,16 +179,41 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
+                    );
+                    self.dcx().err(error_stmt).span(contract.name.span).emit();
```
DaniPopes: the span is wrong

crates/sema/src/ast_passes.rs
DaniPopes: the bug is that this check is missing `visibility.is_none()`, easier to spot if you merge the two checks together

crates/sema/src/ast_passes.rs
TilakMaddy: For fallback and receive, function.header.name is `None` that means no span field accessible. That's why I thought I'll put it on the contract. Let me know your suggestion

crates/sema/src/ast_passes.rs

crates/sema/src/ast_passes.rs

crates/sema/src/ast_passes.rs
DaniPopes: Right, `self.span` is the current item's span, you can use that

crates/sema/src/ast_passes.rs
```diff
@@ -196,6 +196,24 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
...
+                    Some(if contract.kind.is_interface() { "external" } else { "public" })
+                } else if func.kind.is_fallback() || func.kind.is_receive() {
+                    Some("external")
+                } else {
+                    None
+                } {
+                    self.dcx()
+                        .err("no visibility specified")
+                        .span(self.span)
+                        .help(format!("do you intend to add {suggested_visibility}?"))
```
```suggestion
                        .help(format!("add `{suggested_visibility}` to the declaration"))
```

DaniPopes (approved):
LGTM

PR #163, opened 2024-12-11
feat: func visibility checks for free functions

TilakMaddy:


tests/ui/typeck/duplicate_overloaded_items.sol
```diff
@@ -12,32 +12,32 @@ event E3(uint) anonymous;
...
 
-function f6(string memory) public {} //~ ERROR: function with same name and parameter types declared twice
-function f6(string calldata) public {}
+function f6(string memory) {} //~ ERROR: function with same name and parameter types declared twice
+function f6(string calldata) {}
 
-function f7(string storage) internal {}
-function f7(string memory) public {}
+// function f7(string storage) internal {} 
+// function f7(string memory) {}
```
TilakMaddy, line 40: @DaniPopes If I don't comment this out, it's causing many errors in this file to be skipped (not emitted). Please can you help me understand what's happening? I see that `f8` functions below have also been commented probably for similar reason

DaniPopes: Compilation fails at an earlier stage, so the later stages which emit these errors don't get reached; commenting out is fine

DaniPopes (approved):

PR #165, opened 2024-12-11
feat: underscores and literals validation

TilakMaddy:
https://github.com/ethereum/solidity/blob/df01dcc7b759f53ba45811979d3286562eab7846/libsolidity/analysis/SyntaxChecker.cpp#L288C1-L322C14


crates/sema/src/ast_passes.rs
```diff
@@ -283,6 +315,10 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
 
     // Intentionally override unused default implementations to reduce bloat.
     fn visit_expr(&mut self, _expr: &'ast ast::Expr<'ast>) -> ControlFlow<Self::BreakValue> {
+        let ast::Expr { kind, .. } = _expr;
+        if let ast::ExprKind::Lit(lit, _) = kind {
+            self.check_underscores_in_number_literals(lit);
+        }
         ControlFlow::Continue(())
     }
```
DaniPopes: Now this also has to recurse into other expressions and types, so remove the comment, the `visit_ty` implementation, and add `walk_expr` at the end

tests/ui/resolve/literals_underscores.sol
```diff
@@ -0,0 +1,14 @@
...
+    uint256 b = 100__0; //~ERROR: invalid use of underscores in number literal
+    uint256 c = 1_.4e10; //~ERROR: invalid use of underscores in number literal
+    uint256 d = 3.4_e10; //~ERROR: invalid use of underscores in number literal
+    uint256 e = 3.4e_10; //~ERROR: invalid use of underscores in number literal
+
+    // exception
+    uint256 f = 3._4e10; // Does not show up
+
+    uint256 g = 1_.4e10 + 3.4e_10; //~ERROR: invalid use of underscores in number literal
+    //~^ERROR: invalid use of underscores in number literal
```
TilakMaddy, line 12: I've added this to prove that expression recursion works for this error message

crates/sema/src/ast_passes.rs
```diff
@@ -60,6 +60,38 @@ impl<'sess> AstValidator<'sess, '_> {
                 .emit();
         }
     }
+
+    fn check_underscores_in_number_literals(&self, lit: &ast::Lit) {
+        let ast::LitKind::Number(_) = lit.kind else {
+            return;
+        };
+        let literal_span = lit.span;
+        let literal_str = lit.symbol.to_string();
```
```suggestion
        let literal_str = lit.symbol.as_str();
```

DaniPopes (approved):

PR #169, opened 2024-12-14
feat: pre-checks for upper bound of contracts' sizes

TilakMaddy:


crates/sema/src/typeck/mod.rs
```diff
@@ -15,9 +17,91 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                let Some(sz) = total_size.checked_add(size_contribution) else {
+                    gcx.dcx().err("contract requires too much storage").span(contract_span).emit();
+                    return;
+                };
+                total_size = sz;
+            }
+        }
+    }
+    //let c = gcx.hir.contract(contract_id).name;
+    //println!("{total_size} - {c}");
```
TilakMaddy: I wonder what's the best way to test that this always prints the values 
```
B - 2
A - 77
M - 8
```

B, A, M are contracts in the test file

crates/sema/src/typeck/mod.rs
```diff
@@ -15,9 +17,95 @@ pub(crate) fn check(gcx: Gcx<'_>) {
         gcx.hir.par_source_ids().for_each(|id| {
             check_duplicate_definitions(gcx, &gcx.symbol_resolver.source_scopes[id]);
         }),
+        gcx.hir.par_contract_ids().for_each(|id| {
+            check_storage_size_upper_bound(gcx, id);
```
DaniPopes: please move to the existing for_each above

crates/sema/src/typeck/mod.rs
```diff
@@ -15,9 +17,95 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                };
+                let Some(sz) = total_size.checked_add(size_contribution) else {
+                    gcx.dcx().err("contract requires too much storage").span(contract_span).emit();
+                    return;
+                };
+                total_size = sz;
+            }
+        }
+    }
+    if cfg!(debug_assertions) {
```
DaniPopes: You should add a -Z flag that is only used in the tests, see UnstableFlags and add it to Session

crates/sema/src/typeck/mod.rs
```diff
@@ -15,9 +17,95 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+/// Reference: https://github.com/ethereum/solidity/blob/03e2739809769ae0c8d236a883aadc900da60536/libsolidity/analysis/ContractLevelChecker.cpp#L556C1-L570C2
+fn check_storage_size_upper_bound(gcx: Gcx<'_>, contract_id: hir::ContractId) {
+    let contract_span = gcx.hir.contract(contract_id).span;
+    let contract_items = gcx.hir.contract_items(contract_id);
+    let mut total_size = U256::ZERO;
+    for item in contract_items {
+        if let hir::Item::Variable(variable) = item {
+            // Skip constant and immutable variables
+            if variable.mutability.is_none() {
+                let Some(size_contribution) = variable_ty_upper_bound_size(&variable.ty.kind, gcx)
```
DaniPopes: Use `gcx.type_of` and match on just the type kind rather than the HIR type or item; in general after this point everything should be done through `Ty`

crates/sema/src/typeck/mod.rs
```diff
@@ -15,9 +17,95 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                else {
+                    gcx.dcx().err("contract requires too much storage").span(contract_span).emit();
+                    return;
+                };
+                let Some(sz) = total_size.checked_add(size_contribution) else {
+                    gcx.dcx().err("contract requires too much storage").span(contract_span).emit();
```
DaniPopes: you can unify these with `size().and_then(|| ...)`

DaniPopes (changes requested):

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/cli/src/lib.rs
```diff
@@ -138,6 +138,7 @@ fn run_compiler_with(args: Args, f: impl FnOnce(&Compiler) -> Result + Send) ->
     sess.stop_after = args.stop_after;
     sess.dump = args.unstable.dump.clone();
     sess.ast_stats = args.unstable.ast_stats;
+    sess.storage_sz_ub = args.unstable.storage_sz_ub;
```
TilakMaddy: This region wouldn't run in `#[cfg(test)]` mode, what can I do ?

crates/sema/src/typeck/mod.rs
TilakMaddy: I have added a -Zstorage-sz-ub but it wouldn't let me tack on a #[cfg(test)] 😕

tests/ui/typeck/contract_storage_size_check.sol
```diff
@@ -0,0 +1,41 @@
+//@ compile-flags: -Zstorage-sz-ub
+
```
TilakMaddy, line 2: Here

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +12,91 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                            .span(contract_span)
+                            .emit();
+                        return;
+                    }
+                }
+            }
+        }
+    }
+
+    if gcx.sess.storage_sz_ub {
```
DaniPopes: let's call it like "print-contract-storage-sizes"

crates/cli/src/lib.rs
DaniPopes: It can't and shouldn't be cfg(test)

crates/cli/src/cli.rs
```diff
@@ -138,6 +138,9 @@ pub struct UnstableFeatures {
     #[cfg(test)]
     #[arg(long)]
     test_value: Option<usize>,
+
+    #[arg(long)]
+    pub storage_sz_ub: bool,
```
DaniPopes: please move to above help

crates/sema/src/typeck/mod.rs
TilakMaddy: We only check the upper bound of storage size, so I'm afraid that wouldn't be accurate depiction of what users are looking for. 

for example:
If there's 

```
contract C {
  int128 a;
  int128 b;
}
```
Currently, this totals to "2", which is less than 2^256 so passes the check. (this check is pre-condition to make sure the slot allocation algorithm will work properly later)

Whereas someone wanting to print storage size would expect 128 + 128 = 256 bits so `1` slot. (because of how solidity allocates the slots)

If we go with this, do we make another flag for the _actual_ contract storage size then?

TilakMaddy: Tbh, I feel like this check isn't necessary at all lol ! I don't why solidity does it... Maybe I should convert it to actual contract storage size haha

TilakMaddy: @DaniPopes isn't this more like "print-contract-max-storage-sizes" ?

DaniPopes: > @DaniPopes isn't this more like "print-contract-max-storage-sizes" ?

sure

DaniPopes: > If we go with this, do we make another flag for the actual contract storage size then?

This probably will go with the actual storage implementation later on, the "storage-layout" output

crates/config/src/opts.rs
```diff
@@ -158,6 +158,10 @@ pub struct UnstableOpts {
     #[cfg_attr(feature = "clap", arg(long))]
     pub ast_stats: bool,
 
+    /// Print contract's max storage size.
+    #[cfg_attr(feature = "clap", arg(long))]
+    pub contract_max_storage_size: bool,
```
DaniPopes:
```suggestion
    pub print_contract_max_storage_size: bool,
```

TilakMaddy:
@DaniPopes I have also added a test case to demonstrate what happens when more than 2^256 - 1 storage slots are requested.

crates/cli/src/cli.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +12,91 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+            let elem_size = ty_upper_bound_storage_var_size(ty, gcx)?;
+            uint.checked_mul(elem_size)
+        }
+        TyKind::Struct(struct_id) => {
+            let strukt = gcx.hir.strukt(struct_id);
+            // Reference https://github.com/ethereum/solidity/blob/03e2739809769ae0c8d236a883aadc900da60536/libsolidity/ast/Types.cpp#L2303C1-L2309C2
+            let mut total_size = U256::from(1);
+            for field_id in strukt.fields {
+                let variable = gcx.hir.variable(*field_id);
+                let t = gcx.type_of_hir_ty(&variable.ty);
```
DaniPopes: please don't use `type_of_hir_ty` since it recomputes the type which is already computed for the variable; instead use `type_of_item` and `struct_field_types`, and also you have to check for recursive struct so as to not recurse infinitely with `struct_recursiveness`

applies above in `check_storage_size_upper_bound` too

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +12,91 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                    }
+                }
+            }
+        }
+    }
+
+    if gcx.sess.opts.unstable.contract_max_storage_size {
+        let full_contract_name = format!("{}", gcx.contract_fully_qualified_name(contract_id));
+        eprintln!("{full_contract_name} requires {total_size} maximum storage");
+    }
```
DaniPopes: We can turn this into an error instead of eprintln so that we can use an annotation like `//~ ERROR: storage upper bound: 2` in the test

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +12,91 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+                            .emit();
+                        return;
+                    }
+                }
+            }
+        }
+    }
+
+    if gcx.sess.opts.unstable.contract_max_storage_size {
+        let full_contract_name = format!("{}", gcx.contract_fully_qualified_name(contract_id));
```
DaniPopes: This `format!` is unnecessary

DaniPopes (changes requested):

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/config/src/opts.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
TilakMaddy: @DaniPopes I have applied the suggestion and got the tests passing but I have 2 doubts

* Why check again for struct recursiveness when it's already done [here](https://github.com/paradigmxyz/solar/blob/98ca2f1f915516290c5bd28c61a6ce3ec6eb4cc2/crates/sema/src/ty/mod.rs#L687C1-L689C11) ? Also I'm not sure where to add the logic

* Why does doing `gcx.type_of_item` wrap all the contract's elementary-type-storage-variables in `Ref(.., DataLocation::Storage)` ? Is it by design or bug? Because technically there could be "references" to "storage  variables" used within functions that which does NOT make them the storage variables themseleves?

[Given `gcx.type_of_hir_ty(..)` doesn't do this "wrapping"] So they are not quite interchangeable 🤔 . . .

PR #170, opened 2024-12-15
feat: warn when payable fallback function found but no receive function

TilakMaddy:
https://github.com/ethereum/solidity/blob/03e2739809769ae0c8d236a883aadc900da60536/libsolidity/analysis/ContractLevelChecker.cpp#L544C1-L554C2

### DIFF
--- a/crates/sema/src/typeck/mod.rs
+++ b/crates/sema/src/typeck/mod.rs
@@ -5,6 +5,7 @@ use crate::{
 };
 use rayon::prelude::*;
 use solar_data_structures::{map::FxHashSet, parallel};
+use solar_interface::error_code;
 
 pub(crate) fn check(gcx: Gcx<'_>) {
     parallel!(
@@ -31,7 +32,8 @@ fn check_payable_fallback_without_receive(gcx: Gcx<'_>, contract_id: hir::Contra
         gcx.dcx()
             .warn("contract has a payable fallback function, but no receive ether function")
             .span(contract.name.span)
-            .help("consider adding a `receive() external payable { ... }`")
+            .code(error_code!(3628))
+            .help("consider changing fallback to receive")
             .emit();
     }
 }
--- a/crates/sema/src/typeck/mod.rs
+++ b/crates/sema/src/typeck/mod.rs
@@ -11,11 +11,35 @@ pub(crate) fn check(gcx: Gcx<'_>) {
         gcx.sess,
         gcx.hir.par_contract_ids().for_each(|id| {
             check_duplicate_definitions(gcx, &gcx.symbol_resolver.contract_scopes[id]);
+            check_payable_fallback_without_receive(gcx, id);
         }),
         gcx.hir.par_source_ids().for_each(|id| {
             check_duplicate_definitions(gcx, &gcx.symbol_resolver.source_scopes[id]);
         }),
     );
+}
+
+fn check_payable_fallback_without_receive(gcx: Gcx<'_>, contract_id: hir::ContractId) {
+    let mut payable_fallback_found = false;
+    let mut receive_found = false;
+    for item in gcx.hir.contract_items(contract_id) {
+        if let hir::Item::Function(f) = item {
+            if f.kind.is_fallback() && f.state_mutability.is_payable() {
+                payable_fallback_found = true;
+            } else if f.kind.is_receive() {
+                receive_found = true;
+            }
+        }
+    }
+
+    if payable_fallback_found && !receive_found {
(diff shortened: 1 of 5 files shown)


crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +11,37 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
         gcx.hir.par_source_ids().for_each(|id| {
             check_duplicate_definitions(gcx, &gcx.symbol_resolver.source_scopes[id]);
         }),
     );
 }
 
+fn check_payable_fallback_without_receive(gcx: Gcx<'_>, contract_id: hir::ContractId) {
+    let mut payable_fallback_found = false;
+    let mut receive_found = false;
+    for item in gcx.hir.contract_items(contract_id) {
```
DaniPopes: fallback and receive are stored in contract directly, don't need to scan for them

TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +11,31 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+    let payable_fallback_found =
+        contract.fallback.is_some_and(|f| gcx.hir.function(f).state_mutability.is_payable());
+
+    let receive_found = contract.receive.is_some();
+
+    if payable_fallback_found && !receive_found {
+        gcx.dcx()
+            .warn("contract has a payable fallback function, but no receive ether function")
+            .span(contract.name.span)
+            .help("consider adding a `receive() external payable { ... }`")
```
DaniPopes: "consider changing fallback to receive" maybe?

crates/sema/src/typeck/mod.rs
```diff
@@ -11,13 +11,31 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+    let contract = gcx.hir.contract(contract_id);
+
+    let payable_fallback_found =
+        contract.fallback.is_some_and(|f| gcx.hir.function(f).state_mutability.is_payable());
+
+    let receive_found = contract.receive.is_some();
+
+    if payable_fallback_found && !receive_found {
+        gcx.dcx()
+            .warn("contract has a payable fallback function, but no receive ether function")
```
DaniPopes: can you also add the error code

DaniPopes (approved):
lgtm, I would like to hold off on this because there is currently no way to turn off warnings

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅

crates/sema/src/typeck/mod.rs
TilakMaddy: ✅ Added 3628
Reference: https://github.com/search?q=repo%3Aethereum%2Fsolidity%203628&type=code

TilakMaddy:
Sorry wait...

crates/sema/src/typeck/mod.rs
```diff
@@ -5,19 +5,39 @@ use crate::{
...
+            .span(contract.name.span)
+            .code(error_code!(3628))
+            .help("consider changing fallback to receive")
+            .emit();
```
TilakMaddy: I guess I'll do a span_help here also : )

TilakMaddy: ✅

tests/ui/parser/payable_fallback_without_receive.sol
```diff
@@ -0,0 +1,15 @@
...
+}
+
+contract P2 is P1 {
+    receive() external payable {}
+}
+
+contract P3 {
+    fallback() external payable {}
+
+    receive() external payable {}
```
DaniPopes, line 13: Please add annotations for the warnings //~

TilakMaddy: It's a bit strange 🤔 .. I tried 

* Same line variant (//~ WARN: ...)
* Next line variant(//~^ WARN: ...)

but they both fail on `cargo uitest payable_fallback` 😕

DaniPopes: pushed a fix https://github.com/paradigmxyz/solar/pull/170/commits/843313581ad4202911c2f2b954b823d2756b80f4
the diagnostic was missing the main span (`.span()`) but then it also required a workaround for a edge case in `ui_test`

DaniPopes (approved):

DaniPopes (approved):

PR #171, opened 2024-12-15
feat: check duplicate calls to base constructor

TilakMaddy:
https://github.com/ethereum/solidity/blob/df01dcc7b759f53ba45811979d3286562eab7846/libsolidity/analysis/ContractLevelChecker.h#L73C2-L78C4


crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -196,6 +196,84 @@ impl super::LoweringContext<'_, '_, '_> {
...
+
+                for bc_error in bc_errors {
+                    self.dcx()
+                        .err(format!(
+                            "contract makes multiple calls to base constructor {}",
+                            bc_error.base_con_name
+                        ))
+                        .span(main_contract.name.span)
+                        .emit();
+                    self.dcx().note("first call here").span(bc_error.first_call_span).emit();
```
DaniPopes, line 270: these should be notes on the error, this creates a separate diagnostic

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -196,6 +196,84 @@ impl super::LoweringContext<'_, '_, '_> {
...
+                            continue;
+                        };
+                        if let Some(old_value) = ctor_args.insert(base_id, m.name.span()) {
+                            let base_con_name = self.hir.contract(base_id).name.as_str();
+                            bc_errors.push(BCError::from(base_con_name, old_value, m.name.span()));
+                        }
+                    }
+                }
+
+                for bc_error in bc_errors {
```
DaniPopes, line 262: you can make this a closure that is called instead of pushing

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -196,6 +196,84 @@ impl super::LoweringContext<'_, '_, '_> {
         }
     }
 
+    #[instrument(level = "debug", skip_all)]
+    pub(super) fn check_base_ctor_args(&self) {
+        for main_contract in self.hir.contracts() {
```
DaniPopes, line 201: why is this using the AST instead of being a HIR check like the other ones you did? this shouldnt happen during lowering

DaniPopes:
hey @TilakMaddy are you still interested in this PR? otherwise no problem we can take over

DaniPopes:
actually this pr was superseded by https://github.com/paradigmxyz/solar/issues/304 but the other one is still valid

PR #172, opened 2024-12-15
feat: check placeholders inside unchecked blocks

TilakMaddy:
https://github.com/ethereum/solidity/blob/03e2739809769ae0c8d236a883aadc900da60536/libsolidity/analysis/SyntaxChecker.cpp#L368C1-L379C2


DaniPopes (approved):
lgtm

PR #178, opened 2024-12-31
Fixed issues with JSON key usage, variable naming, and redundant semi…

mdqst:
A few issues in the script that could lead to errors or confusion during execution, and I’ve made improvements to address them:

1. **Fixed usage of `$file` in JSON keys:**
   The previous implementation used `$file` as part of the JSON key, which could cause issues because `$file` is the full file path, not just the filename. I’ve adjusted this to extract only the filename without the path or extension, ensuring it aligns with how Solc generates contract names in the JSON output.

   **Fix:** I used `basename` to extract the filename:
   ```bash
   filename=$(basename "$file" .sol)
   jq -r ".contracts[\"$filename:$contract\"][\"generated-sources-runtime\"][0].contents"
   ```

2. **Renamed the variable `basename` to avoid conflicts:**
   The variable `basename` was conflicting with the built-in `basename` command in Unix-like systems. I’ve renamed it to `file_basename` to avoid any potential issues.

3. **Removed unnecessary semicolon after `shift`:**
   The semicolon after the `shift` command was not needed, as it doesn't affect the syntax. I’ve removed it for cleaner and more readable code.

### DIFF
--- a/scripts/combined_json.sh
+++ b/scripts/combined_json.sh
@@ -5,13 +5,12 @@ set -eo pipefail
 
 file="$1"
 [ -z "$file" ] && echo "Usage: $0 <file>" && exit 1
-shift;
-basename="$(basename "$file" .sol)"
-
+shift
+file_basename="$(basename "$file" .sol)"
 contract="${2:-C}"
 
 solc --combined-json generated-sources-runtime "$file" --optimize \
 | jq . \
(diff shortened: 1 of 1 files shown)


DaniPopes:
This change is unnecessary, and looks like AI

PR #205, opened 2025-02-08
feat(sema): implement function pointer member resolution

VolodymyrBg:
- Add implementation for function pointer member resolution
- Add support for .selector and .address members
- Remove todo!() macro in members.rs
- Improve type system completeness for function pointers

This change completes the implementation of function pointer type  members in the Solidity type system, allowing proper resolution  of .selector (returns bytes4) and .address (returns address) members  when working with function pointers.

Part of the type system improvements.


DaniPopes:
Hey, thanks for the contribution, however this requires more checks and is going to be implemented in this branch: https://github.com/paradigmxyz/solar/tree/dani/typeck

PR #210, opened 2025-02-24
feat(hir): expose hir from parsing context

grandizzy:
- access Hir from calling ParsingContext::parse_and_lower_to_hir

### DIFF
--- a/crates/sema/src/lib.rs
+++ b/crates/sema/src/lib.rs
@@ -25,12 +25,14 @@ mod ast_lowering;
 mod ast_passes;
 
 mod parse;
-use crate::hir::{Arena, Hir};
 pub use parse::{ParsedSource, ParsedSources, ParsingContext};
 
 pub mod builtins;
 pub mod eval;
+
 pub mod hir;
+pub use hir::{Arena, Hir};
+
 pub mod ty;
 
 mod typeck;
--- a/crates/sema/src/lib.rs
+++ b/crates/sema/src/lib.rs
@@ -25,6 +25,7 @@ mod ast_lowering;
 mod ast_passes;
 
 mod parse;
+use crate::hir::{Arena, Hir};
 pub use parse::{ParsedSource, ParsedSources, ParsingContext};
 
 pub mod builtins;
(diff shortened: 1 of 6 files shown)


crates/sema/src/lib.rs
```diff
@@ -25,6 +25,7 @@ mod ast_lowering;
 mod ast_passes;
 
 mod parse;
+use crate::hir::{Arena, Hir};
```
DaniPopes: should maintain style of the other fns, using hir:: as a module instead of importing

crates/sema/src/lib.rs
```diff
@@ -100,6 +101,32 @@ pub fn parse_and_resolve(pcx: ParsingContext<'_>) -> Result<()> {
     Ok(())
 }
 
+/// Parses and lowers to HIR, recursing into imports.
+pub fn parse_and_lower_to_hir<'hir>(
```
DaniPopes: I want to avoid any code dup here and also avoid using Hir directly, but instead through gcx; I think what we can do here is something like this
```rust
fn parse_and_resolve() {
    let gcx = *parse_and_lower()?;
    analysis(gcx)?;
    Ok(())
}

fn parse_and_lower<'hir>(pcx: '_, hir_arena: 'hir) -> impl Deref<Target=Gcx<'hir>> {
    // all the code currently in parse_and_resolve ...
    GcxWrapper(...)
}

struct GcxWrapper(GlobalContext);
impl Deref for GcxWrapper {
    // move the unsafe here
}
impl Drop for GcxWrapper {
    // move from OnDrop closure
}
```

This way we can keep the drop impl and avoid a closure for using Gcx

DaniPopes (changes requested):

crates/sema/src/lib.rs
grandizzy: changed in https://github.com/paradigmxyz/solar/pull/210/commits/a61687faaa4a44402832078b0df77393df41690c

crates/sema/src/lib.rs
grandizzy: I made changes in https://github.com/paradigmxyz/solar/pull/210/commits/a650e8bba8e1fc59e6499302c8918ad8f3ab5dad to avoid code dupe but my attempts to access it through Gcx failed the macos tests with `thread 'solar-0' panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lasso-0.7.3/src/threaded_rodeo.rs:518:32:` so kind of blocked on this, @DaniPopes pls advise how to proceed to have this one merged. thank you!

DaniPopes (approved):

PR #239, opened 2025-03-29
fix: unsound transmute

gio256:
The following use of `std::mem::transmute` in `Builtin::from_index` was previously unsound, as there are few layout [guarantees](https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.rust.layout) made for `repr(Rust)` types. This `transmute` relies on the assumption that an enum has the same alignment as the smallest integer type that can hold all its variants, but `repr(Rust)` enums make no such guarantee.

https://github.com/paradigmxyz/solar/blob/27f20c504cacf59c19c902dc925ff85e49bb1c37/crates/sema/src/builtins/mod.rs#L266-L277

This is fixed by tagging the `Builtin` enum `repr(u8)`, which provides the [guarantees](https://doc.rust-lang.org/reference/type-layout.html#primitive-representation-of-field-less-enums) needed here. I have also added a safety comment explaining the remaining assumptions (e.g. the enum is field-less).


DaniPopes (approved):
Nice catch, thanks!

PR #248, opened 2025-04-02
fix: public transient assert

grandizzy:


tests/ui/resolve/transient.sol
```diff
@@ -0,0 +1,3 @@
+contract A {
+    public transient tempValue; //~ ERROR: expected contract item (function, variable, struct, or modifier definition), found keyword `public`
```
DaniPopes: missing type, this is the wrong error

grandizzy: changed the test and removed the transient.stderr, with the assertion the test would actually panic

DaniPopes (approved):
Thanks!

PR #276, opened 2025-04-19
perf: add and default to using mimalloc

DaniPopes:
mimalloc consistently is performing equal to or better than jemalloc, is smaller in binary size, is more portable and is faster to compile


gakonst:
Should we consider for reth?

DaniPopes:
Possibly but it does use the jemalloc profiling features so won't be fully identical

Issue #303, opened 2025-05-02
Implement `ContractChecker`


DaniPopes:
- [ ] `checkDuplicateFunctions`
- [ ] `checkDuplicateEvents`
- [ ] `checkReceiveFunction`
- [ ] `OverrideChecker`, `checkAbstractDefinitions` https://github.com/paradigmxyz/solar/issues/301
- [ ] `checkBaseConstructorArguments` -- should be implemented as part of HIR lowering
- [ ] `checkExternalTypeClashes`
- [x] `checkHashCollisions` (implemented in `interface_functions`)
- [ ] `checkLibraryRequirements`
- [ ] `checkBaseABICompatibility`
- [x] `checkPayableFallbackWithoutReceive` https://github.com/paradigmxyz/solar/issues/170
- [ ] `checkStorageSize` https://github.com/paradigmxyz/solar/pull/169

Reference solc: https://github.com/ethereum/solidity/blob/d42f92bd68d76db01b0dc17477cd4c7716059d93/libsolidity/analysis/ContractLevelChecker.cpp#L84-L102

nadtech-hub:
will do if assigned

PR #312, opened 2025-05-11
feat: typecheck for external type clashes

DhairyaSethi:
implement `checkExternalTypeClashes` in an effort for #303 

[solc reference](https://github.com/ethereum/solidity/blob/d42f92bd68d76db01b0dc17477cd4c7716059d93/libsolidity/analysis/ContractLevelChecker.cpp#L430-L465), includes all relevant ui tests present in solc in one test file

notes:
- we don't use `gcx.hir.functions` method instead of `gcx.hir.contract_item_ids` since we want to include inherited methods
- unlike the solc implementation: 
  - we don't run the pre-processing on external variables since their external is already populated & covered in the functions branch
  - we omit libraries since they produce external methods & it's methods are implicitly omitted in the `if(functionType->interfaceFunctionType())` [branch](https://github.com/ethereum/solidity/blob/d42f92bd68d76db01b0dc17477cd4c7716059d93/libsolidity/analysis/ContractLevelChecker.cpp#L440)

### DIFF
--- a/crates/sema/src/typeck/mod.rs
+++ b/crates/sema/src/typeck/mod.rs
@@ -1,10 +1,14 @@
 use crate::{
     ast_lowering::resolve::{Declaration, Declarations},
-    hir::{self, Res},
+    hir::{self, Item, ItemId, Res},
     ty::{Gcx, Ty},
 };
+use alloy_primitives::{keccak256, U256};
 use rayon::prelude::*;
-use solar_data_structures::{map::FxHashSet, parallel};
+use solar_data_structures::{
+    map::{FxHashMap, FxHashSet},
+    parallel,
+};
 use solar_interface::error_code;
 
 pub(crate) fn check(gcx: Gcx<'_>) {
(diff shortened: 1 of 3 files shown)


crates/sema/src/typeck/mod.rs
```diff
@@ -13,13 +17,48 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+        for (i, &item) in items.iter().enumerate() {
+            if let Some(&dup) = items.iter().skip(i + 1).find(|&&other| {
+                !same_external_params(gcx, gcx.type_of_item(item), gcx.type_of_item(other))
+            }) {
+                gcx.dcx()
+                    .err(
+                        "function overload clash during conversion to external types for arguments",
+                    )
+                    .code(error_code!(9914))
+                    .span(gcx.item_span(dup))
```
DaniPopes: I believe it's better to report on the `item` span here, and add a span_note for the "other declaration is here" on `dup`

DhairyaSethi: ty good call, pushed [9f8e0fb](https://github.com/paradigmxyz/solar/pull/312/commits/9f8e0fb789c4d6cd3dd1fbdd0e98dda7352c7433)

solc errors on the `dup` declaration instead, so can switch the ordering to match that as well (error on dup, help on the first seen decl) lmk

DaniPopes (approved):

crates/sema/src/typeck/mod.rs
```diff
@@ -13,13 +17,48 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
+    if gcx.hir.contract(contract_id).kind.is_library() {
+        return;
+    }
+
+    let mut external_declarations: FxHashMap<U256, Vec<ItemId>> = FxHashMap::default();
+
+    for item_id in gcx.hir.contract_item_ids(contract_id) {
+        match gcx.hir.item(item_id) {
+            Item::Function(f) if f.is_part_of_external_interface() => {
+                let s = keccak256(gcx.item_signature(item_id));
```
DaniPopes: You'll need to make it pub(crate)

```suggestion
                let s = gcx.item_selector(item_id);
```

crates/sema/src/typeck/mod.rs
```diff
@@ -13,13 +17,48 @@ pub(crate) fn check(gcx: Gcx<'_>) {
...
         }),
     );
 }
 
+fn check_external_type_clashes(gcx: Gcx<'_>, contract_id: hir::ContractId) {
+    if gcx.hir.contract(contract_id).kind.is_library() {
+        return;
+    }
+
+    let mut external_declarations: FxHashMap<U256, Vec<ItemId>> = FxHashMap::default();
```
DaniPopes:
```suggestion
    let mut external_declarations: FxHashMap<B256, Vec<ItemId>> = FxHashMap::default();
```

crates/sema/src/typeck/mod.rs
DhairyaSethi: done [68f23ef](https://github.com/paradigmxyz/solar/pull/312/commits/68f23efa440a33b4eedd90e35a18a345847879db) (limiting vis to crate is a good call)

we don't necessarily need to use keccak here since the key is not used except for collecting; wondering if just FxHasher is more performant (will hashing twice, once to find bucket in the map), or we can tell the map to drop key references (of `item_signature`) since we will never iterate over them

crates/sema/src/typeck/mod.rs
DhairyaSethi: [68f23ef](https://github.com/paradigmxyz/solar/pull/312/commits/68f23efa440a33b4eedd90e35a18a345847879db)

crates/sema/src/typeck/mod.rs
DaniPopes: yeah but this is fine since we already calculated all the selectors for external functions so we're just reusing the hash, and rehashing is very cheap

DaniPopes (approved):
lgtm, please re-bless the tests

DhairyaSethi:
@DaniPopes ty, done

PR #316, opened 2025-05-14
docs: delete hyperlink to pull_request_template

gap-editor:
we don't have PULL_REQUEST_TEMPLATE.md so i think we need to delete this hyperlink like we have it [HERE](https://github.com/paradigmxyz/reth/blob/main/CONTRIBUTING.md?plain=1#L152)


DaniPopes:
I've removed the section in 13b943afb335509431f678cc2fa6c0a96dd067b2

PR #318, opened 2025-05-16
feat: add span in FunctionHeader

### LINKED ISSUE #317: Add Span to FunctionHeader
### Describe the feature

Having the Span for a function header would be useful for third party crate integrating Solar as parser and using that information (for highlighting for instance). Currently, it seems like the only way to use the Span of a function header is to start from the whole Item and substract the ItemFunction.body_span (ie `item.span().with_hi(body_span.lo())`), which isn't ideal and has hedge cases (iic, only works for 1 True Bracket Style for instance).

### Additional context

The change would mostly adding a field header_span in https://github.com/paradigmxyz/solar/blob/86b5bd276e72be4745cfb5cbcd5c7e029152b36e/crates/ast/src/ast/item.rs#L435 and populate it in https://github.com/paradigmxyz/solar/blob/86b5bd276e72be4745cfb5cbcd5c7e029152b36e/crates/parse/src/parser/item.rs#L163

Happy to open a PR if it can help!

simon-something:
Add and populate a span in FunctionHeader

Closes #317

### DIFF
--- a/crates/ast/src/ast/item.rs
+++ b/crates/ast/src/ast/item.rs
@@ -433,6 +433,9 @@ impl ItemFunction<'_> {
 /// A function header: `function helloWorld() external pure returns(string memory)`.
 #[derive(Debug, Default)]
 pub struct FunctionHeader<'ast> {
+    /// The span of the function header.
+    pub span: Span,
+
     /// The name of the function.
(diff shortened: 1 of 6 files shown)


DaniPopes (approved):
Thanks!

Issue #304, opened 2025-05-02
Resolve constructor base arguments


DaniPopes:
### Describe the feature

During HIR lowering, for each contract we need to resolve where the base arguments are.

In the following example:
```solidity
contract A {
    constructor(uint) {}
}

contract C is A(1) {}
contract D is A { constructor() A(1) {} }
contract E is D {}
```

All `C`, `D`, and `E` are valid ways to pass `1` to `A`'s inherited constructor.

We want to have a way to resolve the arguments for all of the base constructors; this can be done with a `IndexSlice<BaseIndex, &'hir [Expr<'hir>]>` that maps the base index (index of the `linearized_bases` list) to its arguments (expressions).

### Additional context

This is implemented in solc in `ContractLevelChecker::checkBaseConstructorArguments`: https://github.com/ethereum/solidity/blob/d42f92bd68d76db01b0dc17477cd4c7716059d93/libsolidity/analysis/ContractLevelChecker.cpp#L323

nadtech-hub:
will do if assigned

DhairyaSethi:
gm i can take this on as well, plan would be to set the introduce the indexSlice & populate while lowering the modifiers on contract & [constructor](https://github.com/paradigmxyz/solar/blob/main/crates/sema/src/ast_lowering/resolve.rs#L335) definitions first and then add solc equivalent validations either during lowering itself or at type [check](https://github.com/paradigmxyz/solar/blob/main/crates/sema/src/typeck/mod.rs#L14)

PR #322, opened 2025-05-20
feat: resolve ctor base args

DhairyaSethi:
in an effort for #304

notes:
- extracted to a separate method instead of lowering when resolving modifiers [here](https://github.com/paradigmxyz/solar/blob/86b5bd276e72be4745cfb5cbcd5c7e029152b36e/crates/sema/src/ast_lowering/resolve.rs#L331-L335) bc it's cleaner this way to resolve per contract since that loop runs per function
- should allow named constructor args?
- omits override/abstract and argument count checks


crates/sema/src/hir/mod.rs
```diff
@@ -537,6 +540,10 @@ pub struct Contract<'hir> {
     /// Note that this only includes items defined in the contract itself, not inherited items.
     /// For getting all items, use [`Hir::contract_items`].
     pub items: &'hir [ItemId],
+    /// The constructor base arguments (if any).
+    ///
+    /// The index maps against the position in `linearized_bases`.
+    pub base_args: IndexVec<BaseIndex, &'hir [Expr<'hir>]>,
```
DaniPopes: you want this to be `&'hir [&'hir [Expr<'hir>]]`, since otherwise the Vec will never get deallocated

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -234,6 +234,86 @@ impl super::LoweringContext<'_, '_, '_> {
             c.receive = receive;
         }
     }
+
+    #[instrument(level = "debug", skip_all)]
+    pub(super) fn resolve_base_args(&mut self) {
+        let next_id = &AtomicUsize::new(0);
```
DaniPopes: IDs should be unique so you can move the `&AtomicUsize::new(0)` out from `resolve_symbols` and take it in as an argument in both functions

crates/sema/src/hir/mod.rs
DhairyaSethi: ah ok so a raw slice like https://github.com/paradigmxyz/solar/pull/322/commits/d4a2f94b41f987ddaa992e3c997af27a05949945?
should it map to an array `ExprId` instead of `Expr` struct instead since we're allocating twice here?

crates/sema/src/ast_lowering/resolve.rs
DhairyaSethi: https://github.com/paradigmxyz/solar/pull/322/commits/ed15ef8d2bceab0439bdd6e116953dbfd71adc55 (had also realised impl was omitting cases when arguments were soley passed in contract modifiers so now we chain the two cases instead)

PR #329, opened 2025-05-23
cache len

malik672:
cached the len since it's used multiple times


DaniPopes:
`<[u8]>::len`/`str::len` has no overhead, it's already in a register

malik672:
> `<[u8]>::len`/`str::len` has no overhead, it's already in a register

never a perf(knew it made no diff), wanted to make changes to the unescape function and just decided to seperate them

Issue #315, opened 2025-05-14
Truncated line ending in single-line doc-comment on Windows


beeb:
I tried this code on Windows (with `\r\n` line endings):

```solidity
contract Test {
    /// @notice An enum
    enum TestEnum {
        First,
        Second
    }
}
```

I expected to see this happen: the `DocComment` should have a `symbol: " @notice An enum\r\n"` or maybe `symbol: " @notice An enum"`. 

Instead, this happened: `symbol: " @notice An enum\r"`.

### Meta

`solar --version`:
```
0.1.3
```

DaniPopes:
Related line of code https://github.com/paradigmxyz/solar/blob/2de670d576c0b27c10967aa82c5f5870545c694c/crates/parse/src/lexer/cursor/mod.rs#L203

DaniPopes:
https://github.com/paradigmxyz/solar/pull/340

PR #340, opened 2025-05-28
fix: windows eol lexing

simon-something:
Windows, unlike other OS, uses CRLF (`\r\n`) as end-of-line delimiter. As the lexer relies on the end-of-line when handling `//`/`///` single line comments, this creates a cross-platform inconsistent behavior (the final `\r` ends up included in the comment parsed, see  #315 ).

This PR adds a test for this hedge case and a suggested fix (reusing memchr ability to match 2 different bytes in a given haystack)

### DIFF
--- a/crates/parse/src/lexer/cursor/mod.rs
+++ b/crates/parse/src/lexer/cursor/mod.rs
@@ -198,7 +198,8 @@ impl<'a> Cursor<'a> {
         // `////` (more than 3 slashes) is not considered a doc comment.
         let is_doc = matches!(self.first(), b'/' if self.second() != b'/');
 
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):
In rustc this is handled differently:
- https://github.com/rust-lang/rust/blob/77101febcc9662a076ff43887497cd5b30674d93/compiler/rustc_parse/src/lexer/mod.rs#L576
- https://github.com/rust-lang/rust/blob/77101febcc9662a076ff43887497cd5b30674d93/compiler/rustc_span/src/lib.rs#L2303

However I think this is also fine since there is no perf loss; thanks!

PR #344, opened 2025-05-29
Perf: Replace num_int with Rug

malik672:
Replace num_int with Rug

From: https://github.com/tczajka/bigint-benchmark-rs


DaniPopes:
Thanks, it might be faster but I don't really care about it as I'm either going to remove arbitrary size integers or just disallow them after parsing

This also adds a few extra dependencies, including a C library which is avoidable

PR #347, opened 2025-06-03
perf(lexer): optimize `is_id_continue_byte` using bitwise operations

malik672:
**Description:**
This PR optimizes the is_id_continue_byte function by replacing the matches! macro with direct bitwise operations for better performance in identifier validation.

**Changes:**
Refactored is_id_continue_byte to use is_id_start_byte() + explicit digit range check
Replaced matches! pattern with bitwise operations (& for range checks, || for combining conditions)
Maintained identical functionality while improving performance

**Performance Impact:**
Benchmarks show measurable improvement in identifier validation performance, which is critical for parsing hot paths where this function may be called millions of times.

Honestly idk why this is faster despite having low uOps Per Cycle compared to the previous impl

### DIFF
--- a/crates/parse/src/lexer/cursor/mod.rs
+++ b/crates/parse/src/lexer/cursor/mod.rs
@@ -41,7 +41,7 @@ pub const fn is_id_continue(c: char) -> bool {
     is_id_continue_byte(ch2u8(c))
 }
 /// Returns `true` if the given character is valid in a Solidity identifier.
-#[inline(always)]
+#[inline]
 pub const fn is_id_continue_byte(c: u8) -> bool {
     let is_number = (c >= b'0') & (c <= b'9');
     is_id_start_byte(c) || is_number
@@ -62,9 +62,6 @@ pub const fn is_ident(s: &str) -> bool {
 ///
 /// See [`is_ident`] for more details.
 pub const fn is_ident_bytes(s: &[u8]) -> bool {
-    // Note: valid idents can only contain ASCII characters, so we can use the byte representation
-    // here.
-
     let [first, ref rest @ ..] = *s else {
         return false;
     };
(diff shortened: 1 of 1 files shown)


malik672:
> @DaniPopes we can further elimate stack frame if we were do something like this:
> 
> https://godbolt.org/z/c5jsb3zxn

of course this severely hurts readability

malik672:
> > @DaniPopes we can further elimate stack frame if we were do something like this:
> > https://godbolt.org/z/c5jsb3zxn
> 
> of course this severely hurts readability
> 
> ```
> pub const fn is_ident_bytes(s: &[u8]) -> bool {
>     // Note: valid idents can only contain ASCII characters, so we can use the byte representation
>     // here.
>     let [first, ref rest @ ..] = *s else {
>         return false;
>     };
> 
>     if !is_id_start_byte(first) {
>         return false;
>     }
> 
>     let mut i = 0;
>     while i < rest.len() {
>         let a = rest[i];
>         if !is_id_start_byte(a) | (a >= b'0') & (a <= b'9') {
>             return false;
>         }
>         i += 1;
>     }
> 
>     true
> }
> 
> #[inline]
> pub const fn is_id_start_byte(c: u8) -> bool {
>     matches!(c, b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'$')
> }
> 
> // #[inline]
> // pub const fn is_id_continue_byte(c: u8) -> bool {
> //     //  let is_number = (c >= b'0') & (c <= b'9');
> //     // is_id_start_byte(c) | is_number
> //     // matches!(c, b'a'..=b'z' | b'A'..=b'Z' | b'0'..=b'9' | b'_' | b'$')
> // }
> ```

or just consider using [inline(always)] instead of inline

malik672:
`inline(always)` from local benchwark on 
```#[inline]
pub const fn is_id_start_byte(c: u8) -> bool {
    matches!(c, b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'$')
}
```

increases perf due to elimination of stack frames, the reason why inline was not possible to offer the same adv was probably because it was being called in a hot path(loop probably)

DaniPopes:
inline(always) doesn't make a difference, in fact probably without inline at all as well, since the function is so small

PR #356, opened 2025-06-13
feat: add PGO optimization to cargo-dist builds

DaniPopes:
Adds Profile-Guided Optimization (PGO) to distributed binaries for improved performance.

The build script automatically detects when PGO can be applied (native builds) and falls back to standard optimized builds for cross-compilation targets. Expected performance improvement is 10-20% for typical workloads.

Supported platforms get PGO optimization:
- x86_64-unknown-linux-gnu
- x86_64-apple-darwin  
- aarch64-apple-darwin
- x86_64-pc-windows-msvc

Cross-compiled platforms continue with standard builds:
- aarch64-unknown-linux-gnu
- x86_64-unknown-linux-musl

### DIFF
--- a/scripts/pgo-build.sh
+++ b/scripts/pgo-build.sh
@@ -1,0 +1,112 @@
+#!/usr/bin/env bash
+# Build script for cargo-dist that applies PGO when possible
+set -euo pipefail
+
+# Color codes for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
+
+# Logging functions
+log_info() {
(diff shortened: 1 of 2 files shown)


scripts/pgo-build.sh
```diff
@@ -0,0 +1,127 @@
...
+fi
+
+BINARY_PATH="target/$BUILD_TARGET/dist/$BINARY_NAME"
+if [ -f "$BINARY_PATH" ]; then
+    log_info "Binary successfully built at: $BINARY_PATH"
+    ls -lh "$BINARY_PATH"
+else
+    log_error "Binary not found at expected location: $BINARY_PATH"
+    exit 1
+fi
```
onbjerg: no reason for this

PR #359, opened 2025-06-16
perf:Optimize Lexer

malik672:
This optimizes `unescape_literal` by creating a fast path for ASCII


malik672:
> Please resolve and rebase, I don't think this is worth it because is_ascii() is O(N) so it goes over every byte twice, but I want to see the codspeed results on latest main



> Please resolve and rebase, I don't think this is worth it because is_ascii() is O(N) so it goes over every byte twice, but I want to see the codspeed results on latest main



> Please resolve and rebase, I don't think this is worth it because is_ascii() is O(N) so it goes over every byte twice, but I want to see the codspeed results on latest main

you are right, closing this

PR #363, opened 2025-06-18
Is digit ascii

malik672:
change `is_ascii_digit` to use ` (c.wrapping_sub(b'0')) < 10` which uses 106 less cycles and 600 less instructions
https://godbolt.org/z/7fY1zc1sY

underneath is_ascii_digit uses `matches!(c, b'0'..=b'9')` ti check whether it's within range or not which is less efficient(depending on context) than `(c.wrapping_sub(b'0')) < 10` and `c >= b'0' && c <= b'9'`

This is not really relevant since `is_digit_ascii` does not show up on the profiler at all any gain will come from `is_id_byte`


onbjerg (changes requested):
there does not seem to be any noticeable perf gain, although this is imo a reduction in readability

where would the perf gain noticeable? can you provide a benchmark?

PR #366, opened 2025-06-22
Refactor: Remove redundant EoF check

malik672:
The is_eof check in the Cursor::eat_while method of the Solidity lexer is redundant, as self.first() returns EOF (b'\0') when input is exhausted, and all predicates  (is_whitespace_byte, is_id_continue_byte) reject EOF, causing the loop to exit naturally. Additionally, advance_token returns RawToken::EOF on empty input, and the Iterator converts RawTokenKind::Eof to None. 

This does not affect perf in anyway


onbjerg (changes requested):
please remove the .ds_store files

DaniPopes (approved):
makes sense even if it doesn't affect perf

PR #371, opened 2025-06-27
chore: fn header spans

0xrusowsky:
## Motivation

necessary for the formatter to properly handle comments between the fn header keywords, like in this example:
```solidity
// function prefix
function withComments( // function name postfix
    // x1 prefix
    uint256 x1, // x1 postfix
    // x2 prefix
    uint256 x2, // x2 postfix
    // x2 postfix2
    /*
        multi-line x3 prefix
    */
    uint256 x3 // x3 postfix
)
    // public prefix
    public // public postfix
    // pure prefix
    pure // pure postfix
    // modifier1 prefix
    modifier1 // modifier1 postfix
    // modifier2 prefix
    modifier2 /*
                mutliline modifier2 postfix
                */
    // modifier3 prefix
    modifier3 // modifier3 postfix
    returns (
        // y1 prefix
        uint256 y1, // y1 postfix
        // y2 prefix
        uint256 y2, // y2 postfix
        // y3 prefix
        uint256 y3
        // y3 postfix
    );
```

## Solution

i added all the spans as new fields under the `FunctionHeader` struct cause i wanted to minimize the impact on the current public API (the only user-facing change is:
```rs
// BEFORE
pub struct FunctionHeader {
    pub virtual_: bool`
}

// NOW
pub struct FunctionHeader {
    pub virtual_: Option<Span>`
}

impl FunctionHeader {
   fn virtual_(&self) -> bool;
}
```

@DaniPopes if you think that the struct it too verbose and u don't mind packing fields, we could use also tuples (or any other suggestion that u may have)


0xrusowsky:
regarding failing tests, if i run `cargo uibless` when i then run `cargo test` i get:
```
The application panicked (crashed).
Message:  tests root directory does not exist: testdata/solidity/test/;
you may need to initialize submodules: `git submodule update --init --checkout`
Location: tools/tester/src/lib.rs:76

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
error: test failed, to rerun pass `-p solar-compiler --test tests`
```

crates/ast/src/ast/item.rs
```diff
@@ -439,17 +439,37 @@ pub struct FunctionHeader<'ast> {
     /// The name of the function.
     /// Only `None` if this is a constructor, fallback, or receive function.
     pub name: Option<Ident>,
-    /// The parameters of the function.
+
+    /// The parameters of the function and its span, which includes the brackets.
     pub parameters: ParameterList<'ast>,
```
DaniPopes, line 467: For this you can make ParameterList a struct itself containing the array and the span

crates/ast/src/ast/item.rs
```diff
@@ -439,17 +439,37 @@ pub struct FunctionHeader<'ast> {
...
+    pub parameters_span: Span,
 
+    /// The visibility keyword and its span.
+    pub visibility_span: Option<Span>,
     pub visibility: Option<Visibility>,
```
DaniPopes: How about creating a `Spanned<T>` struct and making these `Option<Spanned<T>>`?

crates/ast/src/ast/item.rs
```diff
@@ -431,27 +454,59 @@ impl ItemFunction<'_> {
...
 }
 
+impl<'ast> Default for FunctionHeader<'ast> {
+    fn default() -> Self {
+        Self {
+            span: Default::default(),
+            name: Default::default(),
+            parameters: Default::default(),
+            visibility: Default::default(),
+            state_mutability: Spanned { span: Span::DUMMY, data: StateMutability::default() },
```
DaniPopes: we can derive Default for Spanned

crates/ast/src/ast/stmt.rs
```diff
@@ -93,7 +97,7 @@ pub enum StmtKind<'ast> {
     UncheckedBlock(Block<'ast>),
 
     /// A while statement: `while (i < 42) { ... }`.
-    While(Box<'ast, Expr<'ast>>, Box<'ast, Stmt<'ast>>),
+    While(Spanned<Box<'ast, Expr<'ast>>>, Spanned<Box<'ast, Stmt<'ast>>>),
```
DaniPopes: why here? Stmt and Expr are already spanned

crates/interface/src/span.rs
```diff
@@ -227,3 +230,27 @@ impl Span {
         }
     }
 }
+
+/// A value paired with a source code location.
+///
+/// Wraps any value with a [`Span`] to track its location in the source code.
+/// Implements `Deref` and `DerefMut` for transparent access to the inner value.
+#[derive(Clone, Copy, Debug)]
```
DaniPopes: Default

DaniPopes (changes requested):

crates/ast/src/ast/item.rs
0xrusowsky: that's what i initially did, but `Visibility` doesn't impl `Default`

crates/ast/src/ast/stmt.rs
0xrusowsky: i wanted to also encapsulate the parentheses/braces, but i think that you are right and i should have enough info to make it work as is

DaniPopes (approved):

PR #378, opened 2025-07-08
Refactor: Optimize Lexer Error Handling and Add Test for Invalid Characters

malik672:
This pr adds a reduces string allocation and adds a test case for repeated_invalid_characters 

This should be a perf but the repeats branch is never touched on any of my testcase(Picked from solars), what it does is reduce string allocation


onbjerg:
Thank you for contributing, this isn't a big improvement in terms of perf since this is in a cold path, and the readability isn't much better either

PR #379, opened 2025-07-10
chore: store SessionGlobals inside of Session

DaniPopes:
Allows reusing the same `SessionGlobals` across different `enter` points for the same `Session`, which is intuitively what `enter` should do, rather than the current behavior of only sharing the source map and not the interner.


DaniPopes:
`SessionGlobals` initialization was moved from `enter` to `Session`, no actual perf impact in benchmarks

PR #385, opened 2025-07-24
feat(sema): add helper methods to Function

KumaCrypto:
While using HIR, found that some additional helpers could improve the user experience with functions : )

### DIFF
--- a/crates/ast/src/ast/item.rs
+++ b/crates/ast/src/ast/item.rs
@@ -613,7 +613,7 @@ impl DataLocation {
 }
 
 // How a function can mutate the EVM state.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, EnumIs)]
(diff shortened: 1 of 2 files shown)


crates/ast/src/ast/item.rs
```diff
@@ -613,7 +613,7 @@ impl DataLocation {
 }
 
 // How a function can mutate the EVM state.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, EnumIs)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, EnumIs, PartialOrd)]
```
DaniPopes:
```suggestion
#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, EnumIs, PartialOrd, Ord)]
```

PR #389, opened 2025-07-29
chore: spanned data location + subdenomination

0xrusowsky:
# Motivation

these spans are required by the formatter to print comments between the data location and the subdomain keywords:

```solidity
uint256[] /* comment1 */ memory /* comment2 */ sample3; // comment3

value = someVeryVeryVeryLongVariableNameForTheMultiplierForEtherValue
    * 1 /* comment1 */ ether; // comment2
```


0xrusowsky:
sounds good, for the time being i'll add some comments and we can change the tests once we migrate to `fmt-2`

PR #391, opened 2025-07-29
chore: show benches for macbook m2 air

DaniPopes:


benches/README.md
onbjerg, line 1: remember to update versions & bench machine

DaniPopes:
Was just to see the diff, I don't think I want to maintain results for multiple machines every time i want to update them

PR #396, opened 2025-08-12
feat: support `cargo binstall`

onbjerg:
Adds some metadata to support [`cargo binstall`](https://github.com/cargo-bins/cargo-binstall). Currently it is not supported because the crate is named `solar-compiler`, but the binary and binary directory are named `solar`.

Equivalent `cargo binstall` line:

```sh
cargo binstall solar-compiler --bin-dir "solar-{target}/{bin}{binary-ext}" \ 
  --pkg-url "{repo}/releases/download/v{version}/solar-{target}{archive-suffix}"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ You can find a more detailed list in the [pinned GitHub issue](https://github.co
   - [x] Lexing
   - [x] Parsing
   - [ ] Semantic analysis
-    - [x] Symbol resolution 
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):
LGTM! :rocket:

DaniPopes (changes requested):
Actually can you also mention it in the README?

Issue #370, opened 2025-06-25
Provide owned wrappers for AST, HIR, Gcx


DaniPopes:
### Describe the feature

Currently all of these borrow from an arena, which has bad ergonomics in general, and is impossible to store them in a struct to use later. On top of that, they also require entering and staying in the session context for both creating and using them, otherwise the thread local context gets cleared after exiting and the data structures become unusable

PR #397, opened 2025-08-12
feat!: add Compiler

DaniPopes:
Introduce `solar_sema::Compiler`. This is a unified API for the full compiler pipeline, as well as encapsulating the `enter(|| _)` requirement by only allowing access to the context when inside of the closure.

See the HIR example and tests for the usage diff.
Migrations:
- foundry-compilers https://github.com/foundry-rs/compilers/pull/300
- foundry: https://github.com/foundry-rs/foundry/pull/11371

Closes https://github.com/paradigmxyz/solar/issues/370.


examples/src/hir.rs
```diff
@@ -12,24 +12,34 @@ fn main() -> Result<(), EmittedDiagnostics> {
...
+    // Create a new compiler.
+    let mut compiler = Compiler::new(sess);
+
     // Enter the context and parse the file.
     // Counter will be parsed, even if not explicitly provided, since it is a dependency.
-    let _ = sess.enter(|| -> solar::interface::Result<()> {
-        // Set up the parser.
-        let hir_arena = ThreadLocal::new();
-        let mut parsing_context = ParsingContext::new(&sess);
+    let _ = compiler.enter_mut(|compiler| -> solar::interface::Result<()> {
```
onbjerg: can we showcase that you can ret from here by just returning `contracts`?

0xrusowsky: i think it would be cool to showcase how u can enter the compiler the first time, only parse, and then exit + then u can enter again to lower and get the HIR.

this way we would show a complete end-to-end example, with something like:
```rs
#[test]
fn main() -> Result<(), EmittedDiagnostics> {
    let paths = [Path::new("src/AnotherCounter.sol")];

    // Create a new session with a buffer emitter.
    // This is required to capture the emitted diagnostics and to return them at the end.
    let sess = Session::builder().with_buffer_emitter(solar::interface::ColorChoice::Auto).build();

    // Create a new compiler.
    let mut compiler = Compiler::new(sess);

    // Enter the context and parse the file.
    // Counter will be parsed, even if not explicitly provided, since it is a dependency.
    let _ = compiler.enter_mut(|compiler| -> solar::interface::Result<()> {
        // Parse the files.
        let mut parsing_context = compiler.parse();
        parsing_context.load_files(paths)?;
        parsing_context.parse();

        // Inspect the AST.
        let _inspect_the_asts = compiler.gcx().sources.asts();

        Ok(())
    });

    // Showcase that u can freely exit and reenter

    // Enter the context and lower the AST to populate the HIR.
    let _ = compiler.enter_mut(|compiler| -> solar::interface::Result<()> {
        let ControlFlow::Continue(()) = compiler.lower_asts()? else {
            // Can't continue because HIR was not populated,
            // possibly because it was requested in `Session` with `stop_after`.
            return Ok(());
        };

        // Inspect the HIR.
        let gcx = compiler.gcx();
        let mut contracts = gcx.hir.contracts().map(|c| c.name.to_string()).collect::<Vec<_>>();
        contracts.sort(); // No order is guaranteed.
        assert_eq!(contracts, ["AnotherCounter".to_string(), "Counter".to_string()]);

        Ok(())
    });

    // Maybe comment explicitly mentioning that u can and still reenter + from now on, there is no need for `enter_mut` anymore.

    // Return the emitted diagnostics as a `Result<(), _>`.
    // If any errors were emitted, this returns `Err(_)`, otherwise `Ok(())`.
    compiler.sess().emitted_errors().unwrap()
}
```

crates/sema/src/compiler.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /// Enters the compiler context.
+    pub fn enter<T: Send>(&self, f: impl FnOnce(&CompilerRef<'_>) -> T + Send) -> T {
+        self.0.sess.enter(|| f(CompilerRef::new(&self.0)))
+    }
+
+    /// Enters the compiler context with mutable access.
+    ///
+    /// This is only necessary on the first access to parse sources.
```
0xrusowsky: technically speaking, since it is necessary when parsing and when lowering + it is not mandatory to it together, if u don't parse and lower at once, and instead enter twice, you need to `enter_mut` both times

onbjerg (approved):

DaniPopes:
Regression on enter due to extra check for debugging, it shouldn't matter at all.

PR #401, opened 2025-08-15
feat: basic lsp

### LINKED ISSUE #417: Basic LSP lifecycle
### Describe the feature

The foundation for a working LSP is the following, which we should implement:

- Responding to lifecycle events ([`initialize`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize), [`initialized`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialized), [`shutdown`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown), [`exit`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit))
- Capability negotiation between server and client
- An LSP config and basic config discovery (e.g. `solar.toml` and probably also parsing `foundry.toml`)
- Basic text document management, i.e. [`textDocument/didOpen`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen), [`textDocument/didChange`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didChange), [`textDocument/didClose`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didClose).
- Parsing and lowering files when documents change, including in-memory only changes for documents that are managed by the LSP client

### Additional context

_No response_

onbjerg:
Adds a basic LSP exposed as `solar lsp`. The default behavior for `solar` is still to compile.

The LSP itself is implemented as a thin VFS that sits on top of files on disk. It's pretty dumb, so every time a file is changed, the entirety of the project is reparsed and relowered. Diagnostics from this process is published to the client.

The VFS itself makes use of [ropes](https://en.wikipedia.org/wiki/Rope_(data_structure)) for file content.

The structure is heavily inspired by [`rust-analyzer`](https://github.com/rust-lang/rust-analyzer/), albeit with some simplifications given Solar is pretty fast and Solidity projects are usually fairly small.

Closes #417


Cargo.lock
onbjerg, line 1: +400 lines in `Cargo.lock` with 2 dependencies added has to be illegal

onbjerg:
Going to migrate to [`async-lsp`](https://github.com/oxalica/async-lsp) since [`tower-lsp`](https://github.com/ebkalderon/tower-lsp) is both [unmaintained](https://github.com/ebkalderon/tower-lsp/issues/427) and [semantically incorrect](https://github.com/ebkalderon/tower-lsp/issues/284).

Cargo.lock
```diff
@@ -1272,6 +1383,37 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "686f825264d630750a544639377bae737628043f20d38bbc029e8f29ea968a7e"
+dependencies = [
+ "idna_adapter",
+ "smallvec",
+ "utf8_iter",
+]
+
+[[package]]
+name = "idna_adapter"
```
DaniPopes, line 1646: lets pin this in solar-compiler so that our binary pins it but it's not enforced on the library users

onbjerg: done, didn't really affect the lock though

DaniPopes: yeah it was already at 1.1.0 but it would get bumped on a cargo update

crates/lsp/src/proto.rs
```diff
@@ -0,0 +1,30 @@
...
+    };
+    let start = rope.byte_of_utf16_code_unit(start_line + range.start.character as usize);
+    let end_line = if range.end.line > rope.line_len() as u32 {
+        0usize
+    } else {
+        rope.byte_of_line(range.end.line as usize)
+    };
+    let end = rope.byte_of_utf16_code_unit(end_line + range.end.character as usize);
+
+    std::ops::Range { start, end }
```
DaniPopes:
```suggestion
    start..end
```

crates/lsp/src/global_state.rs
```diff
@@ -0,0 +1,60 @@
...
+        }))
+    }
+
+    pub(crate) fn on_initialized(&mut self, _: InitializedParams) -> NotifyResult {
+        self.info_msg("solar initialized".into());
+        ControlFlow::Continue(())
+    }
+}
+
+impl GlobalState {
```
DaniPopes: maybe take in `impl Into<String>`

crates/lsp/src/global_state.rs
```diff
@@ -0,0 +1,60 @@
...
+    ServerInfo, notification as notif,
+};
+use solar_config::version::SHORT_VERSION;
+use solar_interface::data_structures::sync::RwLock;
+
+use crate::{NotifyResult, vfs::Vfs};
+
+pub(crate) struct GlobalState {
+    client: ClientSocket,
+    pub(crate) vfs: Arc<RwLock<Vfs>>,
```
DaniPopes, line 27: doesn't look like this is shared with anything (yet?) will it be? if not then you can drop the lock since we have mut access on handlers

crates/lsp/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    // Lifecycle
+    router
+        .request::<req::Initialize, _>(GlobalState::on_initialize)
+        .notification::<notif::Initialized>(GlobalState::on_initialized)
+        .request::<req::Shutdown, _>(|_, _| std::future::ready(Ok(())))
+        .notification::<notif::Exit>(|_, _| ControlFlow::Break(Ok(())));
+
+    // Notifications
+    router
```
DaniPopes, line 45: why don't we implement the LanguageServer trait?

crates/lsp/src/global_state.rs
onbjerg: it will be shared across tasks

crates/solar/Cargo.toml
```diff
@@ -38,17 +38,23 @@ solar-ast.workspace = true
 solar-config.workspace = true
 solar-data-structures.workspace = true
 solar-interface.workspace = true
+solar-lsp.workspace = true
```
DaniPopes: make optional with "cli" feature

onbjerg:
Hi @mmsaki, it's currently a draft and not entirely complete, so it's expected to not work in all cases. Can you share what project you tried using it in for future reference? I'll make sure it works before merging

khanbilal732 (approved):

0xAlcibiades:
@DaniPopes any interest in getting this updated and into main? Was just looking for a solidity LSP today and came across this; looks interesting.

mmsaki:
@0xAlcibiades I've been waiting on this too for a while too. If you want to try an LSP I am actively building one lmk what you think; try it https://github.com/asyncswap/solidity-language-server

DaniPopes:
#870

PR #414, opened 2025-08-19
feat!: make `Literal`s implement `Copy`

DaniPopes:
Currently literals include `Arc<[u8]>` for strings, and numbers/rationals contain `BigInt` which is `Vec<u32/u64>` internally.

WIP.


tools/tester/src/solc/solidity.rs
```diff
@@ -61,8 +61,10 @@ pub(crate) fn should_skip(path: &Path) -> Result<(), &'static str> {
     #[rustfmt::skip]
     if matches!(
         stem,
-        // Exponent is too large, but apparently it's fine in Solc because the result is 0.
+        // Exponent is too large, but apparently it's fine in Solc because the result is 0 or it get evaluated at compile time.
```
DaniPopes:
```suggestion
        // Exponent is too large, but apparently it's fine in Solc because the result is 0 or it gets evaluated at compile time.
```

PR #428, opened 2025-08-23
perf(sema): better parallel parser scheduling

DaniPopes:
The current algorithm is batched:
1. parse all unparsed sources in parallel, waiting for all to finish
2. resolve all imports
3. add new sources to the list and repeat

Rewrite to be fully recursive:
1. spawn a parse job for all unparsed sources
2. each job resolves imports after parsing, spawning new jobs recursively for new sources
3. top level function exits once all spawned jobs are done thanks to `rayon::scope`

This process has to lock while adding sources to the list in 2. but results it better usage of resources as new files don't have to wait for the current batch to finish before they begin parsing and so on.

This is +~0% to +20%... varies a lot depending on input but always equal or an improvement from what I've measured.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -179,3 +179,5 @@ thread_local = "1.1"
 typed-arena = "2.0"
 unicode-width = "0.2"
 vergen = "8"
+
+[patch.crates-io]


crates/sema/src/parse.rs
```diff
@@ -121,7 +121,7 @@ impl<'gcx> ParsingContext<'gcx> {
         let mut sources = std::mem::take(self.sources);
         if !sources.is_empty() {
             let arenas = self.arenas;
-            if self.sess.is_sequential() {
+            if self.sess.is_sequential() || (sources.len() == 1 && !self.resolve_imports) {
```
onbjerg, line 124: why this change

DaniPopes: Avoiding parallel overhead for simple case of 1 source, this happens in https://github.com/foundry-rs/compilers/pull/300: https://github.com/foundry-rs/compilers/pull/300/files#diff-45071469861c7983c74b8fde1e9e245080da88a98f2ad8ffd6b16e177dbec67fR367-R371

PR #450, opened 2025-08-27
chore!: hide more source map implementation details

DaniPopes:
- hide and rename `StableSourceFileId`
- remove `SourceFileHash` as its unused, can be readded later if needed
- clean up some methods in `SourceMap`
- add `SourceMap::{clear, take}`

### DIFF
--- a/crates/interface/src/source_map/mod.rs
+++ b/crates/interface/src/source_map/mod.rs
@@ -5,7 +5,7 @@ use once_map::OnceMap;
 use solar_data_structures::{
     fmt,
     map::FxBuildHasher,
-    sync::{ReadGuard, RwLock},
+    sync::{RwLock, RwLockReadGuard},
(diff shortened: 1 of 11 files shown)


crates/interface/src/source_map/mod.rs
```diff
@@ -420,6 +412,13 @@ impl SourceMap {
         Ok(FileLines { file: lo.file, lines })
     }
 
+    /// Returns the source file and the range of text corresponding to the given span.
+    ///
+    /// See [`span_to_source`](Self::span_to_source).
+    pub fn span_to_range(&self, sp: Span) -> Result<std::ops::Range<usize>, SpanSnippetError> {
+        self.span_to_source(sp).map(|(_, range)| range)
```
onbjerg, line 419: ooo nice

PR #454, opened 2025-08-28
chore!: move VERSION to config::version::SEMVER_VERSION and log it

DaniPopes:


onbjerg (approved):

onbjerg (commented):
CI is failing because of messed up features

PR #455, opened 2025-08-28
feat: add another utility method for extracting diagnostics

DaniPopes:


crates/interface/src/diagnostics/context.rs
```diff
@@ -213,6 +213,32 @@ impl DiagCtxt {
...
+    /// # }
+    /// ```
+    #[inline]
+    pub fn emitted_diagnostics_result(
+        &self,
+    ) -> Option<Result<EmittedDiagnostics, EmittedDiagnostics>> {
+        let inner = self.inner.lock();
+        let diags = EmittedDiagnostics(inner.emitter.local_buffer()?.to_string());
+        Some(if inner.has_errors() { Err(diags) } else { Ok(diags) })
+    }
```
onbjerg, line 240: to clarify, `Some(Ok(_))` then means it's just warnings and so on?

DaniPopes: yep

PR #464, opened 2025-08-31
feat(parser): implement recursion depth limit for `expr`, `stmt`, and `yul_stmt`

### LINKED ISSUE #460: Implement recursion limit in Parser
### Describe the feature

Implement recursion limit for `Parser` and re-enable these tests: https://github.com/paradigmxyz/solar/blob/ebfcc1d7df9353f3ef2047f009bea4531aac8b62/tools/tester/src/solc/solidity.rs#L45-L47

### Additional context

_No response_

mablr:
## Motivation
Close #460

## Solution
- Add `PARSER_RECURSION_LIMIT` constant (128 depth)
- Track recursion depth in `parse_expr`, `parse_stmt`, and `parse_yul_stmt`
- Return errors with span notes when limit exceeded
- Update test skipping logic since recursion guard is now implemented


crates/parse/src/parser/mod.rs
```diff
@@ -19,6 +19,11 @@ mod stmt;
 mod ty;
 mod yul;
 
+/// Maximum allowed recursive descent depth for selected parser entry points.
+///
+/// This limit is applied to `parse_expr`, `parse_stmt`, and `parse_yul_stmt`.
+const PARSER_RECURSION_LIMIT: usize = 128;
```
mablr, line 25: I have no idea what is the expected limit.

I tried randomly with 256 but it makes `max_depth_reached_4.sol` test successful. Then I reduced the limit to 128 so that this test successfully fails ;)

DaniPopes: solc has 1200 but it's shared across every function in the parser, both are arbitrary so it's fine as long as it's decently high and it doesn't reject valid/reasonable programs

crates/parse/src/parser/mod.rs
```diff
@@ -54,6 +59,13 @@ pub struct Parser<'sess, 'ast> {
     in_yul: bool,
     /// Whether the parser is currently parsing a contract block.
     in_contract: bool,
+
+    /// Current recursion depth for expression parsing via `parse_expr`.
+    expr_recursion_depth: usize,
```
DaniPopes: I think we should also have a single counter that we call with `with_recursion_limit(f: impl FnOnce(&mut Self) -> PResult<T>) -> PResult<T>`, it can also take in a string for the context like "expression"

tools/tester/src/solc/solidity.rs
DaniPopes, line 1: can you please also add a test under tests/ui/parser/ that's just a bunch of blocks and you assert the error? See CONTRIBUTING.md and other tests in the directory on how to write one

crates/parse/src/parser/mod.rs
mablr: I implemented a single shared counter and `with_recursion_limit` method. Is this what you expected ?

DaniPopes (approved):
LGTM, thanks!

DaniPopes (approved):

PR #469, opened 2025-09-01
Test: auto vectorization

malik672:


malik672:
> None of the benchmarks have changed at all, so this seems to give us no perf gain, but it does lessen readability. Can you please provide a benchmark that shows this improves performance? Single runs of a profiler do not show performance speedups or slowdowns, and the numbers on your screenshots look to be within expected variance.

You are right but Instruments is very reliable and each "run" aggregates thousands of samples over the period, so it's inherently averaging out variance and Godbolt `LLVM MCA` cycles will hardly ever show incorrect stuff

PR #470, opened 2025-09-01
feat(ast): naive natspec

0xrusowsky:
## Motivation

there is this foundry issue to use a 3rd-party library lint natspec comments:
- https://github.com/foundry-rs/foundry/issues/11471

right now solar doesn't support natspec, which means that the natspec lints that this 3rd-party supports can't be directly implemented on `forge-lint`.

so i thought i'd be a good opportunity to introduce an initial version.

## Solution

this proposal adds the required AST types and implements a minimal (naive) parser, which could be then used by the linter to achieve what community asked for in the linked ticket.

### Notes

- i tried implementing it in a backwards-compatible way, although i believe that the only consumer for the existig `DocComments` may be the wip formatter.

- obviously this feature has never been a priority, as it's never been in the roadmap + it could have a negative impact on performance and increase the size of the generated AST, so we could gate it under a feature flag.

- yul tests are currently failing due to the usage of "non-std" tags. I understand what they do but i wasn't aware of them nor saw them in the official soliditylang spec, so i didn't add support for them (but we could if u think we should)

### DIFF
--- a/crates/parse/src/parser/mod.rs
+++ b/crates/parse/src/parser/mod.rs
@@ -1235,19 +1235,19 @@ mod tests {
                 check_natspec_item(sm, natspec_items[i].0, natspec_items[i].1, snip, kind, name, content)
             };
 
-            check(0, "@title", "title", None, Some("MyContract"));
-            check(1, "@author", "author", None, Some("Alice"));
-            check(2, "@notice", "notice", None, Some("This is a notice"));
+            check(0, "title", "title", None, Some("MyContract"));
+            check(1, "author", "author", None, Some("Alice"));
+            check(2, "notice", "notice", None, Some("This is a notice"));
             let span3 = sm.span_to_snippet(natspec_items[3].1.span).unwrap();
             check(3, &span3, "notice", None, Some("that spans multiple lines"));
             let span4 = sm.span_to_snippet(natspec_items[4].1.span).unwrap();
             check(4, &span4, "notice", None, Some("and continues here"));
-            check(5, "@dev", "dev", None, Some("This is dev documentation"));
-            check(6, "@param", "param", Some("x"), Some("The input parameter"));
-            check(7, "@return", "return", Some("result"), Some("The return value"));
-            check(8, "@inheritdoc", "inheritdoc", Some("BaseContract"), Some(""));
-            check(9, "@custom:security", "custom", Some("security"), Some("High priority"));
-            check(10, "@solidity", "internal", Some("solidity"), Some("memory-safe"));
+            check(5, "dev", "dev", None, Some("This is dev documentation"));
+            check(6, "param", "param", Some("x"), Some("The input parameter"));
+            check(7, "return", "return", Some("result"), Some("The return value"));
+            check(8, "inheritdoc", "inheritdoc", Some("BaseContract"), Some(""));
+            check(9, "custom:security", "custom", Some("security"), Some("High priority"));
+            check(10, "solidity", "internal", Some("solidity"), Some("memory-safe"));
 
             assert_eq!(sm.span_to_snippet(docs.span()).unwrap(), "/// @title MyContract\n/// @author Alice\n/// @notice This is a notice\n/// that spans multiple lines\n/// and continues here\n/// @dev This is dev documentation\n/// @param x The input parameter\n/// @return result The return value\n/// @inheritdoc BaseContract\n/// @custom:security High priority\n/// @solidity memory-safe");
         });
@@ -1290,15 +1290,15 @@ mod tests {
                 check_natspec_item(sm, sym, &items[i], span, kind, name, content)
             };
 
-            check(0, "@title", "title", None, Some("MyContract"));
-            check(1, "@author", "author", None, Some("Alice"));
-            check(2, "@notice", "notice", None, Some("This is a notice\n * that spans multiple lines\n * and continues here"));
-            check(3, "@dev", "dev", None, Some("This is dev documentation"));
-            check(4, "@param", "param", Some("x"), Some("The input parameter"));
-            check(5, "@return", "return", Some("result"), Some("The return value"));
-            check(6, "@inheritdoc", "inheritdoc", Some("BaseContract"), Some(""));
-            check(7, "@custom:security", "custom", Some("security"), Some("High priority"));
-            check(8, "@src", "internal", Some("src"), Some("0:123:456"));
+            check(0, "title", "title", None, Some("MyContract"));
+            check(1, "author", "author", None, Some("Alice"));
+            check(2, "notice", "notice", None, Some("This is a notice\n * that spans multiple lines\n * and continues here"));
+            check(3, "dev", "dev", None, Some("This is dev documentation"));
+            check(4, "param", "param", Some("x"), Some("The input parameter"));
+            check(5, "return", "return", Some("result"), Some("The return value"));
+            check(6, "inheritdoc", "inheritdoc", Some("BaseContract"), Some(""));
+            check(7, "custom:security", "custom", Some("security"), Some("High priority"));
+            check(8, "src", "internal", Some("src"), Some("0:123:456"));
 
             assert_eq!(sm.span_to_snippet(docs.span()).unwrap(), "/**\n * @title MyContract\n * @author Alice\n * @notice This is a notice\n * that spans multiple lines\n * and continues here\n * @dev This is dev documentation\n * @param x The input parameter\n * @return result The return value\n * @inheritdoc BaseContract\n * @custom:security High priority\n * @src 0:123:456\n */");
         });


DaniPopes:
i think this is fine to support in the parser. we'll want something eventually for @\solidity / @\solar anyway, and flattening a project wants to update inheritdoc tags

for performance this doesn't really matter, feature flag is not needed

for reporting errors, i'd check with solc how they do natspec parsing and if they report a regular parser error if not respected

0xrusowsky:
@beeb feel free to contribute to the repo, but in our opinion it doesn't make sense to introduce a parsing dependency in a compiler that is being built from teh ground up

beeb:
Alright, I can understand, no worries!

crates/ast/src/ast/mod.rs
```diff
@@ -123,6 +130,58 @@ impl DocComments<'_> {
...
+    Return { tag: Ident },
+    Inheritdoc { tag: Ident },
+    Custom { tag: Ident },
+
+    // Special tags reserved for internal purposes.
+    Internal { tag: Ident },
+}
+
+/// Internal tags.
+pub const INTERNAL_TAGS: [&str; 4] = ["solidity", "src", "use-src", "ast-id"];
```
0xrusowsky: @DaniPopes looks like solc has the following behavior:
- they use internal tags for yul
- they silently ignore unknown tags in yul (don't error)

so i've replicated it

https://github.com/argotorg/solidity/blob/2ca5fb3b6adcb1a8fb2c0904fb37526121cf2c72/libyul/AsmParser.cpp#L151

0xrusowsky:
uff, the perf regression is real...

will try to optimize it

crates/ast/src/ast/mod.rs
```diff
@@ -85,34 +85,41 @@ impl std::ops::Deref for Arena {
 
 /// A list of doc-comments.
 #[derive(Default)]
-pub struct DocComments<'ast>(pub BoxSlice<'ast, DocComment>);
+pub struct DocComments<'ast> {
+    /// The raw doc comments.
+    pub comments: BoxSlice<'ast, DocComment>,
+    /// The parsed Natspec, if it exists.
+    pub natspec: Option<NatSpec<'ast>>,
```
DaniPopes: can't we merge this inside of DocComment itself? we want to avoid making DocComments bigger and i think coupling makes more sense

crates/ast/src/ast/mod.rs
```diff
@@ -174,8 +174,14 @@ pub enum NatSpecKind {
     Return { tag: Ident },
     Inheritdoc { tag: Ident },
     Custom { tag: Ident },
+
+    // Special tags reserved for internal purposes.
+    Internal { tag: Ident },
 }
 
+/// Internal tags.
+pub const INTERNAL_TAGS: [&str; 4] = ["solidity", "src", "use-src", "ast-id"];
```
DaniPopes: prefix with NATSPEC_ and use &[&str]

crates/ast/src/ast/mod.rs
DaniPopes: maybe move to mod natspec?

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+        }
+
         self.docs.clear();
-        docs.into()
+        self.alloc_smallvec(parsed_comments).into()
+    }
+
+    /// Parses NatSpec items from a single doc comment.
+    #[cold]
+    fn parse_natspec_for_comment(
```
DaniPopes: this doesnt need to be cold

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+        comment: &RawDocComment,
+        in_yul: bool,
+        dcx: &DiagCtxt,
+    ) -> SmallVec<[ast::NatSpecItem; 6]> {
+        let mut items = SmallVec::<[ast::NatSpecItem; 6]>::new();
+
```
DaniPopes: you can bail early if the string doesn't contain '@'

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+                let (tag, rest) =
+                    tag_line.split_once(char::is_whitespace).unwrap_or((tag_line, ""));
+
+                // Calculate span: from '@' to end of tag name.
+                let tag_hi = comment.span.lo().0 + PREFIX_BYTES + acc_bytes + tag_offset as u32;
+                let tag_lo = tag_hi + tag.len() as u32 + 1; // +1 for '@'
+                span = Some(Span::new(BytePos(tag_hi), BytePos(tag_lo)));
+
+                let tag_symbol = Symbol::intern(tag);
```
DaniPopes: dont need to intern, can match on the str and only intern what we store

crates/ast/src/ast/natspec.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Clone, Copy, Debug)]
+pub enum NatSpecKind {
+    Title,
+    Author,
+    Notice,
+    Dev,
+    Param { tag: Ident },
+    Return { tag: Ident },
+    Inheritdoc { tag: Ident },
+    Custom { tag: Ident },
```
DaniPopes: what do the `tag`s here mean? please also doc comment with the corresponding tag like ``/// `@title` ``

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+            if let Some(k) = kind.take() {
+                items.push(ast::NatSpecItem { span: span.take().unwrap(), kind: k });
+            }
+        }
+
+        // Line comments: '///', Block comments: '/**'.
+        const PREFIX_BYTES: u32 = 3;
+        let mut acc_bytes = 0u32;
+
+        for line in comment.symbol.as_str().lines() {
```
DaniPopes: memchr_iter('\n') should be faster

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
 
     #[cold]
     fn parse_doc_comments_inner(&mut self) -> DocComments<'ast> {
-        let docs = self.arena.alloc_thin_slice_copy((), &self.docs);
+        let mut parsed_comments = SmallVec::<[DocComment<'ast>; 6]>::new();
```
DaniPopes: early return if self.docs is empty

crates/parse/src/parser/mod.rs
```diff
@@ -868,9 +870,105 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+    }
+
+    /// Parses NatSpec items from a single doc comment.
+    #[cold]
+    fn parse_natspec_for_comment(
+        comment: &RawDocComment,
+        in_yul: bool,
+        dcx: &DiagCtxt,
+    ) -> SmallVec<[ast::NatSpecItem; 6]> {
+        let mut items = SmallVec::<[ast::NatSpecItem; 6]>::new();
```
DaniPopes: please leave a comment with reference to the solc code

0xrusowsky:
@DaniPopes after all the perf improvements we went down from an up-to ~13% regression (with 6 repos showing regressions) to "just" up-to 6.4% (in just one repo)

do you think it's acceptable? what else could i improve?

crates/ast/src/ast/natspec.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// An annotation for end-users.
+    Notice,
+    /// `@dev`
+    ///
+    /// A technical annotation for developers.
+    Dev,
+    /// `@param <name>`
+    ///
+    /// Documents a parameter. The `tag` field contains the parameter name.
+    Param { tag: Ident },
```
DaniPopes: we can give these `tag` better names like `name`, `contract` like in the docs

crates/ast/src/ast/natspec.rs
```diff
@@ -0,0 +1,86 @@
...
+    Internal { tag: Ident },
+}
+
+/// Internal natspec tags.
+pub const NATSPEC_INTERNAL_TAGS: &[&str] = &["solidity", "src", "use-src", "ast-id"];
+
+/// Raw doc-comment data without parsed NatSpec.
+/// Used for temporary storage before parsing.
+#[derive(Debug, Clone, Copy)]
+pub struct RawDocComment {
```
DaniPopes: I think it's fine if we have a single DocComment and keeping the `natspec` defaulted. this lets us avoid extra allocation for `self.docs`

crates/parse/src/parser/mod.rs
```diff
@@ -1007,3 +1028,267 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+            flush_item(&mut items, &mut kind, &mut span);
+
+            let (tag, rest_start) = split_once_ws(content, bytes, tag_start, line_end);
+
+            // Calculate span: from '@' to end of tag name.
+            let tag_lo = comment.span.lo().0 + PREFIX_BYTES + (line_start + tag_offset) as u32;
+            let tag_hi = tag_lo + tag.len() as u32 + 1; // +1 for '@'
+            span = Some(Span::new(BytePos(tag_lo), BytePos(tag_hi)));
+
+            kind = Some(match tag {
```
DaniPopes, line 1101: i wonder if we should leave an offset/len pair for consumers to quickly find the relevant content of the tag? for example if i have `@title abcd 123` i want to get the actual contents of the natspec up to the new line right? so i would want something that i can index into `doc.symbol.as_str()[...]`

in this case we wont need to intern any contents

0xrusowsky: i initially discarded the approach that u propose because of its added parsing overhead, but it would definitely provide a better devex.

however, to do it properly, we first need to discuss the design challenge that multiline comments represent... while this is not a problem for block comments, our lexer currently treats this as 3 separate single line comments:
```solidity
/// @param x The first parameter
/// that continues here
/// and continues more
```

which means that unless we want to potentially output incomplete content, we would have to modify the lexer to merge these together (which kinda breaks the semantics of `CommentKind::Line`, and would probably require some minor `forge fmt` changes)

it looks like solc merges consecutive line doc cmnts [Scanner::scanSingleLineDocComment()](https://github.com/argotorg/solidity/blob/2ca5fb3b6adcb1a8fb2c0904fb37526121cf2c72/liblangutil/Scanner.cpp#L348-L389)

wdyt?

beeb: Quick note: the [spec](https://docs.soliditylang.org/en/latest/natspec-format.html#tags) specifies that comments without an `@` at the beginning default to being considered as `@notice`. In the exemple you gave, the second line should be parsed as `@notice that continues here` according to the spec.

> As a special case, if no tags are used then the Solidity compiler will interpret a /// or /** comment in the same way as if it were tagged with `@notice`.

beeb: Although one could argue that we could apply this rule to a block of `///` comments only if there is no `@` prefix for any of them, and otherwise default to the last seen `@` prefix.

0xrusowsky: personally i interpreted it as the latter, but now that you mention it... the former also makes sense and would make it easier haha

will proactively do that and if @DaniPopes believes it should be the other way around, we can discuss how to proceed and whether or not the lexer should be modified

PS: parsing perf was degraded a little bit because of that, but i think the delta is acceptable compared to the total overhead that natspec already introduced

beeb: It probably doesn't align with user expectation, especially when it comes to `forge doc`, but that's the spec... Whether we should follow it or not is another matter XD

crates/parse/src/parser/mod.rs
```diff
@@ -148,7 +148,7 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
             prev_token: Token::DUMMY,
             expected_tokens: Vec::with_capacity(8),
             last_unexpected_token_span: None,
-            docs: Vec::with_capacity(4),
+            docs: SmallVec::with_capacity(4),
```
DaniPopes: this should be left as a vec

crates/parse/src/parser/mod.rs
```diff
@@ -733,7 +733,6 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
         }
         self.prev_token = std::mem::replace(&mut self.token, next);
         self.expected_tokens.clear();
-        self.docs.clear();
```
DaniPopes, line 736: I don't removing this is right, if parse_docs is not called then docs are accumulated to the next item that does

```
/// a
x;
/// b
uint y = 0;
```

the statement will get the `a` doc when it shouldnt

crates/parse/src/parser/mod.rs
```diff
@@ -863,14 +870,12 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
     pub fn parse_doc_comments(&mut self) -> DocComments<'ast> {
-        if !self.docs.is_empty() { self.parse_doc_comments_inner() } else { Default::default() }
-    }
-
-    #[cold]
-    fn parse_doc_comments_inner(&mut self) -> DocComments<'ast> {
-        let docs = self.arena.alloc_thin_slice_copy((), &self.docs);
-        self.docs.clear();
-        docs.into()
+        if !self.docs.is_empty() {
```
DaniPopes: please revert these changes, they're not equivalent and you're adding overhead to all callers

crates/parse/src/parser/mod.rs
```diff
@@ -1007,3 +1012,368 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+/// Splits a string slice at the first whitespace character using the `memchr` crate.
+/// Returns the content up to the whitespace and the position of the first following non-blank char.
+#[inline]
+fn split_once_ws<'a>(
+    content: &'a str,
+    bytes: &'a [u8],
+    start: usize,
+    end: usize,
+) -> (&'a str, usize) {
+    let ws_pos = memchr::memchr3(b' ', b'\t', b'\r', &bytes[start..end])
```
DaniPopes: you can use trim_ascii_start here, its likely that there isn't more than a single space around tags

DaniPopes:
I think this is fine for now, merging doccomments would require more changes

crates/parse/src/parser/mod.rs
0xrusowsky: [`9b54b9f` (#470)](https://github.com/paradigmxyz/solar/pull/470/commits/9b54b9f5173c7e51c7083927142203954cd5f99d)

crates/parse/src/parser/mod.rs
0xrusowsky: unsure why u commented on L1152 rather than on L1157-L1161, as i use/need `memchr3` to find the length of the tag.

here the commit to use `trim_ascii_start` as requested: [`566e3a1`(#470)](https://github.com/paradigmxyz/solar/pull/470/commits/566e3a100f59eba3916f6de3cc23e2487f5d7933)

DaniPopes: len_before = len();
trim;
ws_pos = len_before - len()

no?

0xrusowsky: but the input to this fn is `tag content..`, so for example for `/// @param foo Foo explanation`:
- fn input: `"param foo Foo explanation"`
- fn output: `("param", 6)`

crates/parse/src/parser/mod.rs
```diff
@@ -1007,3 +1017,362 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+    let mut prev_line_end = 0;
+    for line_end in memchr::memchr_iter(b'\n', bytes).chain(std::iter::once(bytes.len())) {
+        if let Some(tag_offset) = memchr::memchr(b'@', &bytes[line_start..line_end]) {
+            let tag_start = line_start + tag_offset + 1;
+            flush_item(&mut items, &mut kind, &mut span, content_start, prev_line_end);
+
+            let (tag, rest_start) = split_once_ws(content, bytes, tag_start, line_end);
+
+            // Calculate span: from '@' to end of tag name.
+            let tag_lo = comment_span.lo().0 + PREFIX_BYTES + (line_start + tag_offset) as u32;
```
DaniPopes: can you add a doccomment in the span visitor test to verify this? tests/ui/parser/span_visitor.sol

0xrusowsky: i was including the '@' by mistake... fixed + tested [`6fe920e` (#470)](https://github.com/paradigmxyz/solar/pull/470/commits/6fe920e74a01e3a196eb6daa1d3a349a57c6dc90)

0xrusowsky: if the test is too verbose and u prefer less cmnts lmk and i'll leave just 1

PR #472, opened 2025-09-02
fix(sema): handle linearization failure fallout

DaniPopes:
See https://github.com/foundry-rs/foundry/issues/11504.


crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -443,6 +443,10 @@ impl<'gcx> ResolveContext<'gcx> {
             );
             let contract = self.hir.contract(c_id);
 
+            if contract.linearization_failed() {
+                continue;
+            }
```
mattsse, line 448: ah I see, this simply skips it preventing 0 - 1 -> max alloc

Issue #466, opened 2025-08-31
Implement diagnostic suggestions


DaniPopes:
### Describe the feature

`annotate-snippets 0.12.0` (https://github.com/paradigmxyz/solar/pull/465) adds support for patches to be displayed nicely. Implement `Diag.suggestions` from rustc with all the relevant methods in `Diag`, `DiagBuilder`, `DiagCtxt`, human and JSON emitter.

### Additional context

Rustc reference:
- inline simple suggestion into a `help:` label https://github.com/rust-lang/rust/blob/f73bcd50a433b7c17225ca98b44064236e7e84b3/compiler/rustc_errors/src/emitter.rs#L227
- rustc json emitter https://github.com/rust-lang/rust/blob/f73bcd50a433b7c17225ca98b44064236e7e84b3/compiler/rustc_errors/src/json.rs#L189

0xrusowsky:
yes, please assign it to me 🙌

PR #474, opened 2025-09-03
feat!: diagnostic suggestions

0xrusowsky:
ref:
- https://github.com/paradigmxyz/solar/issues/466

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -449,8 +449,13 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
                 if let Some(ret_name) = ret.name {
                     self.dcx()
                         .err("return parameters in function types may not be named")
-                        .span(ret.span)
-                        .span_help(ret_name.span, format!("remove `{ret_name}`"))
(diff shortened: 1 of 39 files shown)


0xrusowsky:
@DaniPopes i tried to do a more performant version of the impl, but i am not sure if it is worth it or not, so i did it on a diff branch.

here you can see the diff, and if u think it's worth it i can merge it: https://github.com/0xrusowsky/solar/pull/2

DaniPopes (commented):
can you replace a few "help" notes in this repo with a suggestion so we can also test it here?

crates/sema/src/ast_passes.rs
```diff
@@ -449,8 +449,13 @@ impl<'ast> Visit<'ast> for AstValidator<'_, 'ast> {
                 if let Some(ret_name) = ret.name {
                     self.dcx()
                         .err("return parameters in function types may not be named")
-                        .span(ret.span)
-                        .span_help(ret_name.span, format!("remove `{ret_name}`"))
+                        .span(ret_name.span)
```
DaniPopes: revert span

PR #476, opened 2025-09-04
perf: use SIMD for lexer 

malik672:
This PR implements bulk processing optimizations for the Solar lexer,
 
 from benchmark works less efficiently on files with small sol files.
 Obviously this could be made way better

### DIFF
--- a/crates/parse/src/lexer/char_class_table.rs
+++ b/crates/parse/src/lexer/char_class_table.rs
@@ -1,0 +1,191 @@
+//! Character classification lookup table for SIMD lexing.
+//!
+//! This module provides a pre-computed lookup table that classifies each ASCII character
+//! into categories useful for lexical analysis. Instead of using multiple branches per
+//! character, we can use a single table lookup or SIMD shuffle operation.
+
+/// Character classification flags
+const WHITESPACE: u8 = 0b0000_0001;
+const ID_START: u8   = 0b0000_0010;
+const ID_CONTINUE: u8= 0b0000_0100;
+const DIGIT: u8      = 0b0000_1000;
+const HEX_DIGIT: u8  = 0b0001_0000;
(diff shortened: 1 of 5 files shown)


test_simd.rs
```diff
@@ -0,0 +1,35 @@
+use solar_parse::Lexer;
```
onbjerg, line 1: this is not really useful as it replicates existing tests but less complete

crates/parse/src/lexer/char_class_table.rs
```diff
@@ -0,0 +1,191 @@
...
+pub(super) const fn is_digit_fast(c: u8) -> bool {
+    (classify_byte(c) & DIGIT) != 0
+}
+
+/// Check if character is a hexadecimal digit using lookup table.
+#[allow(dead_code)]
+#[inline]
+pub(super) const fn is_hex_digit_fast(c: u8) -> bool {
+    (classify_byte(c) & HEX_DIGIT) != 0
+}
```
onbjerg, line 110: no point in adding these if they are unused

crates/parse/src/lexer/char_class_table.rs
```diff
@@ -0,0 +1,191 @@
...
+#[inline]
+const fn classify_byte(c: u8) -> u8 {
+    CHAR_CLASS_TABLE[c as usize]
+}
+
+/// Check if character is whitespace using lookup table.
+#[inline]
+pub(super) const fn is_whitespace_fast(c: u8) -> bool {
+    (classify_byte(c) & WHITESPACE) != 0
+}
```
onbjerg, line 83: I doubt this is faster since this destroys caches and the entire lookup table above was constructed just to check for whitespaces

onbjerg (changes requested):
-7% is not a hit we are willing to take, especially not on smaller files as most Solidity files are in fact very small.

PR #479, opened 2025-09-05
refactor(lexer): Inline token glueing into Cursor

### LINKED ISSUE #477: Inline token `glue`ing into `Cursor`
Currently we have 2 stages in the lexer, Cursor and Lexer. Cursor handles reading bytes and generating RawTokens, and Lexer handles "cooking" them by interning strings and creating Spans.

Lexer also has an outer loop that calls `bump` on itself and glues tokens together when possible when there is no whitespace between them. This is unnecessary and can be handled directly in the Cursor. https://github.com/paradigmxyz/solar/blob/ebfcc1d7df9353f3ef2047f009bea4531aac8b62/crates/parse/src/lexer/mod.rs#L117

We should be able to remove `Token::glue` by instead handling all the token combinations inside of `Cursor` here: https://github.com/paradigmxyz/solar/blob/04b8466d1b413674364c3140a22fb5541591b656/crates/parse/src/lexer/cursor/mod.rs#L134

This should simplify the lexer, improve performance a bit, and make Cursor handle all tokens instead of just 1-character ones

mablr:
- Updated `next_token` method in `Lexer` to directly use the result of `bump()` without attempting to glue tokens.
- Added new raw token kinds for various operators: `EqEq`, `Le`, `Ge`...
- Updated `Cursor` to handle multi-character operator tokens and their corresponding raw token kinds.
- Removed the `glue` methods from `TokenKind` and `Token` impl (no longer needed)

Closes #477

### DIFF
--- a/crates/parse/src/lexer/cursor/mod.rs
+++ b/crates/parse/src/lexer/cursor/mod.rs
@@ -137,6 +137,10 @@ impl<'a> Cursor<'a> {
             b'/' => match self.first() {
                 b'/' => self.line_comment(),
                 b'*' => self.block_comment(),
+                b'=' => {
+                    self.bump();
+                    RawTokenKind::BinOpEq(solar_ast::token::BinOpToken::Slash)
+                }
                 _ => RawTokenKind::Slash,
             },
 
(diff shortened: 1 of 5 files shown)


crates/parse/src/lexer/cursor/mod.rs
```diff
@@ -168,18 +172,150 @@ impl<'a> Cursor<'a> {
...
+            }
+            b'>' => {
+                match self.first() {
+                    b'=' => {
+                        self.bump();
+                        RawTokenKind::Ge
+                    }
+                    b'>' => {
+                        self.bump();
+                        if self.first() == b'>' {
```
DaniPopes: this one can be a match too

crates/parse/src/lexer/cursor/mod.rs
```diff
@@ -168,18 +172,150 @@ impl<'a> Cursor<'a> {
             b']' => RawTokenKind::CloseBracket,
```
DaniPopes: comment on this list should be updated

crates/parse/src/lexer/mod.rs
```diff
@@ -113,18 +113,7 @@ impl<'sess, 'src> Lexer<'sess, 'src> {
...
-            (next_token, preceded_by_whitespace) = self.bump();
-            if preceded_by_whitespace {
-                break;
-            } else if let Some(glued) = self.token.glue(next_token) {
-                self.token = glued;
-            } else {
-                break;
-            }
-        }
+        let (next_token, _) = self.bump();
```
DaniPopes: We can remove preceded_by_whitespace entirely, and possibly `self.token` too, and so at that point `bump` can become `next_token`

mablr: That would be actually way cleaner 👍

crates/parse/src/lexer/mod.rs
```diff
@@ -203,6 +175,21 @@ impl<'sess, 'src> Lexer<'sess, 'src> {
                 RawTokenKind::Caret => TokenKind::BinOp(BinOpToken::Caret),
                 RawTokenKind::Percent => TokenKind::BinOp(BinOpToken::Percent),
```
DaniPopes: we should consolidate rawtokenkind::* that map to binop to just rawtokenkind::binop now that we're here

crates/parse/src/lexer/mod.rs
```diff
@@ -39,9 +39,6 @@ pub struct Lexer<'sess, 'src> {
     /// Cursor for getting lexer tokens.
```
DaniPopes: the "concatenating tokens together" should be removed from the struct's docs

DaniPopes (approved):
Thanks!

PR #498, opened 2025-09-12
feat: configurable logging destination

onbjerg:
Allow the tracing output to be written to either stdout or stderr, defaulting to stdout. Can be extended with support for files later, if desired.

Useful for #401 to get more detailed output

### DIFF
--- a/crates/cli/src/utils.rs
+++ b/crates/cli/src/utils.rs
@@ -1,8 +1,10 @@
 //! Utility functions used by the Solar CLI.
 
-use std::io::{self};
-
 use solar_interface::diagnostics::DiagCtxt;
+use std::io;
+
+#[cfg(feature = "tracing")]
+use solar_sema::ast::Either;
 
 #[cfg(feature = "mimalloc")]
 use mimalloc as _;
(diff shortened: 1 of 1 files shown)


crates/cli/src/utils.rs
```diff
@@ -43,9 +45,56 @@ pub const fn new_allocator() -> Allocator {
...
+#[derive(Default)]
+pub enum LogDestination {
+    #[default]
+    Stdout,
+    Stderr,
+}
+
+#[cfg(feature = "tracing")]
+impl<'a> tracing_subscriber::fmt::MakeWriter<'a> for LogDestination {
+    type Writer = LogWriter<'a>;
```
DaniPopes: i dont mind having box dyn write here

crates/cli/src/utils.rs
```diff
@@ -43,9 +45,56 @@ pub const fn new_allocator() -> Allocator {
...
+
+    fn make_writer(&'a self) -> Self::Writer {
+        match self {
+            Self::Stdout => LogWriter::Stdout(std::io::stdout().lock()),
```
DaniPopes: idk if we want to lock here, does tracing create it on demand for every log?

onbjerg: idk, if we use `Stdout` directly, any call to `Write` will also lock internally: https://doc.rust-lang.org/stable/src/std/io/stdio.rs.html#813-836

crates/cli/src/utils.rs
onbjerg: https://github.com/paradigmxyz/solar/compare/4a1bfb020fe7f536af5eb044642497ddb6463bb9..e993cf4acf88efa3a00996d759f7b64c723ee90c

crates/cli/src/utils.rs
DaniPopes: yes but if it creates a writer at the beginning this will lock stdout forever

onbjerg: it's per event https://github.com/tokio-rs/tracing/blob/c297a37096ae1562d453624984aa5f924ab490a1/tracing-subscriber/src/fmt/fmt_layer.rs#L1003

DaniPopes (approved):

PR #502, opened 2025-09-13
perf(lexer): use lookup tables for char info

DaniPopes:
Turns out lookup table is faster due to less instructions and branches, which outweighs the hit from having one in the first place. CC @malik672


malik672:
> Turns out lookup table is faster due to less instructions and branches, which outweighs the hit from having one in the first place. CC @malik672

Ohh the lookup table optimization, never knew the increase will be this big tbh.

onbjerg (approved):

PR #503, opened 2025-09-13
Enable AVX2+FMA for benchmark builds

### LINKED ISSUE #494: Enable modern SIMD features for per-platform builds
### Describe the feature

suggestions for additional target features:
x86_64:
+avx2, +fma: avoids the Intel execution issues @onbjerg mentioned
Covers 95%+ of deployed x86_64 hardware

aarch64:
+neon - Standard baseline for ARM64, every ARM64 chip has it

@onbjerg @DaniPopes

malik672:
Fixes #494

Enables modern SIMD features for benchmark builds

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -12,6 +12,8 @@ env:
 jobs:
   codspeed:
     runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: "-C target-feature=+avx2,+fma"
     steps:
       - uses: actions/checkout@v5
         with:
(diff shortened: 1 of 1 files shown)


onbjerg:
We could maybe just do `-C target-cpu=haswell`? It's from 2013, and is pretty broadly supported. If we want to be a bit more conservative, we could also do `sandybridge` if we want more compatibility, but it's older (2011)

malik672:
done @onbjerg @DaniPopes

malik672:
> Hey @malik672, thanks for taking the time on this, however we don't see any noticeable perf gains here. We think going forward we're just going to do runtime detection of simd features and enable when appropriate, which keeps portability. The cost of checking this at runtime is pretty small.

I'd be happy to work on NEON support for ARM64 as part of the runtime detection implementation if that would be helpful ?

PR #510, opened 2025-09-13
perf(parser): pass RawToken in registers

DaniPopes:
Same as https://github.com/paradigmxyz/solar/pull/509 for `RawToken`.


DaniPopes:
Actually the compiler does do this already for RawToken :thinking: so this change is a noop

PR #517, opened 2025-09-14
feat(parser): introduce `Recovered` enum to improve code readability

### LINKED ISSUE #514: Replace `bool /*  recovered */` with enum Recovered in Parser
The parser has a lot of places that use an anonymous `bool` argument/return value to indicate whether an operation is in an error path but recovered and is continuing anyway. We can use the "newtype pattern", creating enum Recovered to make the code easier to read.

mablr:
Closes #514

### DIFF
--- a/crates/parse/src/parser/mod.rs
+++ b/crates/parse/src/parser/mod.rs
@@ -132,6 +132,13 @@ impl SeqSep {
     }
 }
 
+/// Indicates whether the parser took a recovery path and continued.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Recovered {
+    No,
+    Yes,
+}
+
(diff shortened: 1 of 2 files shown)


crates/parse/src/parser/mod.rs
```diff
@@ -132,6 +132,13 @@ impl SeqSep {
     }
 }
 
+/// Indicates whether the parser took a recovery path and continued.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Recovered {
+    No,
+    Yes,
+}
```
mablr, line 140: I hesitated to place it in `parse/src/lib.rs` along with error type ...

DaniPopes (approved):

PR #521, opened 2025-09-16
bench: lowering benchmarks

### LINKED ISSUE #520: Add lowering benchmarks
### Describe the feature

Parsers can now also lower sources since #519, but we still need to implement `Parser::lower` for `Solar`. We should ensure that all sources marked as `Capabilites::all()` can successfully be lowered, otherwise demote them to `Capabilites::parse_and_lex()`.

### Additional context

_No response_

onbjerg:
Note that the performance regressions are not real in the sense that the previous benchmarks would allocate on the first benchmark iteration and then just do lookups in the subsequent ones, whereas the benchmarks now get a full new session on each iteration, which means we don't reuse anything from previous iterations.
Closes #520

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2598,6 +2598,7 @@ dependencies = [
  "slang_solidity",
  "solang-parser",
  "solar-parse",
+ "solar-sema",
  "tree-sitter",
  "tree-sitter-solidity",
 ]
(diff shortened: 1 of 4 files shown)


DaniPopes:
Regressions are due to creating a new session for every benchmark run instead of once per benchmark. This is fine because it's actually more realistic since e.g. before we would allocate all the symbols on the first run and then on subsequent runs it would only need to lookup, which is not what happens in a realistic scenario

PR #538, opened 2025-09-25
fix(diagnostics): render footer below suggestions

0xrusowsky:
### Motivation

when initially implementing code suggestions i didn't realize that i was printing them after the footers, so we would get this ugly and non-intuitive output:
```
 --> <test.sol>:4:17
  |
4 |         uint256 my_var = 0;
  |                 ^^^^^^
  |
  = help: some footer help msg that should be displayed at the very bottom
help: mutable variables should use mixedCase
  |
4 -         uint256 my_var = 0;
4 +         uint256 myVar = 0;
  |
```

instead, we should print footers at the very bottom
```
 --> <test.sol>:4:17
  |
4 |         uint256 my_var = 0;
  |                 ^^^^^^
  |
help: mutable variables should use mixedCase
  |
4 -         uint256 my_var = 0;
4 +         uint256 myVar = 0;
  |
  = help: some footer help msg that should be displayed at the very bottom
```

### DIFF
--- a/crates/interface/src/diagnostics/emitter/human.rs
+++ b/crates/interface/src/diagnostics/emitter/human.rs
@@ -223,7 +223,6 @@ impl HumanEmitter {
 
         // Dummy subdiagnostics go in the main group's footer, non-dummy ones go as separate groups.
         let subs = |d| diagnostic.children.iter().filter(move |sub| sub.span.is_dummy() == d);
-        let footers = subs(true).map(|sub| message_from_subdiagnostic(sub, self.supports_color()));
         let sub_groups = subs(false).map(|sub| {
             let mut g = Group::with_title(title_from_subdiagnostic(sub, self.supports_color()));
             if let Some(sm) = sm {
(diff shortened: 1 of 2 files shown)


tests/ui/typeck/duplicate_selectors.stderr
```diff
@@ -15,6 +14,7 @@ note: second function
    |
 LL |     function BlazingIt4490597615() public {}
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   = note: the function signatures `mintEfficientN2M_001Z5BWH()` and `BlazingIt4490597615()` produce the same 4-byte selector `0x00000000`
```
DaniPopes: i mean this is not really correct, the note is on the first main diagnostic

0xrusowsky: chagned the render logic to: main > suggestions > footer > sub-group

this way suggestions are consistently printed between main and footer (before this would only happen for inlined suggestions)

PR #543, opened 2025-09-30
feat(ast): spanned optional commasep elements

0xrusowsky:
## Motivation

these 2 foundry issues require knowing the location of the comma separators in optional elements, as we currently don't have any information about `None` elems.

- [ ] https://github.com/foundry-rs/foundry/issues/11836
- [ ] https://github.com/foundry-rs/foundry/issues/8557

## Solution

introduce a new type `SpannedOption<T>`:
```rs
/// An `Option`-like enum that tracks the source location of an absent value.
///
/// This type is used to represent comma-separated items that can be omitted.
/// - The [`Some`](Self::Some) variant holds the parsed item `T`. The item `T` itself is expected to
///   be a spanned type.
/// - The [`None`](Self::None) variant holds the [`Span`] of the empty slot, typically the location
///   of the comma separator.
#[derive(Clone, Copy, Debug)]
pub enum SpannedOption<T> {
    Some(T),
    None(Span),
}
```

and wrap the optional comma-separated elements, so that the wrapper span tracks the separators (opening/closing brace for first/last elements, and commas for others):
```rs
// before
StmtKind::DeclMulti(Box<'ast, [Option<VariableDefinition<'ast>>]>, Box<'ast, Expr<'ast>>)

// after
StmtKind::DeclMulti(Box<'ast, [SpannedOption<VariableDefinition<'ast>>]>, Box<'ast, Expr<'ast>>)
```
```rs
// before
ExprKind::Tuple(Box<'ast, [Option<Box<'ast, Expr<'ast>>>]>)

// after
ExprKind::Tuple(Box<'ast, [SpannedOption<Box<'ast, Expr<'ast>>>]>)
```


crates/parse/src/parser/expr.rs
```diff
@@ -325,10 +327,22 @@ fn token_precedence(t: Token) -> usize {
...
+
+    let vec = list
+        .iter_mut()
+        .map(|spanned_item| {
+            // SAFETY: Caller must ensure that all elements are `Some`.
+            unsafe { spanned_item.data.take().unwrap_unchecked() }
+        })
+        .collect::<Vec<_>>();
+
+    alloc_fn(vec)
```
0xrusowsky: @DaniPopes thoughts on the extra arena allocation?

imo this is an acceptable trade-off, since these lists are typically small and it allows us to have a simple solution.

however, if you'd prefer to avoid the allocation, I can switch to the complex approach and try to manually modify the data in the existing arena region.

crates/ast/src/ast/expr.rs
```diff
@@ -99,7 +99,7 @@ pub enum ExprKind<'ast> {
     Ternary(Box<'ast, Expr<'ast>>, Box<'ast, Expr<'ast>>, Box<'ast, Expr<'ast>>),
 
     /// A tuple expression: `(a,,, b, c, d)`.
-    Tuple(Box<'ast, [Option<Box<'ast, Expr<'ast>>>]>),
+    Tuple(Box<'ast, [Spanned<Option<Box<'ast, Expr<'ast>>>>]>),
```
DaniPopes: hmm actually this would add an extra redundant span in case of Some, we need another type that is Some(expr) None(span)

0xrusowsky: reimplemented using a new `SpannedOption<T>` type as suggested

crates/parse/src/parser/stmt.rs
```diff
@@ -493,30 +544,50 @@ mod tests {
...
                 Ok(())
             })
             .unwrap();
         }
 
+        // Updated test cases with correct expected spans
         check(&[
             ("()", &[]),
-            ("(a)", &[Some("a")]),
-            // ("(,)", &[None, None]),
```
DaniPopes, line 508: these commented ones shouldn't parse because it's invalid syntax

crates/parse/src/parser/stmt.rs
```diff
@@ -254,39 +266,84 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+        self.parse_optional_items_seq_required(delim, &mut out, prev_sep_span, f)
             .map(|()| self.alloc_smallvec(out))
     }
 
     fn parse_optional_items_seq_required<T>(
         &mut self,
         delim: Delimiter,
-        out: &mut SmallVec<[Option<T>; 8]>,
+        out: &mut SmallVec<[SpannedOption<T>; 8]>,
+        mut prev_sep_span: Span,
```
DaniPopes: intuitively i think the span should point to the space in between the commas, not the commas themselves, e.g.

```
(aa ,, cc)
 ^^ 1
     ^ 2 where this has length = 0 after the first comma
        ^^ 3
```

but maybe doesn't matter that much. is there precedent for this? e.g. the solang parser fmt implementation?

crates/parse/src/parser/stmt.rs
```diff
@@ -254,39 +263,74 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
     /// E.g. `(a, b) => [Some, Some]`, `(, a,, b,) => [None, Some, None, Some, None]`.
+    ///
+    /// All elements are wrapped in a `SpannedOption<T>`, so that even for uninformed elements,
+    /// AST consumers can be aware of the location of the separators.
     pub(super) fn parse_optional_items_seq<T>(
         &mut self,
         delim: Delimiter,
         mut f: impl FnMut(&mut Self) -> PResult<'sess, T>,
-    ) -> PResult<'sess, BoxSlice<'ast, Option<T>>> {
+    ) -> PResult<'sess, SmallVec<[SpannedOption<T>; 8]>> {
```
0xrusowsky, line 271: i changed the return type to avoid doing the arena alloc here.

this change allows us to convert `SmallVec<[SpannedOption<T>; 8]` to `SmallVec<[Option<T>; 8]>` before allocating, and safely transmute as the original code did (otherwise i needed to do an extra arena alloc and leak the original allocated arena region)

crates/parse/src/parser/stmt.rs
```diff
@@ -465,25 +503,19 @@ impl<'ast> IndexAccessedPath<'ast> {
...
 #[cfg(test)]
 mod tests {
     use super::*;
     use solar_interface::{Result, Session, source_map::FileName};
 
     #[test]
     fn optional_items_seq() {
-        fn check(tests: &[(&str, &[Option<&str>])]) {
+        #[allow(clippy::type_complexity)]
+        fn check(tests: &[(&str, &[(Option<&str>, &str)])]) {
```
onbjerg: what is the extra `&str` for? afaict its always the same as `T` in `Some(T)` or `""` if it's `None`

0xrusowsky: i used it to verify that the snippets from the spans where accurate.

i guess now that dani asked for `None(Span)` to point to the space between commas it doesn't make sense anymore, and i could directly check vs an empty string

onbjerg: gotcha, yes pls remove

crates/interface/src/span.rs
```diff
@@ -305,3 +305,61 @@ impl<T> Spanned<T> {
...
+impl<T> SpannedOption<T> {
+    pub fn is_none(&self) -> bool {
+        matches!(&self, Self::None(..))
+    }
+
+    pub fn is_some(&self) -> bool {
+        matches!(&self, Self::Some(..))
+    }
+
+    pub fn map<U, F>(self, f: F) -> Option<U>
```
DaniPopes: return spannedoption and have a separate method to drop the span returning Option

crates/parse/src/parser/stmt.rs
```diff
@@ -505,14 +558,15 @@ mod tests {
...
+            // invalid syntax
             // ("(,)", &[None, None]),
             ("(a,)", &[Some("a"), None]),
             ("(,b)", &[None, Some("b")]),
             ("(a,b)", &[Some("a"), Some("b")]),
             ("(a,b,)", &[Some("a"), Some("b"), None]),
+            // invalid syntax
             // ("(,,)", &[None, None, None]),
             ("(a,,)", &[Some("a"), None, None]),
-            ("(a,b,)", &[Some("a"), Some("b"), None]),
```
DaniPopes, line 515: why was this removed?

0xrusowsky: ups, was a mistaske

0xrusowsky: [`fb57d02` (#543)](https://github.com/paradigmxyz/solar/pull/543/commits/fb57d026c0a428c9ec1fa5789bab627b494512e2)

crates/interface/src/span.rs
0xrusowsky: [`fb57d02` (#543)](https://github.com/paradigmxyz/solar/pull/543/commits/fb57d026c0a428c9ec1fa5789bab627b494512e2)

crates/parse/src/parser/stmt.rs
```diff
@@ -197,18 +197,22 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
     fn parse_simple_stmt_kind(&mut self) -> PResult<'sess, StmtKind<'ast>> {
         let lo = self.token.span;
         if self.eat(TokenKind::OpenDelim(Delimiter::Parenthesis)) {
-            let mut empty_components = 0usize;
+            let mut none_elements = SmallVec::<[_; 8]>::new();
             while self.eat(TokenKind::Comma) {
-                empty_components += 1;
+                none_elements.push(self.prev_token.span.shrink_to_hi());
```
DaniPopes, line 202: can we push them as None here directly?

0xrusowsky: not really cause we use this for both `variables` (vardef) and `components` (expr), so i'd have to set the type of the generic in `SpannedOption<T>`

DaniPopes (approved):

PR #555, opened 2025-10-03
feat(ast): add some methods to BinOpKind, DataLocation

DaniPopes:


crates/ast/src/ast/expr.rs
```diff
@@ -294,6 +283,14 @@ impl UnOpKind {
     pub const fn is_postfix(self) -> bool {
         !self.is_prefix()
     }
+
+    /// Returns `true` if the operator is modifying.
```
onbjerg: is mutating a better word here?

onbjerg (approved):

PR #559, opened 2025-10-05
feat: implement HIR builder

### LINKED ISSUE #64: Implement HIR builder
### Describe the feature

Implement a HIR builder that constructs HIR nodes using a bunch of helper methods.

Example:
```rust
pub struct HirBuilder<'hir> { /* ... */ }
impl<'hir> HirBuilder<'hir> {
    pub fn expr(&self, kind: hir::ExprKind<'hir>) -> hir::Expr<'hir> { /* ... */ }
}
```

An initial goal for this would be to replace the manual HIR node construction when desugaring loops and creating variables' getter functions, adding the necessary methods along the way.

### Additional context

_No response_

stevencartavia:
closes #64

### DIFF
--- a/crates/sema/src/ast_lowering/resolve.rs
+++ b/crates/sema/src/ast_lowering/resolve.rs
@@ -607,17 +607,23 @@ impl<'gcx> ResolveContext<'gcx> {
         self.hir.functions[id].parameters = self.arena.alloc_slice_copy(&parameters);
         self.hir.functions[id].returns = self.arena.alloc_slice_copy(&returns);
         self.hir.functions[id].body = {
-            let mk_expr =
-                |kind| &*self.arena.alloc(hir::Expr { id: self.next_id.next(), kind, span });
-            let mk_stmt = |kind| hir::Stmt { span, kind };
+            let builder = self.hir_builder();
 
             // `<gettee><"[" param "]" for param in params>`
             let res = Res::Item(hir::ItemId::Variable(gettee));
-            let mut expr = mk_expr(hir::ExprKind::Ident(self.arena.alloc_as_slice(res)));
+            let mut expr = builder.expr(
+                self.next_id(),
+                hir::ExprKind::Ident(self.arena.alloc_as_slice(res)),
+                span,
+            );
             for &param in &parameters {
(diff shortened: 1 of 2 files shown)


onbjerg (changes requested):
hi! thanks for your interest in this. builder apis in rust usually follow a specific pattern, see e.g. https://rust-unofficial.github.io/patterns/patterns/creational/builder.html

ideally our builders fit the same pattern

crates/sema/src/hir/mod.rs
```diff
@@ -264,6 +264,88 @@ impl<'hir> Hir<'hir> {
...
+
+    /// Creates a builder for constructing HIR nodes.
+    pub fn builder(arena: &'hir bumpalo::Bump) -> HirBuilder<'hir> {
+        HirBuilder::new(arena)
+    }
+}
+
+/// A builder for constructing HIR nodes.
+pub struct HirBuilder<'hir> {
+    arena: &'hir bumpalo::Bump,
```
DaniPopes, line 317: let's also have the next_id here too

crates/sema/src/hir/mod.rs
```diff
@@ -274,12 +274,20 @@ impl<'hir> Hir<'hir> {
 /// A builder for constructing HIR nodes.
 pub struct HirBuilder<'hir> {
     arena: &'hir bumpalo::Bump,
+    next_id: Cell<u32>,
```
DaniPopes: we need to take this by reference so we have unique ids.

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -1133,6 +1150,12 @@ impl<'gcx> ResolveContext<'gcx> {
     fn next_id<I: Idx>(&self) -> I {
         self.next_id.next()
     }
+
+    /// Creates a HIR builder.
+    fn hir_builder(&self) -> hir::HirBuilder<'gcx> {
+        let cell = self.arena.alloc(Cell::new(self.next_id.current()));
```
DaniPopes: this is incorrect, please just pass a reference to next_id

PR #561, opened 2025-10-07
fix(parser): neg. exp. empty check, disallow plus

0xalpharush:
The lexer already raises an error but the parser didn't handle the 1st panic in https://github.com/foundry-rs/foundry/issues/12001

### DIFF
--- a/crates/parse/src/parser/lit.rs
+++ b/crates/parse/src/parser/lit.rs
@@ -317,7 +317,9 @@ fn parse_rational<'ast>(
     if rat.is_some_and(str::is_empty) {
         return Err(LitError::EmptyRational);
     }
-    if exp.is_some_and(str::is_empty) {
+    if let Some(power) = exp
+        && (power.is_empty() || power == "-")
+    {
         return Err(LitError::EmptyExponent);
     }
 
@@ -677,6 +679,8 @@ mod tests {
         check_int("0001", Err(IntegerLeadingZeros));
 
         check_int("0.", Err(EmptyRational));
+        check_int_full("0e-", true, Err(EmptyExponent));
+        check_int_full("0e", true, Err(EmptyExponent));
 
         check_int("0", Ok("0"));
         check_int("0e0", Ok("0"));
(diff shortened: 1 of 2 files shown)


crates/parse/src/parser/lit.rs
```diff
@@ -317,7 +317,9 @@ fn parse_rational<'ast>(
     if rat.is_some_and(str::is_empty) {
         return Err(LitError::EmptyRational);
     }
-    if exp.is_some_and(str::is_empty) {
+    if let Some(power) = exp
+        && (power.is_empty() || power == "-")
```
DaniPopes: what about '+' here too?

0xalpharush: I didn't think it was valid due to [this](https://github.com/paradigmxyz/solar/blob/69cfbed1407fcb7b70a0dcf35f5ea2e348a7026c/crates/parse/src/lexer/cursor/mod.rs#L467-L470) but solc does indeed error with `Invalid exponent` on `uint x = 0.0e+;`

Fixing this would require the lexer to not treat this as a binop
```
thread 'parser::lit::tests::rational' panicked at crates/parse/src/parser/lit.rs:519:13:
assertion `left == right` failed: expected exactly 1 token: [Token { kind: Literal(Rational, "0e"), span: <test>:1:1: 1:3 }, Token { kind: BinOp(Plus), span: <test>:1:3: 1:4 }]
```

0xalpharush: https://github.com/paradigmxyz/solar/pull/561/commits/540a8f5d250469b4ac69f24d1024c373dd8ef438

crates/parse/src/parser/lit.rs
```diff
@@ -677,6 +684,12 @@ mod tests {
         check_int("0001", Err(IntegerLeadingZeros));
 
         check_int("0.", Err(EmptyRational));
+        check_int_full("0e-", true, Err(EmptyExponent));
+        check_int_full("0e", true, Err(EmptyExponent));
+
+        check_int_full("0e+", true, Err(RationalPlusExponent));
+        check_int("0e+0", Err(RationalPlusExponent));
```
0xalpharush, line 691: Solc actually [accepts](https://github.com/argotorg/solidity/blob/77c1dae9a27f879456ec84f6ca70c514309bed23/libsolidity/ast/Types.cpp#L934) this case, `0e+0`, oddly enough. I think it is fine to disallow personally.

Issue #560, opened 2025-10-06
`IntLiteral` always has `negative == false`


onbjerg:
Expressions such as `-1` are parsed as a `UnaryOp` and a positive `IntLiteral`.

Two directions we can go in:

- When parsing `-`, we could peek and see if the next peeled literal is an integer literal, which we then construct as negative (preferred)
- We remove `negative` from `IntegerLiteral` and resolve this later

Test cases:

```solidity
contract C {
    function f() public {
        int8 a = -1;  // negative int literal
        int8 b = -(1); // should be peeled and turned into a negative int literal
        int8 c = -(-1); // unary op negate, negative int literal
        int8 d = -(1 + 2); // unary op negate, binop of two positive int literals
    }
}
```

mablr:
Apparently, it’s not just the case of integer literals, but also rational literals.

```sol
int8 r = -1.2e2;
```

```
[...]
Expr {
    span: src/intliteral.sol:7:18: 7:24,
    kind: Unary(
        UnOp {
            span: src/intliteral.sol:7:18: 7:19,
            kind: Neg,
        },
        Expr {
            span: src/intliteral.sol:7:19: 7:24,
            kind: Lit(
                Lit {
                    span: src/intliteral.sol:7:19: 7:24,
                    symbol: "1.2e2",
                    kind: Number(120),
                },
                None,
            ),
        },
    ),
},
[...]
```

onbjerg:
@mablr Sorry, I was going off of the typeck branch, where `TyKind::IntLiteral` already signals whether it is negative or not, but is always marked as positive: https://github.com/paradigmxyz/solar/blob/731af88c60f93cfe28d7578f610d51c499e0cd2e/crates/sema/src/ty/ty.rs#L518-L519

onbjerg:
Yep, it should not be off main. You can base it off of mine or Dani's, whichever you prefer. It might be easier to base off mine in the sense that once this is fixed, typechecking should fail for `uint256 = -1;` and should pass for `int256 = -1;` (currently both fail). But up to you and no preference.

PR #564, opened 2025-10-07
feat(sema): implicit integer literal conversions

### LINKED ISSUE #627: Integer literal implicit conversion
### Describe the feature

Integer literals are implicitly convertible to any integer type it fits in. Note that an unsigned integer literal can be implicitly converted to a signed integer, but a signed integer literal cannot be converted to an unsigned integer.

`solc` relies on constant folding in the AST, but we will skip that for the time being.

### Additional context

_No response_

onbjerg:
Handles implicit integer resizing (larger → smaller) for `uintN` and `intN`, as well as integer literal conversions. This is plagued by #560 however.

Closes #627

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -352,6 +352,17 @@ impl<'gcx> Ty<'gcx> {
         )
     }
 
+    /// Returns the [`TypeSize`] if the type has a size.
+    pub fn type_size(self) -> Option<TypeSize> {
+        match self.kind {
+            TyKind::Elementary(ElementaryType::Int(size)) => Some(size),
+            TyKind::Elementary(ElementaryType::UInt(size)) => Some(size),
+            TyKind::IntLiteral(_, size) => Some(size),
+            TyKind::Elementary(ElementaryType::FixedBytes(size)) => Some(size),
+            _ => None,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


crates/sema/src/ty/ty.rs
```diff
@@ -424,6 +435,16 @@ impl<'gcx> Ty<'gcx> {
             return Ok(());
         }
 
+        // integers (signed and unsigned) can implicitly be converted to a larger integer. this also
+        // handles integer literals
+        if self.is_integer()
+            && other.is_integer()
+            && (!self.is_signed() || other.is_signed())
```
DaniPopes, line 442: ?
```suggestion
            && self.is_signed() == other.is_signed()
```

onbjerg: no, an unsigned integer can be assigned to a signed integer and thus the implicit conversion works :)

if lhs is signed, you can assign unsigned or signed
if lhs is unsighed, you can only assign unsigned

DaniPopes: there should be a difference between literals which can coerce to either signed or unsigned, and typed integers, which can only upcast

onbjerg: yeah makes sense but unrelated to ts line

PR #566, opened 2025-10-08
feat: handle negative literal integers

### LINKED ISSUE #560: `IntLiteral` always has `negative == false`

mablr:
Closes #560

- Updated `LitKind::Number` variant to handle negative numbers
- Folding of neg unop expr into negative lit - with `LitKind::Number` set to negative - at ast lowering stage

### status

- [x] Handle`-1` as negative integer literal
- [x] Handle by peeling `-(1)` or `-((2))` as negative integer literal as well
- [x] Handle`-(-1)` as unary op negate on negative integer literal (due to first item)
- [x] Handle`-(1 + 2)`  as unary op negate, bin op of two positive int literals

### todo

- [x] *fix typeck* handled by #564
- [ ] add tests (testcases from initial issue) may be introduced instead in the PR mentioned above as the implicit integer conversion would be nice to have a ui test...

### DIFF
--- a/crates/parse/src/parser/expr.rs
+++ b/crates/parse/src/parser/expr.rs
@@ -109,20 +109,6 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
             })
         } else if let Some(unop) = self.token.as_unop(false) {
             self.bump(); // unop
-            // Special handling for negation followed by integer literals
-            if unop.kind == UnOpKind::Neg && self.check_int_lit() {
-                let (lit, subdenomination) = self.parse_lit(true)?;
-                if let LitKind::Number(n, _) = lit.kind {
-                    let span = lo.to(self.prev_token.span);
-                    let negative_lit = self.alloc(Lit {
-                        span,
-                        symbol: lit.symbol,
-                        kind: LitKind::Number(n, true), // Set negative to true
-                    });
-                    return Ok(self
-                        .alloc(Expr { span, kind: ExprKind::Lit(negative_lit, subdenomination) }));
-                }
-            }
             self.parse_unary_expr(None).map(|expr| {
                 let span = lo.to(self.prev_token.span);
                 self.alloc(Expr { span, kind: ExprKind::Unary(unop, expr) })
--- a/crates/sema/src/ast_lowering/resolve.rs
+++ b/crates/sema/src/ast_lowering/resolve.rs
@@ -1084,7 +1084,25 @@ impl<'gcx> ResolveContext<'gcx> {
             )),
             ast::ExprKind::TypeCall(ty) => hir::ExprKind::TypeCall(self.lower_type(ty)),
             ast::ExprKind::Type(ty) => hir::ExprKind::Type(self.lower_type(ty)),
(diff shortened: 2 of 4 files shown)


crates/parse/src/parser/expr.rs
```diff
@@ -109,6 +109,20 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
...
+            // Special handling for negation followed by integer literals
+            if unop.kind == UnOpKind::Neg && self.check_int_lit() {
+                let (lit, subdenomination) = self.parse_lit(true)?;
+                if let LitKind::Number(n, _) = lit.kind {
+                    let span = lo.to(self.prev_token.span);
+                    let negative_lit = self.alloc(Lit {
+                        span,
+                        symbol: lit.symbol,
+                        kind: LitKind::Number(n, true), // Set negative to true
+                    });
```
mablr: Dirty pattern, sorry, until the `Lit` parsing logic will be modified.

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -1084,7 +1084,25 @@ impl<'gcx> ResolveContext<'gcx> {
...
+                {
+                    let neg_int_lit = self.arena.alloc(ast::Lit {
+                        span: expr_span,
+                        symbol: *symbol,
+                        kind: ast::LitKind::Number(*number, !negative),
+                    });
+                    hir::ExprKind::Lit(neg_int_lit)
+                } else {
+                    hir::ExprKind::Unary(*op, self.lower_expr(expr))
+                }
```
onbjerg, line 1104: will this not fail on `-(-1)`? i.e. it would be interpreted as a negative integer literal instead of the negation of a negative literal

mablr: Here is the hir dump for `-(-1)`:
```rust
Expr {
    id: ExprId(2),
    kind: Unary(
        UnOp {
            span: <stdin>:1:45: 1:46,
            kind: Neg,
        },
        Expr {
            id: ExprId(1),
            kind: Tuple(
                [
                    Some(
                        Expr {
                            id: ExprId(0),
                            kind: Lit(
                                Lit {
                                    span: <stdin>:1:47: 1:49,
                                    symbol: "1",
                                    kind: Number(-1),
                                },
                            ),
                            span: <stdin>:1:47: 1:49,
                        },
                    ),
                ],
            ),
            span: <stdin>:1:46: 1:50,
        },
    ),
    span: <stdin>:1:45: 1:50,
},
 ```

mablr: My bad 😅 !!
It actually works well w/ `-(-1)`.

PR #570, opened 2025-10-10
fix(lowering): bring `this` and `super` into scope for constructor inheritance

### LINKED ISSUE #569: bug(lowering): unresolved symbol `this` in contract constructor inheritance
a user raised this issue in foundry https://github.com/foundry-rs/foundry/issues/12047 pointing out that solar fails to process this code:

```solidity
contract Foo {
    constructor(Bar) {}
}

contract Bar is Foo(this) {}
```

the error message seems to indicate that this is a problem in the lowering phase:

```
error: unresolved symbol `this`
 --> src/Aaa.sol:8:29
  |
8 | contract Bar is Foo(address(this)) {}
  |                             ^^^^
```

0xrusowsky:
## Motivation

aims to close #569 by creating a manual scope in "base scope" with only `this`

### DIFF
--- a/crates/sema/src/ast_lowering/resolve.rs
+++ b/crates/sema/src/ast_lowering/resolve.rs
@@ -409,18 +409,25 @@ impl<'gcx> ResolveContext<'gcx> {
         for c_id in self.hir.contract_ids() {
             let contract = self.hir.contract(c_id);
 
-            // Initialize without contract scope, but manually add a local scope with only `this` to
-            // allows builtins to be accessible in base constructor arguments while keeping state
-            // variables inaccessible.
+            // Initialize without contract scope, but manually add a local scope with only `this`
+            // and `super` to allow builtins to be accessible in base constructor
+            // arguments while keeping state variables inaccessible.
             self.init(contract.source, None, None);
             self.scopes.enter();
-            let this_decl = Declaration { res: Res::Builtin(Builtin::This), span: Span::DUMMY };
-            self.scopes.current_scope().declare_unchecked(sym::this, this_decl);
+            let scope = self.scopes.current_scope();
+            scope.declare_unchecked(
+                sym::this,
+                Declaration { res: Res::Builtin(Builtin::This), span: Span::DUMMY },
+            );
+            scope.declare_unchecked(
+                sym::super_,
+                Declaration { res: Res::Builtin(Builtin::Super), span: Span::DUMMY },
+            );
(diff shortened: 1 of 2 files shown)


crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -411,7 +411,16 @@ impl<'gcx> ResolveContext<'gcx> {
...
+            // Initialize without contract scope, but manually add a local scope with only `this` to
+            // allows builtins to be accessible in base constructor arguments while keeping state
+            // variables inaccessible.
             self.init(contract.source, None, None);
+            self.scopes.enter();
+            let span = Span::DUMMY;
+            let this_decl = Declaration { res: Res::Builtin(Builtin::This), span };
+            self.scopes.current_scope().declare_unchecked(sym::this, this_decl);
+
             self.hir.contracts[c_id].bases_args = self.arena.alloc_from_iter(
```
DaniPopes: please move all the code inside of the scope into a new method, this currently doesn't cover the `continue`

crates/sema/src/ast_lowering/resolve.rs
```diff
@@ -407,70 +407,86 @@ impl<'gcx> ResolveContext<'gcx> {
...
-                        span: ast_base.span(),
-                        id: base_id.into(),
-                        args: self.lower_call_args(&ast_base.arguments),
-                    },
-                ),
-            );
-            let contract = self.hir.contract(c_id);
+            self.scopes.enter();
+            let this_decl = Declaration { res: Res::Builtin(Builtin::This), span: Span::DUMMY };
+            self.scopes.current_scope().declare_unchecked(sym::this, this_decl);
```
DaniPopes: what about `super`?

0xrusowsky: you are right, i ran:
```sh
solc - <<EOF
  // SPDX-License-Identifier: MIT
  pragma solidity ^0.8.0;

  abstract contract WithArgs {
      constructor(uint256 b) {}
  }

  abstract contract WithAddrArgs {
      constructor(address) {}
  }

  abstract contract Base {
      function getValue() public virtual returns (uint) { return 42; }
  }

  contract M is Base, WithAddrArgs(address(this)), WithArgs(super.getValue()) {
      constructor() {}
  }
EOF
```

and it looks like solc also supports both `this` and `super`

PR #572, opened 2025-10-13
chore(lowering): allow empty sources

0xrusowsky:
## Motivation

despite solar only supporting solidity v0.8.*, foundry needs to be backwards compatible with older versions. For instance, a user who linted a solar-incompatible project was confused to get a bunch of errors https://github.com/foundry-rs/foundry/issues/11668#issuecomment-3376414455

because of that, solar consumers would benefit from:
- treating empty sources as a valid no-op: so that they can filter incompatible sources and  call the compiler without worrying about error with no compatible sources left. This would save us from conditionally lowering if we don't have any sources, which is kinda annoying:
     ```rs
     if !compiler.sess().source_map().is_empty() {
         let _ = compiler.lower_asts();
     }
     ```

notes:
- as we further integrate solar into more pieces of the stack, this will impact us even more users
- if you think that we should keep the `"no files found"` error was for CLI consumers, i'd propose to at least support a config param to prevent the error

### DIFF
--- a/crates/sema/src/lib.rs
+++ b/crates/sema/src/lib.rs
@@ -50,6 +50,7 @@ pub(crate) fn lower(compiler: &mut CompilerRef<'_>) -> Result<ControlFlow<()>> {
     let sess = gcx.sess;
 
     if gcx.sources.is_empty() {
+        debug!("no files found");
         return Ok(ControlFlow::Break(()));
     }
 


crates/interface/src/lib.rs
```diff
@@ -40,6 +40,11 @@ pub use solar_data_structures as data_structures;
 /// Compiler result type.
 pub type Result<T = (), E = ErrorGuaranteed> = std::result::Result<T, E>;
 
+/// The minimum supported Solidity version.
+///
+/// Solar aims to support Solidity 0.8.* and later versions.
+pub const MIN_SOLIDITY_VERSION: (u8, u8, u8) = (0, 8, 0);
```
DaniPopes: please split this to a new PR

crates/sema/src/lib.rs
```diff
@@ -50,9 +50,7 @@ pub(crate) fn lower(compiler: &mut CompilerRef<'_>) -> Result<ControlFlow<()>> {
     let sess = gcx.sess;
 
     if gcx.sources.is_empty() {
-        let msg = "no files found";
-        let note = "if you wish to use the standard input, please specify `-` explicitly";
-        return Err(sess.dcx.err(msg).note(note).emit());
+        return Ok(ControlFlow::Break(()));
```
DaniPopes, line 54: we still want this check, please keep it in cli

DaniPopes, line 54: maybe leave a debug! log here

PR #574, opened 2025-10-13
refactor(parser): move mapping key type validation to semantic analysis

mablr:
- Removed the mapping key type validation from the parser and implemented it in the semantic analysis phase.
- Added specific mapping-key-type parsing method and updated accordingly mapping-type parsing method
- Added a new method to check if a type is a valid mapping key type.
- Introduced ui tests to ensure that only elementary or user-defined types can be used as key types in mappings.


tests/ui/typeck/mapping_key_type_check.sol
```diff
@@ -0,0 +1,28 @@
...
+}
+
+
+contract C {
+    mapping(uint => uint) m0;
+    mapping(E => uint) m1;
+    mapping(U => uint) m2;
+    mapping(L.E1 => uint) m3;
+
+    // TODO: enable typeck feature to trigger this error
```
onbjerg: you can enable typechecking with `//@compile-flags: -Ztypeck` at the top of the file:)

mablr: Ooh wonderfull !! I was actually searching for that. Thanks.

I'll fix that :+1:

crates/sema/src/typeck/checker.rs
```diff
@@ -625,6 +625,18 @@ impl<'gcx> TypeChecker<'gcx> {
...
     }
 
+    #[must_use]
+    fn check_mapping_key_type(&mut self, key: &'gcx hir::Type<'gcx>) -> Ty<'gcx> {
+        let ty = self.gcx.type_of_hir_ty(key);
+        if !matches!(
+            ty.kind,
+            TyKind::Elementary(_) | TyKind::Udvt(_, _) | TyKind::Contract(_) | TyKind::Enum(_)
+        ) {
+            self.dcx().err("Only elementary types, user defined value types, contract types or enums are allowed as mapping keys.").span(key.span).emit();
```
onbjerg:
```suggestion
            self.dcx().err("only elementary types, user defined value types, contract types or enums are allowed as mapping keys.").span(key.span).emit();
```

stylistic nit

onbjerg (changes requested):

tests/ui/parser/mapping_key_type.sol
```diff
@@ -0,0 +1,20 @@
...
+        B
+    }
+}
+
+contract C {
+    mapping(uint => uint) m0;
+    mapping(E => uint) m1;
+    mapping(U => uint) m2;
+    mapping(L.E1 => uint) m3;
+    mapping(uint[] => uint) m4; //~ ERROR: expected `=>`, found `[`
```
onbjerg: the PR overall looks good to me, but I'm mulling over this, I understand that this is technically incorrect syntax but part of me wishes this was a more helpful message. no action needed right now as I'm still figuring out if we want it like this or not

mablr: I see your point.
This line is actually an edge case because an elementary-type is successfully parsed before falling to parse `=>`:
```solidity
mapping(uint[] => uint) m4;
```
but the following:
```solidity
mapping([] => uint) m4;
```
would return a more insightful error message:
```
error: expected one of elementary type name or path, found `[`
```

PR #580, opened 2025-10-18
feat(typeck,ast_lowering): constructor base arguments validation

mablr:
This is a follow-up for #322

- [x] check if argument number is matching constructor params'
- [x] type check args with associated params ty **(but commented rn as it requires #564)**
- [x] handle dummy ctor call args (`is C` != `is C()`)

Pending #564 ...

### DIFF
--- a/crates/sema/src/typeck/checker.rs
+++ b/crates/sema/src/typeck/checker.rs
@@ -763,13 +763,12 @@ impl<'gcx> hir::Visit<'gcx> for TypeChecker<'gcx> {
                             .span(modifier.span)
                             .emit();
                     } else {
-                        for (_arg_expr, _expected_arg_ty) in
+                        for (arg_expr, expected_arg_ty) in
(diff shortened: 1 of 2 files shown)


crates/sema/src/typeck/checker.rs
```diff
@@ -738,6 +738,46 @@ impl<'gcx> hir::Visit<'gcx> for TypeChecker<'gcx> {
...
+                                ctor_param_types.len(),
+                                arg_count
+                            ))
+                            .span(modifier.span)
+                            .emit();
+                    } else {
+                        for (_arg_expr, _expected_arg_ty) in
+                            modifier.args.exprs().zip(ctor_param_types.iter())
+                        {
+                            // TODO: implicit conversion needed here
```
DaniPopes: can you implement this and leave the TODOs just in the tests?

mablr: all done 🫡

PR #586, opened 2025-10-28
chore(ci): bump artifact actions to v5

MamunC0der:
switch the artifact upload/download steps to the v5 Actions release so we pick up the Node 24 compatibility and latest fixes;


onbjerg:
dupe of #568

PR #602, opened 2025-11-16
refactor(interface): Update annotation_snippets emitter

### LINKED ISSUE #592: Update annotate_snippets emitter
### Describe the feature

Upstream has updated to latest `annotate_snippets` version:

https://github.com/rust-lang/rust/blob/4146079cee94242771864147e32fb5d9adbd34f8/compiler/rustc_errors/src/annotate_snippet_emitter_writer.rs

Before, it was heavily outdated so I wrote the rust-like diagnostic -> annotate_snippets conversion myself, but this has a few subtle bugs and might not have the best output.

Our diagnostics are mostly the same as in Rust, so we should be able to replace this conversion to the one that is now in Rust upstream without too much issue.

### Additional context

_No response_

quangloc99:
Resolve #592

I think the linked issue explains the motivation well enough. I'll list the considerations and changes here for clarity:

- I tried to preserve the ported code as much as possible. So code from Rustc
  that are duplicated or not well structured are kept.
- There are some naming differences.
- We do not need to handle translation like in Rustc.
- In Rustc code, the parameter is **consumed**, while in Solar code, it is passed by reference or Cow.
  There are part that data is cloning to preserve the ported code behavior.
- Some parts like `FileName` and source code handling are simplified as Solar
  does not need to handle as much cases as Rustc.
- There is a minor change in the behavior of the emitter. Currently, Solar emits
  the diagnostic in 4 parts, while Rustc does not separate into parts at all.
  I made the change to keep the footer part, and the rest are combined into one like in Rustc.
- This change make the old code in `diagnostics/emitter/rustc.rs` completely obsolete.
  I still kept it though in case that `FileWithAnnotatedLines::collect_annotations`'s behavior is more desired.

Overall, there is only one test output changed, which is due to the mentioned change in emitter behavior.
I'm surprised that the behavior stayed mostly the same.

Please take a look and let me know if there are any issues. I think the code can be refactored further, either in
this PR or in a separate one.

### DIFF
--- a/crates/data-structures/src/lib.rs
+++ b/crates/data-structures/src/lib.rs
@@ -40,6 +40,28 @@ pub use thin_slice::{RawThinSlice, ThinSlice};
 
 pub use smallvec;
 
+/// Pluralize a word based on a count.
+#[macro_export]
+#[rustfmt::skip]
+macro_rules! pluralize {
+    // Pluralize based on count (e.g., apples)
+    ($x:expr) => {
+        if $x == 1 { "" } else { "s" }
+    };
+    ("has", $x:expr) => {
+        if $x == 1 { "has" } else { "have" }
+    };
+    ("is", $x:expr) => {
+        if $x == 1 { "is" } else { "are" }
+    };
+    ("was", $x:expr) => {
+        if $x == 1 { "was" } else { "were" }
+    };
+    ("this", $x:expr) => {
+        if $x == 1 { "this" } else { "these" }
+    };
+}
+
 /// This calls the passed function while ensuring it won't be inlined into the caller.
 #[inline(never)]
 #[cold]
--- a/crates/interface/src/diagnostics/emitter/human.rs
+++ b/crates/interface/src/diagnostics/emitter/human.rs
@@ -6,7 +6,6 @@ use crate::{
         SuggestionStyle, Suggestions, detect_confusion_type, emitter::normalize_whitespace,
         is_different,
     },
-    pluralize,
     source_map::{FileName, SourceFile},
 };
 use annotate_snippets::{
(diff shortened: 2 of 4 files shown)


khanbilal732 (approved):

crates/interface/src/diagnostics/emitter/human.rs
```diff
@@ -367,156 +594,93 @@ impl HumanBufferEmitter {
...
-        Level::Warning => ASLevel::WARNING,
-        Level::Note | Level::OnceNote => ASLevel::NOTE,
-        Level::Help | Level::OnceHelp => ASLevel::HELP,
-        Level::Allow => ASLevel::INFO,
-    }
-    .with_name(if level == Level::FailureNote { None } else { Some(level.to_str()) })
-}
+fn shrink_file(
+    spans: &[Span],
+    _file_name: &FileName,
```
0xrusowsky, line 670: nit: should we drop this param if unused?

crates/data-structures/src/lib.rs
0xrusowsky, line 1: is there a reason to move this macro from `crates/interface/src/lib.rs`?

0xrusowsky (approved):
i think it makes sense overall, just left a couple of nits

crates/data-structures/src/lib.rs
DaniPopes: cannot be in the same crate due to `macro_expanded_macro_exports_accessed_by_absolute_paths`

crates/interface/src/diagnostics/emitter/human.rs
DaniPopes: to keep as close as possible to upstream

PR #624, opened 2025-12-12
feat(sema): implement explicit conversions for fixed-size byte arrays

### LINKED ISSUE #612: Fixed size byte explicit conversions
### Describe the feature

Fixed size bytes are convertible to and from all sizes. A smaller size truncates the bytes on the right-hand side. A larger size right-pads the value with 0s.

Fixed size bytes can also be explicitly converted to integers of the same size.

### Additional context

_No response_

onbjerg:
Add explicit type conversion support for:
- FixedBytes to FixedBytes (any size)
- FixedBytes to UInt (same size only)
- UInt to FixedBytes (same size only)

Signed integers are not allowed per Solidity spec.

Closes #612

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -467,38 +467,35 @@ impl<'gcx> Ty<'gcx> {
     }
 
     /// Checks if the type is explicitly convertible to the given type.
+    ///
+    /// See: <https://docs.soliditylang.org/en/latest/types.html#explicit-conversions>
     #[allow(clippy::result_unit_err)]
     pub fn try_convert_explicit_to(self, other: Self) -> Result<(), ()> {
         if self.try_convert_implicit_to(other).is_ok() {
             return Ok(());
         }
         match (&self.kind, &other.kind) {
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/sema/src/ty/ty.rs
```diff
@@ -445,10 +445,38 @@ impl<'gcx> Ty<'gcx> {
...
+
+            // FixedBytes to UInt: same size only (signed integers not allowed).
+            // See: <https://docs.soliditylang.org/en/latest/types.html#explicit-conversions>
+            (
+                TyKind::Elementary(ElementaryType::FixedBytes(size_from)),
+                TyKind::Elementary(ElementaryType::UInt(size_to)),
+            ) if size_from == size_to => Ok(()),
+
+            // UInt to FixedBytes: same size only (signed integers not allowed).
+            // See: <https://docs.soliditylang.org/en/latest/types.html#explicit-conversions>
```
DaniPopes: should hoist this repeated link to the function docs

PR #626, opened 2025-12-12
feat(sema): address explicit conversion

### LINKED ISSUE #614: Address explicit conversions
### Describe the feature

Addresses can be explicitly converted to/from `bytes20` and `uint160` only.

An `address` can be converted to an `address payble` with `payable(a)`.

### Additional context

_No response_

dipanshuhappy:
Resolves #614 
Add explicit type conversion support for:
address to payable address
uint160 to address and vice versa
address to bytes20 and vice versa

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -466,66 +466,62 @@ impl<'gcx> Ty<'gcx> {
         self.try_convert_explicit_to(other).is_ok()
     }
 
+    /// Checks if the type is explicitly convertible to the given type.
+    ///
+    /// See: <https://docs.soliditylang.org/en/latest/types.html#explicit-conversions>
     #[allow(clippy::result_unit_err)]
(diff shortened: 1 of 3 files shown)


tests/ui/typeck/address_conversions_explicit.sol
```diff
@@ -0,0 +1,23 @@
...
+        bytes32 b32,
+        uint256 u256
+    ) public pure {
+        // Valid conversions
+        bytes20 b = bytes20(a);
+        address a2 = address(b20);
+        uint160 u = uint160(a);
+        address a3 = address(u160);
+        address payable p = payable(a);
+
```
onbjerg, line 17: can you add a case for `address payable p2 = payable(p)`? we want to sanity check that this also works since it is a no-op (payable to payable).

dipanshuhappy: Sure

PR #632, opened 2025-12-13
Feat(sema): explicit bytes conversion

### LINKED ISSUE #613: Dynamic bytes explicit conversions
### Describe the feature

`bytes` (dynamic size) can be converted explicitly to fixed size bytes. If the source `bytes` is smaller than the target fixed-size bytes, the source bytes are right-padded with zeros. If the source `bytes` is larger, the bytes are discarded on the right-hand side to fit.

### Additional context

_No response_

dipanshuhappy:
Resolves #613 
Add explicit conversion for dynamic bytes to any fixed bytes only one way.


DaniPopes (approved):
Thanks!

PR #633, opened 2025-12-13
feat(sema): integer explicit conversions

### LINKED ISSUE #611: Integer explicit conversions
### Describe the feature

Integers are explicitly convertible to and from all integer types.

If an integer is explicitly converted to a smaller size, higher order bits are discarded.

If an integer is explicitly converted to a larger size, the integer is left padded with 0s.

Integers are also explicitly converted to fixed size bytes of the same size.

### Additional context

_No response_

mablr:
Closes #611 

- Introduced integer explicit conversions
- Added tests

Follows Solidity 0.8.0+ rules regarding type conversions (see https://docs.soliditylang.org/en/latest/080-breaking-changes.html#new-restrictions)

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -623,6 +623,14 @@ impl<'gcx> Ty<'gcx> {
         use ElementaryType::*;
         use TyKind::*;
 
+        macro_rules! unreachable {
+            () => {
+                gcx.dcx()
+                    .bug(format!("unreachable explicit conversion from `{self:?}` to `{other:?}`"))
+                    .emit()
+            };
+        }
+
         if self.try_convert_implicit_to(other, gcx).is_ok() {
             return Ok(());
         }
@@ -660,7 +668,8 @@ impl<'gcx> Ty<'gcx> {
 
             // address -> address payable.
             (Elementary(Address(false)), Elementary(Address(true))) => Ok(()),
-            // IntLiteral -> IntLiteral: explicit conversion to a literal type shouldn't be possible.
+            // IntLiteral -> IntLiteral: explicit conversion to a literal type shouldn't be
+            // possible.
             (IntLiteral(_, _), IntLiteral(_, _)) => unreachable!(),
 
             // Int <-> Int: any size allowed (only width changes, sign stays same).


tests/ui/typeck/integer_explicit_conversions.sol
```diff
@@ -0,0 +1,97 @@
...
+        int8 d = int8(42);
+        int16 e = int16(100);
+        int256 f = int256(12345);
+    }
+
+    // Positive literals with various sizes
+    function validPositiveLiteralVariousSizes() public pure {
+        uint128 a = uint128(0);
+        uint32 b = uint32(0xFFFFFFFF);
+        int64 c = int64(9223372036854775807);  // max int64
```
DaniPopes: can you also check overflow fails? same line but with 9223372036854775808

mablr: That's right, mb. The check on int literal is not sufficient btw. I'll fix it to comply w/:

> 2. Explicit conversions between literals and an integer type T are only allowed if the literal lies between type(T).min and type(T).max. [...]

mablr: I think it's actually blocked for the moment (maybe solved by #566)

In fact we need to check the value and the sign of the literal to make this assertion : type(T).min <=  literal_value < type(T).max. 

So, mapping from/to `Ty` objects is not enough, as we need the value from `LitKind::Number`.

DaniPopes: would it be possible with TypeSize storing bits? that way min..=max for intX is X-1 bits and uintX is X bits

mablr: That's what i tried initially, but it seems to be to restrictive given the granularity of TypeSize (byte).

`IntLiteral(positive)` -> `Uint`: OK
```rust
(IntLiteral(false, size_from), Elementary(UInt(size_to))) => { if size_from.bits() <= size_to.bits() { Ok(()) } else { Result::Err(()) }},
```

`IntLiteral(negative)` -> `Int`: I guess OK (but can't test it without #566)
```rust
(IntLiteral(true, size_from), Elementary(Int(size_to))) => { if size_from.bits() <= size_to.bits() { Ok(()) } else { Result::Err(()) } },
```

`IntLiteral(positive)` -> `Int`: too restrictive 
```rust
(IntLiteral(false, size_from), Elementary(Int(size_to))) => { if size_from.bits() < size_to.bits() { Ok(()) } else { Result::Err(()) } },
```
i.e `int8 d = int8(0xF0);` is rejected because `size_from.bits() == 8` and `size_to.bits() == 8` which doesn't reflect the value bounds (−2^(n−1)  to  2^(n−1) − 1)

DaniPopes: right, what if typesize had bit granularity for literals? so that intX::MIN is typesize(X-1 : u16) instead of typesize(X/8 : u8)

mablr: > what if typesize had bit granularity for literals?

That would be great!


Correct me if I’m wrong, i think to produce `Ty` `IntLiteral` objects, it will be necessary anyway to hold the sign in the ast/hir number (the value as u256 and boolean for the sign).

onbjerg:
I'm going to mark this blocked by https://github.com/paradigmxyz/solar/issues/650 (see above discussion).

tests/ui/typeck/integer_explicit_conversions.sol
dipanshuhappy: > > what if typesize had bit granularity for literals?
> 
> That would be great!
> 
> Correct me if I’m wrong, i think to produce `Ty` `IntLiteral` objects, it will be necessary anyway to hold the sign in the ast/hir number (the value as u256 and boolean for the sign).



Yep it seems so, looking at the expression enum [here](
https://github.com/paradigmxyz/solar/blob/e6880a6bfc1e8bd3801ea3f490a79996a6a1c347/crates/ast/src/ast/expr.rs#L95)
Lit only stores the numerical value, while sign is held by  [Unary](
https://github.com/paradigmxyz/solar/blob/e6880a6bfc1e8bd3801ea3f490a79996a6a1c347/crates/ast/src/ast/expr.rs#L118C4-L119C40)

All these are translated in the typechecker to `Ty::IntLiteral` [here](
https://github.com/paradigmxyz/solar/blob/e6880a6bfc1e8bd3801ea3f490a79996a6a1c347/crates/sema/src/typeck/checker.rs#L982C1-L990C47)

mablr (commented):
> Hi @mablr, this is no longer blocked, see https://github.com/paradigmxyz/solar/pull/671

Thanks @onbjerg, i'll finish it soon.

PR #642, opened 2025-12-15
feat(sema): implicit bytes literal conversion to bytes dynamic/fixed

dipanshuhappy:
Resolves #640 
When I run solc on the test case the error message was 
```
TypeError: Type literal_string "ab" is not implicitly convertible to expected type bytes1. Literal is larger than the type.
```
Solar was giving 
```
mismatched types
```
Happy to work on making the error more descriptive on another issue. If so let me know

PR #645, opened 2025-12-16
feat(sema) Implicit function conversion

### LINKED ISSUE #620: Functions implicit conversions
### Describe the feature

A function type `A` is implicitly convertible to a function type `B` iff their parameter types are identical, their return types are identical, their external/internal property is identical, and the state mutability of `A` is more restrictive than the state mutability of `B`.

- `pure` functions can be converted to `view` and `non-payable` functions
- `view` can be converted to `non-payable`
- `payable` can be converted to `non-payable`

No other function type conversions are possible.

See [Solidity's docs](https://docs.soliditylang.org/en/latest/types.html#conversions).

### Additional context

_No response_

dipanshuhappy:
Resolves #620 
Note: Linking the same issue i noted in PR #642  of more detailed error message. Same issue is here. But I will implement the fix in #642. So you can merge this, and for the error message semantic fix, that would be in #642


crates/sema/src/ty/ty.rs
```diff
@@ -529,6 +529,27 @@ impl<'gcx> Ty<'gcx> {
...
+                        // sanity check
+                        (StateMutability::Payable, StateMutability::Payable) => Ok(()),
+                        (StateMutability::NonPayable, StateMutability::NonPayable) => Ok(()),
+                        (StateMutability::Pure, StateMutability::Pure) => Ok(()),
+                        (StateMutability::View, StateMutability::View) => Ok(()),
+                        _ => Result::Err(TyConvertError::Incompatible),
+                    }
+                } else {
+                    Result::Err(TyConvertError::Incompatible)
+                }
```
onbjerg: let's make this less nested by turning the if statement into a short circuit that returns an error, e.g.


```suggestion
                if from.parameters != to.parameters
                    || from.returns != to.returns
                    || from.visibility != to.visibility
                {
					return Result::Err(TyConvertError::Incompatible);
				}

                match (from.state_mutability, to.state_mutability) {
                    (StateMutability::Pure, StateMutability::View) => Ok(()),
                    (StateMutability::Pure, StateMutability::NonPayable) => Ok(()),
                    (StateMutability::View, StateMutability::NonPayable) => Ok(()),
                    (StateMutability::Payable, StateMutability::NonPayable) => Ok(()),
                    // sanity check
                    (StateMutability::Payable, StateMutability::Payable) => Ok(()),
                    (StateMutability::NonPayable, StateMutability::NonPayable) => Ok(()),
                    (StateMutability::Pure, StateMutability::Pure) => Ok(()),
                    (StateMutability::View, StateMutability::View) => Ok(()),
                    _ => Result::Err(TyConvertError::Incompatible),
                }
```

onbjerg (changes requested):

dipanshuhappy:
I didn't realise this was already pushed some hours ago. @onbjerg  @DaniPopes  closing this PR

PR #649, opened 2025-12-17
feat(typeck): implement constant folding for integer literal type preservation

onbjerg:
When both operands of a binary operation are IntLiteral, evaluate the expression at compile time to preserve literal type semantics. This allows expressions like `-(1 + 2)` to work correctly, since the result remains an IntLiteral that can be negated.

Changes:
- Extended IntScalar with explicit sign tracking and signed arithmetic
- Added try_eval_int_literal_binop to evaluate literal binary expressions
- Updated eval_array_len to reject negative array lengths

On top of #648 

Stack:

- #647
- #648
- #649 (this)


crates/sema/src/eval.rs
```diff
@@ -164,64 +199,109 @@ impl IntScalar {
...
+            Mul => self.checked_mul(r).ok_or(EE::ArithmeticOverflow)?,
+            Div => self.checked_div(r).ok_or(EE::DivisionByZero)?,
+            Rem => self.checked_rem(r).ok_or(EE::DivisionByZero)?,
+            Pow => self.checked_pow(r).ok_or(EE::ArithmeticOverflow)?,
+            BitOr => Self::new(self.data | r.data),
+            BitAnd => Self::new(self.data & r.data),
+            BitXor => Self::new(self.data ^ r.data),
+            Shr => Self::new(self.data.wrapping_shr(r.data.try_into().unwrap_or(usize::MAX))),
+            Shl => Self::new(self.data.wrapping_shl(r.data.try_into().unwrap_or(usize::MAX))),
+            Sar => Self::new(self.data.arithmetic_shr(r.data.try_into().unwrap_or(usize::MAX))),
```
DaniPopes: this is >>> which only works on `fixed` types, we should rename this to sometihng so we dont forget its useless

crates/sema/src/eval.rs
```diff
@@ -164,64 +199,109 @@ impl IntScalar {
...
+            Add => self.checked_add(r).ok_or(EE::ArithmeticOverflow)?,
+            Sub => self.checked_sub(r).ok_or(EE::ArithmeticOverflow)?,
+            Mul => self.checked_mul(r).ok_or(EE::ArithmeticOverflow)?,
+            Div => self.checked_div(r).ok_or(EE::DivisionByZero)?,
+            Rem => self.checked_rem(r).ok_or(EE::DivisionByZero)?,
+            Pow => self.checked_pow(r).ok_or(EE::ArithmeticOverflow)?,
+            BitOr => Self::new(self.data | r.data),
+            BitAnd => Self::new(self.data & r.data),
+            BitXor => Self::new(self.data ^ r.data),
+            Shr => Self::new(self.data.wrapping_shr(r.data.try_into().unwrap_or(usize::MAX))),
```
DaniPopes: this should be arithmetic_shr depending on sign

crates/sema/src/eval.rs
```diff
@@ -135,18 +138,50 @@ impl<'gcx> ConstantEvaluator<'gcx> {
...
+    /// Returns the bit length of the integer value.
+    ///
+    /// This is the number of bits needed to represent the value,
+    /// not including the sign bit.
+    pub fn bit_len(&self) -> u64 {
+        self.data.bit_len() as u64
+    }
+
+    /// Returns whether the value is negative.
+    pub fn is_negative(&self) -> bool {
```
DaniPopes, line 177: this should probably be `is_signed` and be a constant value. then `data` should be I256 and is_negative will be I256's is_negative (last bit set). this way all ops are easier to keep track of, since they become regular two's complement

DaniPopes (changes requested):

DaniPopes (approved):

PR #652, opened 2025-12-17
feat(hir): assembly to hir

### LINKED ISSUE #415: Lower Yul to HIR
### Describe the feature

Currently HIR lowering is incomplete, as inline assembly AST nodes are not lowered.

Lower Yul/inline assembly to HIR.
The main caveat here is that we want to try to merge Yul to standard HIR nodes as much as possible. This means, for example, we translate Yul variable declarations to Solidity ones, function calls to binary operators where possible etc.

For example:
```solidity
function f() {
    assembly {
        function yulfn(a, b) -> c, d {}

        let x := add(42, 69)
        let c, d := yulfn(x, 10)
    }
}

// Becomes
function f() {
    unchecked {
        uint256 x = uint256(42) + uint256(69);
        (uint256 c, uint256 d) = yulfn(x, uint256(10));
    }
}

function yulfn(uint256 a, uint256 b) private returns (uint256 c, uint256 d) {}
```

Not literally in source code or AST, but when lowering to HIR we create normal HIR nodes for Yul nodes that map to the same operations.

This means we need to handle all the Yul builtins and add them to the builtin enum. Since Yul functions are untyped we can have a utility here that creates a function type with N arguments and M returns of type uint256.

### Additional context

_No response_

dipanshuhappy:
Resolves #415 
Will add more description to it 
Currently just initialized yul type for HIR


khanbilal732 (approved):

khanbilal732 (approved):

khanbilal732 (approved):

DaniPopes:
Superseded by https://github.com/paradigmxyz/solar/pull/769.

PR #658, opened 2025-12-18
fix(typeck): mark event/error typechecking todo

onbjerg:
We don't actually typecheck events or errors at the moment, so we should mark this as todo.

### DIFF
--- a/crates/sema/src/typeck/checker.rs
+++ b/crates/sema/src/typeck/checker.rs
@@ -164,7 +164,11 @@ impl<'gcx> TypeChecker<'gcx> {
                         todo!()
                     }
                     TyKind::Type(to) => self.check_explicit_cast(expr.span, to, args),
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/sema/src/typeck/checker.rs
```diff
@@ -164,7 +164,10 @@ impl<'gcx> TypeChecker<'gcx> {
                         todo!()
                     }
                     TyKind::Type(to) => self.check_explicit_cast(expr.span, to, args),
-                    TyKind::Event(..) | TyKind::Error(..) | TyKind::Err(_) => callee_ty,
+                    TyKind::Event(..) | TyKind::Error(..) | TyKind::Err(_) => {
```
DaniPopes: `Err(_)` should still be `return callee_ty` to avoid unnecessary errors (to propagate `Err(_)`)

PR #664, opened 2025-12-18
feat(typeck): extending variable declaration rules

### LINKED ISSUE #657: Variable declaration rules
### Describe the feature

We currently do not have most of the checks for variable declarations in https://github.com/argotorg/solidity/blob/73712a01b2de56d9ad91e3b6936f85c90cb7de36/libsolidity/analysis/TypeChecker.cpp#L475-L596

We need to implement these. The entrypoint for this is in https://github.com/paradigmxyz/solar/blob/38f086aa6d0365c7a38b39b0a364b95d31b63dcb/crates/sema/src/typeck/checker.rs#L582-L596

### Additional context

_No response_

Jds-23:
Closes #657 

- `crates/sema/src/ty/mod.rs`
  - Error message for public interface parameter/return types updated to reject types that "contain non-public function pointers" (adds check for `Ty::has_internal_function`).

- `crates/sema/src/ty/ty.rs`
  - Added `Ty::has_internal_function()` to query a new flag.
  - `can_be_exported()` now disallows types that contain non-public (internal/private) function pointers.
  - Added a new `TyFlags::HAS_INTERNAL_FN` bit.
  - Extended `TyFlags` logic to handle `TyKind::FnPtr`:
    - If the fn pointer's visibility is non-public, set `HAS_INTERNAL_FN`.
    - Propagate flags from the fn pointer's parameter and return types (calls add_tys on parameters and returns).

- `crates/sema/src/typeck/checker.rs`
  - Added helper `is_compile_time_constant()` using ConstantEvaluator.
  - `check_assign()` now enforces rules around assigning types that contain mappings:
    - Disallows assigning storage-containing mapping types unless the lvalue is a local/return variable.
    - Emits an error message for invalid assignments.
  - Added `is_local_or_return_variable()` to detect when an expr refers to a local or return variable.
  - Variable declaration handling (when visiting variables):
    - Constants must be initialized and initializers must be compile-time constant (emit errors otherwise).
    - Immutable variables must be value types (emit error otherwise).
    - State variables containing mappings cannot have initializers (emit error).
    - Non-state variables in calldata/memory that contain mappings are rejected (emit error, instructing the type is only valid in storage).
    - Preserves prior expect_ty check for initializers when present.

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -36,6 +36,9 @@ pub enum TyConvertError {
     /// Invalid conversion between types.
     InvalidConversion,
 
+    /// Literal is larger than the target type.
+    LiteralTooLarge,
+
     /// Contract cannot be converted to address payable because it cannot receive ether.
     ContractNotPayable,
 
@@ -59,6 +62,13 @@ impl TyConvertError {
             }
             Self::Incompatible => {
                 format!("expected `{}`, found `{}`", to.display(gcx), from.display(gcx))
+            }
+            Self::LiteralTooLarge => {
+                format!(
+                    "literal `{}` is larger than the type `{}`",
+                    from.display(gcx),
+                    to.display(gcx)
+                )
             }
             Self::ContractNotPayable => {
                 format!(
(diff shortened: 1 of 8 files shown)


crates/sema/src/ty/ty.rs
```diff
@@ -904,11 +915,20 @@ impl TyFlags {
...
             | TyKind::BuiltinModule(_) => {}
 
+            TyKind::FnPtr(f) => {
+                // Non-public (internal/private) function pointers cannot be exported
+                if f.visibility < Visibility::Public {
+                    self.add(Self::HAS_INTERNAL_FN);
+                }
+                // Also propagate flags from parameter and return types
+                self.add_tys(f.parameters);
+                self.add_tys(f.returns);
```
DaniPopes, line 929: this is not correct - the params and returns are not part of the direct members of the type

PR #666, opened 2025-12-20
feat(sema): helpers to validate parameters and arguments

### LINKED ISSUE #662: Helpers to validate parameters and arguments
### Describe the feature

For emitting events and calling functions, it is possible to use either unnamed or named arguments:

```solidity
f(1, 2, 3)
f({ a: 1, b: 2, c: 3})

emit E(6, 7);
emit E({foo: 6, bar: 7})
```

We want helpers to get named parameters for functions and events on `Ty` (and maybe also in `Hir`?). It is important to note that it is not currently possible to fully implement this for functions, so that should be stubbed out for now.

Additionally, we should have a helper function in the typechecker to match `hir::CallArgs` to function parameters. All the arguments must either be named, or unnamed. No mixing. If all arguments are unnamed, types are checked in sequence. Otherwise, named arguments must be matched to their named parameter counterpart.

### Additional context

An example of how arguments are validated in sequence can be found here: https://github.com/paradigmxyz/solar/blob/f6cf16d0087800bfb62b39adbc3639c241f17127/crates/sema/src/typeck/checker.rs#L787-L808

stevencartavia:
closes #662

### DIFF
--- a/crates/sema/src/ty/ty.rs
+++ b/crates/sema/src/ty/ty.rs
@@ -116,8 +116,8 @@ impl<'gcx> Ty<'gcx> {
 
     pub fn as_externally_callable_function(self, gcx: Gcx<'gcx>) -> Self {
         let is_calldata = |param: &Ty<'_>| param.is_ref_at(DataLocation::Calldata);
-        let parameters = self.parameters().unwrap_or_default();
-        let returns = self.returns().unwrap_or_default();
+        let parameters = self.parameters_with_gcx(gcx).unwrap_or_default();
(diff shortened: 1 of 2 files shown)


crates/sema/src/ty/ty.rs
```diff
@@ -705,6 +771,9 @@ pub enum TyKind<'gcx> {
     /// Function pointer: `function(...) returns (...)`.
     FnPtr(&'gcx TyFnPtr<'gcx>),
 
+    /// Regular function reference.
+    Function(hir::FunctionId),
```
onbjerg, line 837: i don't see you constructing this anywhere?

PR #671, opened 2025-12-22
feat(ast): change TypeSize to store bits instead of bytes

### LINKED ISSUE #650: `TypeSize` is not granular enough
### Reproducible steps

`TypeSize` stores `ceil(bit_len/8)`, so `int_literal[1]` covers the range 0-255.

This means we can't distinguish edge cases like 127 (fits in int8) from 128 (doesn't fit), which is problematic for implicit and explicit integer conversions.

See https://github.com/paradigmxyz/solar/pull/633#discussion_r2616586483 for additional discussion.

### Solar version

N/A

onbjerg:
This enables distinguishing edge cases like 127 (fits in int8) from 128 (doesn't fit), which is needed for correct implicit and explicit integer conversions.

- Changed internal storage from `u8` (bytes 0-32) to `u16` (bits 0-256)
- `TypeSize::new()` now takes bits directly
- Added `new_fb_bytes()` for fixed-bytes types that take byte counts
- `bits()` returns stored value, `bytes()` computes `ceil(bits/8)`
- `mk_ty_int_literal` no longer rounds to multiples of 8

Closes #650

Unblocks #647 #648 #649

### DIFF
--- a/crates/ast/src/ast/ty.rs
+++ b/crates/ast/src/ast/ty.rs
@@ -243,6 +243,23 @@ impl TypeSize {
         if bits.is_multiple_of(8) { Self::new(bits) } else { None }
     }
 
+    /// Creates a new `TypeSize` for an integer literal from **bits**.
+    ///
+    /// Panics if `bits` is zero or greater than 256.
+    #[inline]
+    #[track_caller]
+    pub fn new_literal_bits(bits: u16) -> Self {
+        Self::try_new_literal_bits(bits).unwrap_or_else(|| panic!("invalid literal size: {bits}"))
+    }
+
+    /// Creates a new `TypeSize` for an integer literal from **bits**.
+    ///
+    /// Returns None if `bits` is zero or greater than 256.
+    #[inline]
+    pub fn try_new_literal_bits(bits: u16) -> Option<Self> {
+        if bits == 0 { None } else { Self::new(bits) }
+    }
+
     /// Creates a new `TypeSize` for a fixed-bytes type from **bytes**.
     ///
     /// Panics if `bytes` is not in the range 1..=32.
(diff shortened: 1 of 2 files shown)


crates/ast/src/ast/ty.rs
```diff
@@ -252,51 +252,51 @@ impl TypeSize {
...
-        Self::new(bytes)
+        Self::new(bytes as u16 * 8)
     }
 
-    /// Returns the number of **bytes**, with `0` defaulting to `MAX`.
+    /// Returns the number of **bytes**, with `0` defaulting to 32.
     #[inline]
     pub const fn bytes(self) -> u8 {
-        if self.0 == 0 { Self::MAX } else { self.0 }
+        if self.0 == 0 { 32 } else { self.0.div_ceil(8) as u8 }
```
DaniPopes: max/8

crates/ast/src/ast/ty.rs
```diff
@@ -307,7 +307,7 @@ impl TypeSize {
     #[inline]
     #[track_caller]
     pub const fn bytes_keyword(self) -> Symbol {
-        kw::fixed_bytes(self.0)
+        kw::fixed_bytes(self.bytes())
```
DaniPopes: these should be bytes_raw so that 0 maps to the int uint kw without a bit size

crates/sema/src/ty/mod.rs
```diff
@@ -358,16 +358,16 @@ impl<'gcx> Gcx<'gcx> {
     pub fn mk_ty_string_literal(self, s: &[u8]) -> Ty<'gcx> {
         self.mk_ty(TyKind::StringLiteral(
             std::str::from_utf8(s).is_ok(),
-            TypeSize::new(s.len().min(32) as u8).unwrap(),
+            TypeSize::new(s.len().min(32) as u16 * 8).unwrap(),
```
DaniPopes: we should just rm fn new and use from_bytes from_bits instead

onbjerg: new is used internally in `new_fb_bytes` ("from_bytes") and `new_int_bits` ("from_bits"), i guess we can make new internal?

PR #680, opened 2026-01-14
feat(sema): implement view/pure checker

gakonst:
## Summary

Implements the ViewPureChecker pass that validates state mutability annotations on functions and modifiers. This is modeled after solc's ViewPureChecker.cpp.

### Pure function restrictions:
- Cannot read state variables
- Cannot read `block.*`, `msg.*`, `tx.*` magic variables  
- Cannot call non-pure functions

### View function restrictions:
- Cannot modify state variables
- Cannot emit events
- Cannot create contracts
- Cannot send Ether (via call options)

### Additional:
- `msg.value` requires payable context
- Warns when a function could use a more restrictive mutability annotation

## Test files

Added 7 new test files in `tests/ui/typeck/view_pure/`:
- `pure_reads_state.sol` - Pure function reading state variable
- `pure_reads_block.sol` - Pure function reading block.* 
- `pure_reads_msg.sol` - Pure function reading msg.sender/msg.value
- `view_modifies_state.sol` - View function modifying state
- `view_emits_event.sol` - View function emitting event
- `view_creates_contract.sol` - View function creating contract
- `valid_pure_view.sol` - Valid pure/view functions (no errors)

## Reference

Based on solc's ViewPureChecker.cpp (~471 lines)

### DIFF
--- a/crates/sema/src/typeck/mod.rs
+++ b/crates/sema/src/typeck/mod.rs
@@ -13,6 +13,7 @@ use solar_data_structures::{
 use solar_interface::error_code;
 
 mod checker;
+mod view_pure;
 
 pub(crate) fn check(gcx: Gcx<'_>) {
     parallel!(
@@ -29,6 +30,7 @@ pub(crate) fn check(gcx: Gcx<'_>) {
             if gcx.sess.opts.unstable.typeck {
                 // TODO: Parallelize more.
                 checker::check(gcx, id);
+                view_pure::check(gcx, id);
             }
         }),
     );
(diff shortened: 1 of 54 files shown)


gakonst:
## Updated Review: View/Pure Checker ✅

### Implementation
- New `view_pure.rs` module (486 lines)
- Validates state mutability for view/pure functions
- Reports violations with proper error codes

### Solc Compatibility
✅ References `ViewPureChecker.cpp` in module docs
✅ 17 tests ported from solc syntaxTests/viewPureChecker/
✅ CI passing

### Tests Coverage
- State variable read/write violations
- msg.value/msg.sender in pure
- block.* properties  
- Constants and immutables (valid pure access)
- Mapping access, delete, increment/decrement
- Compound assignment operators

### Error Codes
- 2527: pure function reads state
- 8961: view/pure modifies state  
- 5887: msg.value in non-payable

Ready to merge.

Issue #222, opened 2025-03-02
Incorrect Handling of **


Subway2023:
```solidity
contract LargeExponentTest {
    
    function test() public pure returns (uint256) {
        uint256 result = 10 ** 1e10;
        return result;
    }
}
```
solar --version: 0.1.1
solc reports an error when compiling this program, but solar does not.
```
solc test.sol

Error: Built-in binary operator ** cannot be applied to types int_const 10 and int_const 10000000000.
 --> test.sol:5:26:
  |
5 |         uint256 result = 10 ** 1e10;
  |                          ^^^^^^^^^^
```

DaniPopes:
Thanks for these reports, however solar is incomplete, so it's expected that a lot of errors reported in solc are not reported here, it would be best to report these as bugs once we actually implement these features.

Issue #221, opened 2025-03-02
Incorrect Handling of pure Function


Subway2023:
```solidity
contract TestCompiler {
    uint256 someValue = 10;

    function test() public pure returns (uint256) {
        return someValue;
    }
}
```
`solar --version`: 0.1.1
solc reports an error when compiling this program, but solar does not.

```
solc test.sol

Error: Function declared as pure, but this expression (potentially) reads from the environment or state and thus requires "view".
 --> llmGen/solFromError/0eff.sol:6:16:
  |
6 |         return someValue;
  |                ^^^^^^^^^
```

elijahhampton:
This PR can be closed.

View/Pure checker not implemented: https://github.com/paradigmxyz/solar/issues/1

Issue #219, opened 2025-03-02
[ICE]: Type error in error statement.


Subway2023:
```solidity
contract C {
    enum EnumType {A, B, C}

    struct StructType {
        uint x;
    }
    error E(EnumType StructType, StructType test);
}
```
`solar --version`: 0.1.1
solc can compile successfully, but solar throws an error. If I change EnumType **StructType** to EnumType **otherName**, the error does not occur.
```
error: name has to refer to a valid user-defined type
 --> getSeed/seed/error_param_type_shadowed_by_param_name.sol:7:34
  |
7 |     error E(EnumType StructType, StructType test);
  |                                  ^^^^^^^^^^
  |
```

PR #682, opened 2026-01-14
fix(sema): fix typechecker ICE bugs #216 #219 #221 #222

### LINKED ISSUE #216: [ICE]: identifier `this` already declared
```solidity
contract C {
	event this();
}
```
`solar --version`: 0.1.1
solc can compile successfully, but solar throws an error. If I change this to another name, the error does not occur.
```
error: identifier `this` already declared
 --> getSeed/seed/illegal_names_exception.sol:2:8
  |
2 |     event this();
  |       ^^^^
  |

error: aborting due to 1 previous error
```

gakonst:
## Summary

This PR fixes 4 typechecker bugs reported in GitHub issues:

### #216 - Events named `this` and `super` now allowed

Events can coexist with builtins (`this`, `super`) in the same scope without conflicts. The fix adds handling in `conflicting_declaration` to treat builtins as compatible with events.

```solidity
contract C {
    event this();  // Now compiles successfully
    event super(); // Now compiles successfully
}
```

### #219 - Error/event/struct parameter names no longer shadow type names

Parameters for error, event, and struct declarations don't declare names in scope since they don't create a lexical scope where names can be used. This prevents the first parameter's name from shadowing a type name for subsequent parameters.

```solidity
contract C {
    struct StructType { uint x; }
    // Previously failed - StructType was shadowed by the first parameter's name
    error E(uint256 StructType, StructType test);  // Now compiles successfully
}
```

### #221 - Pure functions now validated for state access

Added state mutability validation for pure functions. Pure functions now error when accessing state variables that are not constant or immutable.

```solidity
contract C {
    uint256 stateVar = 10;
    function test() public pure returns (uint256) {
        return stateVar;  // Error: requires "view"
    }
}
```

### #222 - Large exponent literals validated

Exponentiation operations with integer literals that would overflow are now rejected at compile time.

```solidity
contract C {
    function test() public pure returns (uint256) {
        return 10 ** 1e10;  // Error: arithmetic overflow
    }
}
```

## Changes

- `crates/sema/src/ast_lowering/resolve.rs`: 
  - Fixed event/builtin conflict detection
  - Parameters for events, errors, structs, and function type signatures no longer declare in scope

- `crates/sema/src/typeck/checker.rs`:
  - Added `function_state_mutability` tracking
  - Added `check_state_access` validation for pure functions
  - Added power operation validation for integer literals
  - Added `visit_function` to track function state mutability

- `crates/sema/src/eval.rs`: Made `EvalErrorKind::msg` public

## Testing

Added 4 regression tests:
- `tests/ui/resolve/event_this_super.sol`
- `tests/ui/resolve/error_param_type_shadowing.sol`  
- `tests/ui/typeck/literal_pow_overflow.sol`
- `tests/ui/typeck/pure_state_access.sol`

All existing tests continue to pass.

Fixes #216, #219, #221, #222

### DIFF
--- a/crates/sema/src/ast_lowering/resolve.rs
+++ b/crates/sema/src/ast_lowering/resolve.rs
@@ -903,7 +903,7 @@ impl<'gcx> ResolveContext<'gcx> {
         self.arena.alloc_slice_fill_iter(vars.iter().map(|var| self.lower_variable(var, kind).0))
     }
 
-    /// Lowers `var` to HIR and declares it in the current scope.
+    /// Lowers `var` to HIR and declares it in the current scope (if applicable).
     fn lower_variable(
         &mut self,
         var: &ast::VariableDefinition<'_>,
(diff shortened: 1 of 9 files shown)


DaniPopes:
incorrect approach, will revisit later once more of the pipeline is implemented.

PR #726, opened 2026-01-26
feat: PGO

DaniPopes:
needs upstream support in cargo-dist


gakonst (commented):
Nice - we should also give this another shot on Reth / tempo

PR #753, opened 2026-04-26
fix(sema): suppress parse drop warning after errors

josusanmartin:
## Summary

Suppress the `ParsingContext::parse` drop-time warning when diagnostics have already been emitted. This path is legitimate when parsing exits early after an error, for example the missing-file UI test.

## Verification

- `cargo check --workspace`
- `cargo uitest`

This is intended as the small base fix for the frontend performance stack.


josusanmartin:
Superseded by the combined frontend performance PR #754, which includes this one-line fix as its first commit so the stack can be reviewed and merged in one PR.

PR #757, opened 2026-04-30
fix(parse): replace match with ? operator for clippy::question_mark

JFKongphop:
## Summary
Fixes `clippy::question_mark` lint error that was failing in CI.

## Problem
CI was failing because a `match` expression in `parse_seq_inner()` was manually handling `Ok`/`Err` cases where the `?` operator provides cleaner, idiomatic error propagation.

## Solution
Replaced the explicit `match` on `Result` with the `?` operator. The logic and behavior remain identical — the `?` operator automatically returns early on `Err`, while the success case proceeds with the same recovery logic.

### Before
```rust
match self.expect(sep_kind) {
    Ok(recovered_) => {
        if recovered_ == Recovered::Yes {
            recovered = Recovered::Yes;
            break;
        }
    }
    Err(e) => return Err(e),
}
```

### After
```rust
{
    let recovered_ = self.expect(sep_kind)?;
    if recovered_ == Recovered::Yes {
        recovered = Recovered::Yes;
        break;
    }
}
```

### DIFF
--- a/crates/parse/src/parser/mod.rs
+++ b/crates/parse/src/parser/mod.rs
@@ -669,14 +669,12 @@ impl<'sess, 'ast> Parser<'sess, 'ast> {
                     first = false;
                 } else {
                     // check for separator
-                    match self.expect(sep_kind) {
-                        Ok(recovered_) => {
-                            if recovered_ == Recovered::Yes {
-                                recovered = Recovered::Yes;
-                                break;
-                            }
+                    {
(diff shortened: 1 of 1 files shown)


JFKongphop:
why you close it when i build in ci that cannot build because it crash in this case

DaniPopes:
It was already fixed on main by another commit

JFKongphop:
Thank you sir🫡

PR #760, opened 2026-05-08
ci(codegen): run solar benchmark corpus

djolertrk:
Adds `codegen` CI using [solidity-compiler-benchmarks](https://github.com/walnuthq/solidity-compiler-benchmarks)

The workflow builds `solar`, compiles the benchmark corpus with both `solc` and `solar`, then checks:
  - bytecode size
  - gas usage on Anvil
  - runtime output equality for the same calls/inputs

While trying to build the contracts from the set of contracts I shared, some `codegen` issues occurred. In this PR, I fixed a `codegen` bug: cross-block MIR values were not always spilled before leaving a block, so successor blocks could lose values after the stack model was cleared. The fix improves liveness propagation and preallocates/spills live cross-block values in EVM `codegen`.
Besides that, there is `fix(sema): canonicalize bare integer types`, and it should be a separate PR; it fixes `uint/uint256` alias handling, detected in `maple-erc20`.

### DIFF
--- a/.github/scripts/check_codegen_benchmark.py
+++ b/.github/scripts/check_codegen_benchmark.py
@@ -148,6 +148,7 @@ def main() -> int:
     repo_runtime = runtime_issue_details(repo_results)
     for detail in repo_runtime:
         warning(f"repository runtime mismatch recorded: {detail}")
+    failures.extend(f"repo {failure}" for failure in repo_runtime)
 
     if failures:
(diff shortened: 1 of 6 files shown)


djolertrk:
`codegen` CI Action fails, and it is expected, since it detects the mismatch in the runtime, for `solar` vs `solc` compiled contracts.

I will keep fixing the `codegen` so we can have it green.

DaniPopes (approved):

PR #761, opened 2026-05-08
fix(sema): canonicalize bare integer aliases

djolertrk:
This fixes uint/uint256 alias handling.
It was detected during `codegen` testing (https://github.com/paradigmxyz/solar/pull/760), in particular while building `maple-erc20`.


djolertrk:
Ping :) cc @DaniPopes

DaniPopes (approved):

PR #787, opened 2026-05-20
fixup: fix codegen for evm

djolertrk:
Continue with fixing `codegen` to avoid mismatches in runtime comparing to `solc`.

### DIFF
--- a/crates/sema/src/ast_lowering/yul.rs
+++ b/crates/sema/src/ast_lowering/yul.rs
@@ -174,13 +174,12 @@ impl<'a, 'gcx> YulLoweringContext<'a, 'gcx> {
             ast::yul::ExprKind::Path(path) => hir_yul::ExprKind::Path(self.lower_path(path)),
 
             ast::yul::ExprKind::Call(call) => {
-                // Validate the call target exists (builtin or user-defined function).
-                self.validate_call_target(call.name);
+                let res = self.resolve_call_target(call.name);
                 let arguments = self
                     .rcx
                     .arena
                     .alloc_slice_fill_iter(call.arguments.iter().map(|e| self.lower_expr_full(e)));
-                hir_yul::ExprKind::Call(hir_yul::ExprCall { name: call.name, arguments })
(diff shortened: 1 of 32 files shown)


crates/sema/src/ast_lowering/yul.rs
```diff
@@ -216,129 +207,45 @@ impl<'a, 'gcx> YulLoweringContext<'a, 'gcx> {
...
-            return;
-        }
-
-        // Not found.
-        self.rcx
-            .dcx()
-            .err(format!("undefined function `{}`", name.name))
-            .span(name.span)
-            .emit();
+        let _ = name.is_yul_evm_builtin() || self.lookup_yul_function(name.name).is_some();
```
DaniPopes: are you merging main here?

djolertrk: That was a temporary workaround after removing the old parser-side `Yul` call validation. I cleaned it up now (3 new commits in this PR): `sema` resolves `Yul` call targets explicitly, and `codegen` emits diagnostics for targets it cannot lower.

We needed this split because a Solidity file with inline assembly can fail for two different reasons: invalid `Yul`, or valid `Yul` that our `MIR`/`codegen` path does not support yet. The added tests cover that distinction.

DaniPopes: sure, in codegen lowering makes sense; I'm noting that on main i've implemented most of yul type checking so you can probably reuse that if you merge it?

djolertrk: I see, I see. I will definitely check it out in detail. I just briefly looked into it.

djolertrk: Would be better to rebase https://github.com/paradigmxyz/solar/tree/feat/codegen-mir on top of `main` first.

DaniPopes: Yeah, let's do that in another pr

DaniPopes (approved):

PR #791, opened 2026-05-21
refactor(diagnostics): outline builder mutation

DaniPopes:
Forwarded diagnostic builder methods keep their chaining API, but their generated wrappers now inline only the ownership handoff and call a no-inline helper that mutates through `&mut self`. This keeps the cold diagnostic mutation outlined without forcing every forwarded method body to be shaped as `mut self -> Self`.


DaniPopes:
actually `self -> Self` is better because it's passed by value (register).

PR #792, opened 2026-05-21
refactor(sema): skip arena ty list interning

DaniPopes:
This skips the type-list interner lookup when the input slice already lives in the HIR arena. In that case the list can be returned directly with the global context lifetime instead of hashing and probing the intern map again.


DaniPopes:
no hits

PR #795, opened 2026-05-22
fix(sema): check variadic builtins

DaniPopes:
Preserve builtin identity for member lookups and type-check variadic builtins such as `string.concat`, `bytes.concat`, and ABI encode helpers without treating their cached zero-parameter signatures as fixed arity.


DaniPopes:
Superseded by #815.

PR #797, opened 2026-05-22
fix(sema): expose base internal members

DaniPopes:
Expose internal and public base-contract functions through qualified base contract member lookup when checking a contract in that inheritance chain, while keeping external base functions as non-callable contract-type declarations.


DaniPopes:
Superseded by #814.

PR #805, opened 2026-05-22
fix(sema): type abi.decode returns

DaniPopes:
Teach type checking to treat abi.decode as a special ABI builtin whose second argument is a tuple of type expressions and whose return type is the decoded type list.

This covers both single-value and multi-value decode results, including memory locations for decoded reference types.


DaniPopes:
Superseded by #815.

PR #806, opened 2026-05-22
fix(sema): expose scoped state vars

DaniPopes:
Expose visible state variables on a contract type when the referenced contract is the current contract or one of its bases.

This matches solc for reads such as Scope.stateVar inside Scope while keeping internal state variables hidden from unrelated contract scopes.


DaniPopes:
Superseded by #814.

PR #809, opened 2026-05-22
fix(sema): allow indexing storage return values

DaniPopes:
Allows function calls that return storage references to stay usable as storage bases for subsequent indexing, so expressions like `f()[k] = v` type check when `f()` returns a storage mapping.

This mirrors solc's behavior for `test/libsolidity/semanticTests/functionCall/mapping_internal_return.sol`: the function call itself is not a general assignable lvalue, but the returned storage reference must not poison the lvalue state before the mapping index access is checked.


DaniPopes:
Duplicate of #808.

PR #693, opened 2026-01-14
feat(codegen): add solar-codegen crate with MIR and EVM codegen

gakonst:
## Summary

This adds the `solar-codegen` crate which provides MIR (Mid-level Intermediate Representation) and EVM bytecode generation for Solar.

## Test Results

**95 tests pass** across 6 parallel test suites comparing Solar vs solc:

| Metric | Solar | solc | Difference |
|--------|-------|------|------------|
| Bytecode size | 105-1856B | 370-4340B | **61-72% smaller** |
| Gas usage | varies | varies | **2-48% cheaper** |
| Stack depth | ✓ compiles | ✗ "stack too deep" | **Solar wins** |

### Per-Project Results

| Project | Tests | Contracts |
|---------|-------|-----------|
| arithmetic | 28 | Counter, Arithmetic |
| control-flow | 16 | ControlFlow (loops, conditionals, break) |
| storage | 37 | StorageInit, DynamicArray, Mappings (3 levels), Payable |
| events | 2 | Events (LOG0-LOG4) |
| calls | 5 | ExternalCall (direct, chained, with value) |
| stack-deep | 7 | Showcase, StackDeep (16+ params, 20+ locals) |

### Sample Output
```
📦 Compilation + Test Time:
   Solar:   1.02s | solc:   1.67s | -39%

📏 Bytecode Sizes (deployed):
   Arithmetic           Solar:  1413B | solc:  4122B | +65% smaller
   Counter              Solar:   105B | solc:   370B | +71% smaller

⛽ Gas Usage (per test):
   ✓ test_AddBasic()    Solar: 5735 | solc: 11182 | -48.7%
   ✓ test_MulBasic()    Solar: 5941 | solc: 11369 | -47.7%
```

## Architecture

```
HIR → Lowering → MIR → Stack Scheduling → EVM Bytecode
```

### MIR Structure

- **Core Types**: `ValueId`, `InstId`, `BlockId`, `FunctionId` index types
- **MIR Types**: `UInt(u16)`, `Int(u16)`, `Address`, `MemPtr`, `StoragePtr`, `CalldataPtr`, `Function`, `Bool`, `FixedBytes`
- **Value**: SSA values (`Inst`, `Arg`, `Immediate`, `Phi`, `Undef`) with `Immediate` constants
- **InstKind**: Comprehensive instruction kinds (Arithmetic, Bitwise, Comparison, Memory, Storage, Environment, Calls, Control Flow, SSA Phi/Select)
[...]

PR #822, opened 2026-05-25
feat: codegen

### LINKED ISSUE #508: Add project benches
### Describe the feature

Currently all benchmarks are single file lexing + parsing benches. We should include multifile benchmarks as well.

### Additional context

_No response_

DaniPopes:
Implement lowering to MIR (middle intermediate representation), and an EVM code generation backend. Enabled under -Zcodegen, --emit mir, and standard json support.
Tested with Foundry, benchmarked against solc.

Continuing #693. See that PR for some more details.

Closes #508.
Closes #704.


djolertrk:
This PR is merge of main to the codegen branch: https://github.com/paradigmxyz/solar/pull/836

crates/cli/src/lib.rs
```diff
@@ -120,9 +122,195 @@ pub(crate) fn run_pipeline(
...
+
+    if needs_bytecode {
+        emit_bytecode(compiler)?;
+    }
+
     Ok(ControlFlow::Continue(()))
 }
 
+/// Emit textual MIR for all contracts using solar-codegen.
+fn emit_mir(compiler: &mut CompilerRef<'_>) -> Result {
```
DaniPopes: can we unify this with the sema emit? maybe just remove the emit module from sema and handle it in cli, since it's mainly CLI facing thing, printing json
also move this to a separate module

crates/cli/src/standard_json.rs
```diff
@@ -641,14 +664,24 @@ fn make_contract_output(
         contract_name,
         &["evm.bytecode", "evm.bytecode.object"],
```
DaniPopes, line 826: are we handling this correctly? evm.bytecode is a superset of .object, i'm not sure what the exact fields here we're supposed to emit. maybe this ties in to a larger effort of making standard json output match solc as much as possible wrt pure input->output keys

crates/cli/src/standard_json.rs
```diff
@@ -203,6 +205,16 @@ fn compile(
 
             if compile_result.is_ok() && compiler.dcx().has_errors().is_ok() {
```
DaniPopes, line 310: i think we're missing some fields to parse from input, for example optimize bool and optimizer runs

we should define all the possible fields in input struct and explicitly handle each one, even if `let _` with a comment saying why we don't use it

crates/codegen/tests/round_trip.rs
DaniPopes, line 1: we can move this into mir module as an inline `#[cfg(test)] mod tests {...}` since its specifically parsing/printing/reparsing  mir text

crates/config/src/opts.rs
```diff
@@ -92,6 +92,12 @@ pub struct Opts {
     /// Stop execution after the given compiler stage.
     #[cfg_attr(feature = "clap", arg(long, value_enum))]
     pub stop_after: Option<CompilerStage>,
+    /// Disable MIR optimization passes during codegen.
+    #[cfg_attr(feature = "clap", arg(long, hide = true))]
+    pub no_opt: bool,
```
DaniPopes: do we still need this? when we have -Onone

crates/sema/src/typeck/checker.rs
```diff
@@ -104,18 +102,22 @@ impl<'gcx> TypeChecker<'gcx> {
...
-            Ok(value) => {
+        let mut evaluator = ConstantEvaluator::new(self.gcx);
+        match evaluator.try_eval_value(slot) {
+            Ok(ConstValue::Integer(value)) => {
                 if value.as_u256().is_none() {
                     self.dcx().emit_err(slot.span, "base slot of storage layout evaluates to a value outside the range of type `uint256`");
                 }
             }
-            Err(_err) => {}
+            Ok(ConstValue::Bool(_)) => {}
```
DaniPopes: shouldn't this error? what is expected for `contract C at true`

crates/sema/src/typeck/checker.rs
```diff
@@ -2416,6 +2480,31 @@ impl<'gcx> TypeChecker<'gcx> {
                 .emit();
         }
     }
+
+    fn unsupported_codegen_udvt_operator(&self, expr: &'gcx hir::Expr<'gcx>, ty: Ty<'gcx>) -> bool {
```
DaniPopes, line 2495: is this temporary? if so, can you note that?

benches/benches/compile_time.rs
DaniPopes, line 1: let's merge these into benches/src/lib.rs, we already do 90% of this file for other benchmarks

djolertrk:
lgtm

PR #834, opened 2026-06-02
feat: test cleanup and type checking reuse

djolertrk:


DaniPopes:
Can you move the standalone parser/lowering fixes to a pr to main?

djolertrk:
> Can you move the standalone parser/lowering fixes to a pr to main?

Sure! Thanks for the review.

crates/cli/src/mir_opt.rs
```diff
@@ -93,13 +70,13 @@ struct Args {
     input: String,
 }
 
-fn parse_args() -> Result<Args, String> {
-    let argv: Vec<String> = std::env::args().collect();
+/// Parses the arguments following `mir-opt`.
+fn parse_args(argv: &[String]) -> Result<Args, String> {
```
DaniPopes: please use clap, you should be able to have an enum for the commands, see https://github.com/paradigmxyz/solar/pull/401

crates/sema/src/typeck/checker.rs
```diff
@@ -45,6 +45,12 @@ enum AbiDecodeArg {
 pub(super) fn check(gcx: Gcx<'_>, source: hir::SourceId) {
     let mut checker = TypeChecker::new(gcx, source);
     let _ = checker.visit_nested_source(source);
+    // Publish the per-expression types so later phases (codegen) can reuse them
+    // instead of re-deriving them. Keys are globally-unique `ExprId`s, so this
+    // never collides with other sources checked in parallel.
+    for (id, ty) in checker.types.drain() {
+        gcx.insert_expr_type(id, ty);
+    }
```
DaniPopes: done here on main https://github.com/paradigmxyz/solar/pull/833

djolertrk:
> > Can you move the standalone parser/lowering fixes to a pr to main?
> 
> Sure! Thanks for the review.

@DaniPopes here is the frontend part: https://github.com/paradigmxyz/solar/pull/835

crates/cli/src/mir_opt.rs
djolertrk: done.

crates/cli/src/mir_opt.rs
```diff
@@ -0,0 +1,260 @@
...
+  none             No transform; just lower/parse and print
+
+Input formats:
+  *.sol  Solidity contract — lowered through the normal compiler pipeline
+  *.mir  Textual MIR — parsed directly via solar_codegen::mir::parse_module
+";
+
+/// The canonical pass list run by `EvmCodegen::run_optimization_passes`.
+/// Keep in sync with `crates/codegen/src/backend/evm/evm.rs`.
+const DEFAULT_PIPELINE: &[PassName] =
```
DaniPopes, line 41: can we unify these so they're only defined in the codegen crate? like a pass manager that can be queried for known passes or smth like that?

DaniPopes (approved):

crates/cli/src/mir_opt.rs
djolertrk: Hmm, yes, good idea. I will incorporate that in the next PR. Thanks.

PR #841, opened 2026-06-05
Replace bumpalo with stumpalo

decofe:
## Summary
- replace Solar workspace uses of bumpalo with stumpalo
- adapt arena extension helpers to stumpalo public allocation APIs
- keep Solar arena wrappers Send for existing thread-local parallel parsing/typechecking flows

## Tests
- cargo test -p solar-data-structures bump_ext --lib
- cargo check --workspace

Prompted by: @DaniPopes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2529,6 +2529,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "readme-rustdocifier"
+version = "0.1.1"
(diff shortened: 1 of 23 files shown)


DaniPopes:
Closing in favor of #844, which redoes this on current main with stumpalo 0.2.0 and without local unsafe Send impls.

PR #847, opened 2026-06-09
feat: improve transformations furthermore

djolertrk:
@DaniPopes, this removes the `PassName` completely.


crates/cli/src/standard_json.rs
```diff
@@ -723,6 +723,8 @@ fn generate_contract_bytecodes(
             let mut module = lower::lower_contract_with_bytecodes(gcx, contract_id, &all_bytecodes);
             gcx.dcx().has_errors()?;
             let mut codegen = EvmCodegen::new();
+            codegen.set_optimize(!gcx.sess.opts.no_opt);
+            codegen.set_mir_print_after_each(gcx.sess.opts.unstable.mir_print_after_each);
```
DaniPopes: `EvmCodegen::new(gcx)`? we'll have a lot more configs

tests/ui/cli/help.long.stdout
```diff
@@ -28,6 +28,9 @@ Options:
           
           [possible values: parsing, lowering, analysis]
 
+      --no-opt
```
DaniPopes: -O none,gas,size ?

tests/ui/codegen/mapping.stdout
```diff
@@ -3,73 +3,73 @@
...
 fn @approve(arg0: address, arg1: address, arg2: u256) {
   bb0 (entry):
-    mstore 0, arg0
-    mstore 32, 1
-    v9 = keccak256 0, 64
-    mstore 0, arg1
-    mstore 32, v9
-    v15 = keccak256 0, 64
-    sstore v15, arg2
+    mstore 0, arg0 !metadata(memory=scratch)
```
DaniPopes, line 53: what do you think of a malloc instruction? would that help or would it just complicate analyses?

crates/codegen/src/mir/inst.rs
```diff
@@ -730,6 +888,60 @@ impl InstKind {
...
+            | Self::GasLimit
+            | Self::ChainId
+            | Self::Address
+            | Self::Balance(_)
+            | Self::SelfBalance
+            | Self::Gas
+            | Self::BaseFee
+            | Self::BlobBaseFee
+            | Self::BlobHash(_) => EffectKind::EnvironmentRead,
+            _ => EffectKind::Pure,
```
DaniPopes: make this exhaustive

crates/codegen/src/mir/inst.rs
```diff
@@ -730,6 +888,60 @@ impl InstKind {
...
+            Self::Call { .. } | Self::StaticCall { .. } | Self::DelegateCall { .. } => {
+                EffectKind::ExternalCall
+            }
+            Self::InternalCall { .. } => EffectKind::InternalCall,
+            Self::Create(_, _, _) | Self::Create2(_, _, _, _) => EffectKind::Create,
+            Self::Log0(_, _)
+            | Self::Log1(_, _, _)
+            | Self::Log2(_, _, _, _)
+            | Self::Log3(_, _, _, _, _)
+            | Self::Log4(_, _, _, _, _, _) => EffectKind::Log,
```
DaniPopes, line 920: we can probably unify some of these instructions, like logN -> log with a dynamic list of params, *call* -> extcall with a kind const argument etc

crates/codegen/src/pass.rs
```diff
@@ -16,180 +16,300 @@
...
-            "memory-dse" => Self::MemoryDse,
-            "dce" => Self::Dce,
-            _ => return None,
-        })
+    fn make_pass(&self) -> Box<dyn ModulePass> {
+        (self.make_pass)()
     }
 }
 
+fn make_inline_pass() -> Box<dyn ModulePass> {
```
DaniPopes: these can all be inlined as closures inside of the consts, like || Box::new(..)

crates/codegen/src/pass.rs
```diff
@@ -322,17 +446,40 @@ impl PassManager {
...
+                am.invalidate_all();
+            }
+            validate_module_after_pass(module, &pass_name);
         }
-        am
+        (am, changed)
+    }
+}
+
+#[cfg(debug_assertions)]
```
DaniPopes: should be a dynamic config enabled by default by cfg, not just a cfg

crates/codegen/src/analysis/scalar_evolution.rs
```diff
@@ -0,0 +1,358 @@
...
+mod tests {
+    use super::*;
+    use crate::{
+        analysis::LoopAnalyzer,
+        mir::{Function, Immediate, InstKind, Instruction, MirType, Terminator, Value},
+    };
+    use solar_interface::Ident;
+
+    #[test]
+    fn recognizes_base_plus_scaled_induction_variable() {
```
DaniPopes: still not a fan of these; if you need to assert a specific internal state rather than an e2e transformation you can add a -Zprint-something=else flag and assert the stdout

crates/codegen/src/transform/cfg_simplify.rs
```diff
@@ -537,6 +437,14 @@ impl CfgSimplifier {
             _ => {}
         }
     }
+
+    fn refresh_successors(func: &mut Function, block_id: BlockId) {
+        let Some(term) = func.blocks[block_id].terminator.as_ref() else {
+            func.blocks[block_id].successors.clear();
+            return;
+        };
+        func.blocks[block_id].successors = term.successors();
```
DaniPopes: why do we store successors in the first place if we can just do .successors() on the terminator?

tests/ui/cli/help.long.stdout
djolertrk: Yes! It works!
 `-O gas` will be default

tests/ui/codegen/mapping.stdout
djolertrk: Hmmm. Good question. I think it could help, but only if scoped to _**compiler-owned heap allocations**_. For `scratch`/`ABI`/`Yul` memory, raw `mstore` with memory-region metadata is probably simpler and closer to EVM semantics. But, a future `malloc`/`alloc` MIR op could be useful for alias analysis if it returns a pointer (and `size` & `alignment`), but I would not use it to hide fixed scratch slots or user-controlled `Yul` memory writes yet. So my preference is: keep metadata now, and to add `malloc` later only for free-memory allocations where it buys analyses something real.

crates/codegen/src/mir/inst.rs
djolertrk: I have added TODO for these, since it would touch a few building blocks, but I do agree!

crates/codegen/src/pass.rs
djolertrk: I agree, thanks!

crates/codegen/src/transform/cfg_simplify.rs
djolertrk: needed at some point, but now can be removed -- I am removing it

DaniPopes (approved):

PR #849, opened 2026-06-12
fix(codegen): various correctness bug fixes

djolertrk:
Also, introduce `-Zcodegen` option, so we can safely merge it with `main`.


djolertrk:
This also removes:

```
-//@ignore-host: windows
```

from ui tests (only left at 3 places, and will investigate why it fails there).

crates/codegen/src/analysis/cfg.rs
```diff
@@ -0,0 +1,246 @@
...
+            out.push(block);
+            current = self.idom(block).filter(|&idom| idom != block);
+        }
+        out
+    }
+}
+
+/// Returns blocks reachable from the entry.
+#[must_use]
+pub fn reachable_blocks(func: &Function) -> FxHashSet<BlockId> {
```
DaniPopes: can we cache these? also not a big fan of free functions in general, i know ai likes to spam these throughout but id rather have a centralized type thats queried so that you can abstract configs and other stuff you might want to take into account when computing this stuff; this is a more general thing that i'd like to clean up in all of the codegen stuff

crates/codegen/src/lower/expr.rs
```diff
@@ -788,7 +1297,814 @@ impl<'gcx> Lowerer<'gcx> {
...
+
+    /// Returns the constant value of a MIR immediate, if `value` is one.
+    fn const_u256_of(builder: &FunctionBuilder<'_>, value: ValueId) -> Option<U256> {
+        match builder.func().value(value) {
+            crate::mir::Value::Immediate(imm) => imm.as_u256(),
+            _ => None,
+        }
+    }
+
+    fn emit_panic_if_zero(&mut self, builder: &mut FunctionBuilder<'_>, value: ValueId, code: u64) {
```
DaniPopes: code: enum please

crates/codegen/src/lower/mod.rs
```diff
@@ -488,10 +481,62 @@ impl<'gcx> Lowerer<'gcx> {
                 }
                 total.max(1)
             }
+            // Fixed-size arrays occupy one slot per element (no packing),
+            // starting at the base slot. Dynamic arrays keep one length slot.
+            hir::TypeKind::Array(arr) if arr.size.is_some() => {
```
DaniPopes: (more general) hir::Type should not be used after typechecking

crates/codegen/src/mir/module.rs
```diff
@@ -105,10 +106,11 @@ pub struct StorageSlot {
     pub name: Option<Ident>,
 }
 
-/// A 32-byte immutable word appended to runtime code.
+/// A 32-byte immutable word staged in constructor scratch memory and patched
```
DaniPopes: (TODO) 32-byte hardcoding is arbitrary, it can be `<=32`:  https://github.com/argotorg/solidity/pull/16637
and is allowed by standardjson

crates/codegen/src/pass.rs
```diff
@@ -74,11 +75,26 @@ pub const INST_SIMPLIFY_PASS: PassInfo =
     });
 pub const CSE_PASS: PassInfo =
```
DaniPopes: let's have a macro for declaring these?

crates/codegen/src/pass.rs
```diff
@@ -109,20 +130,30 @@ pub const MEMORY_DSE_PASS: PassInfo =
     PassInfo::new("memory-dse", "Local dead memory-store elimination", || Box::new(MemoryDsePass));
 pub const DCE_PASS: PassInfo =
     PassInfo::new("dce", "Dead Code Elimination (fixed-point)", || Box::new(DcePass));
+pub const ADCE_PASS: PassInfo =
+    PassInfo::new("adce", "Aggressive dead-code elimination for dead control regions", || {
+        Box::new(AdcePass)
+    });
 
 /// All known MIR passes exposed to `solar mir-opt`.
 pub const PASS_REGISTRY: &[&PassInfo] = &[
```
DaniPopes:
```suggestion
pub const PASS_REGISTRY: &[PassInfo] = &[
```

crates/codegen/src/pass.rs
```diff
@@ -166,11 +203,17 @@ pub const DEFAULT_CLEANUP_PIPELINE: &[&PassInfo] = &[
     &PURE_EVAL_PASS,
```
DaniPopes: same here

```suggestion
    PURE_EVAL_PASS,
```

crates/codegen/src/pass.rs
```diff
@@ -253,6 +296,9 @@ fn run_cleanup_pipeline_to_fixpoint(
         let mut round_changed = false;
         for &pass in passes {
             let pass_changed = run_pass_with_options(module, pass, options);
+            if pass_changed && std::env::var_os("SOLAR_DEBUG_CLEANUP").is_some() {
```
DaniPopes: cli flag / remove? check codebase for more of this pattern

crates/codegen/src/utils/evm_word.rs
DaniPopes, line 1: we have some sort of const eval for type constants like array lengths. can we reuse?

crates/codegen/src/utils/cfg.rs
DaniPopes, line 1: maybe split utils and passes into their own modules

crates/codegen/src/transform/mod.rs
```diff
@@ -1,42 +1,57 @@
 //! Optimization and transformation passes for the Solar compiler.
 
+pub mod adce;
```
DaniPopes: these shouldnt be pub mod if everything's exported from them already

crates/solar/ice_sweep.rs
DaniPopes, line 1: probably remove

crates/solar/foundry_harness/mod.rs
DaniPopes, line 1: can we fold the foundry tests into the ./tests.rs as a different mode? maybe not possible

tests/ui/codegen/abi_decode_calldata_slice.sol
DaniPopes, line 1: I think we have a lot of codegen tests but little runtime tests. we should add runtime assertions using a mode in tester which calls an individual function and asserts the output like in solc test suite -- for follow ups

crates/codegen/src/lower/expr.rs
DaniPopes, line 1: these files are obviously very complex, im sure it could use some generic clean up prompts reducing redundancy, single-helper methods etc

tests/ui/codegen/abi_decode_calldata_slice.sol
djolertrk: more or less I did it with the external suite, but I completely agree -- it will be my next PR, and I think what it should be priority. thanks!

crates/codegen/src/analysis/cfg.rs
djolertrk: hmm I agree. I will introduce a hash map for this and will try to find other good candidates for refactoring as well.

djolertrk: will include:
- `CallGraphInfo` that can be used in the MIR transforms
- `LoopAnalyzer` now uses the shared `CfgInfo`
- Added central `Validator` and `PhiEliminator`
- also cleaned up liveness so passes use it through `AnalysisManager`

crates/codegen/src/lower/expr.rs
djolertrk: Yes! Will do it. Thanks.

crates/codegen/src/pass.rs
djolertrk: I guess `print_after_each` is enough for debugging, so I will remove it.

crates/codegen/src/pass.rs
djolertrk: yup

crates/codegen/src/utils/cfg.rs
djolertrk: yeah, especially once we introduce more analysis passes that will live out of transform/ dir, it will make even more sense, doing it now

crates/codegen/src/lower/expr.rs
djolertrk: I have refactored it a lot. Should be much much more intuitive now. Thanks!

crates/solar/ice_sweep.rs
djolertrk: it probably makes sense. I did found some bugs, but it should definitely done differently -- I am removing it

crates/codegen/src/utils/evm_word.rs
djolertrk: yup, I used as much as possible api from it, but kept evm specific stuff in utils/evm_word

crates/codegen/src/mir/module.rs
djolertrk: uh, yeah, will add a few TODOs here, thanks

crates/solar/foundry_harness/mod.rs
djolertrk: hmm, let me check this, would be super useful

djolertrk: I am adding it in this PR, along with a CI/CD action

crates/codegen/src/lower/type_query.rs
```diff
@@ -0,0 +1,321 @@
...
+            (TyKind::Struct(s1), TypeKind::Custom(hir::ItemId::Struct(s2))) => s1 == s2,
+            // Enum types
+            (TyKind::Enum(e1), TypeKind::Custom(hir::ItemId::Enum(e2))) => e1 == e2,
+            _ => false,
+        }
+    }
+
+    /// Gets struct info for an expression if it has a struct type.
+    /// Returns (struct_id, field_count) if the expression is a struct.
+    pub(super) fn get_expr_struct_info(
```
DaniPopes, line 108: these files are filled to the brim with basically the same logic as typeck still. the solution is having TypeckResults where we sparsely store ExprId->whatever info we need from typeck, such as call overrides, method resolution (including UDVTs for binops) etc. codegen shouldn't care much that a type is a udvt, overrides and all that high level stuff.

djolertrk: hmm, yeah, lets expose that in sema so later phases/codegen can use it. I tried not to touch it as much as possible, but yeah, that is the clean approach

DaniPopes (approved):

PR #854, opened 2026-06-15
fix: clean up ci

djolertrk:


.github/workflows/bench.yml
```diff
@@ -175,6 +176,7 @@ for t in r['results']:
       - name: Build the benchmark target(s)
         run: cargo codspeed build --profile profiling -p solar-bench --bench criterion
       - name: Run the benchmarks
+        continue-on-error: true
```
DaniPopes: why this? it's the last step

.github/workflows/ci.yml
```diff
@@ -46,6 +46,11 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
       - uses: taiki-e/install-action@nextest
+      - name: Install Foundry
+        if: matrix.os == 'ubuntu-latest' && matrix.rust == 'stable'
```
DaniPopes: why not the entire matrix? if we need to skip we can do `if matrix.foundry`

djolertrk: yeah, lets do it.

.github/workflows/bench.yml
djolertrk: removed

PR #859, opened 2026-06-17
fix(bench): preserve single-threaded session

DaniPopes:
Keep the benchmark session single-threaded after installing the codegen/typeck options.

The codegen merge added an explicit `opts(...)` call after `single_threaded()`, which replaced the whole option set and restored the default thread count. Move `single_threaded()` after `opts(...)` so the Criterion benchmarks keep their intended one-thread configuration while still enabling the codegen-related flags.


mmsaki:
insane performance for a 1 line fix

PR #868, opened 2026-06-19
feat: introduce EVM IR

djolertrk:
Introduce an IR at lower level of abstraction than MIR.


crates/codegen/src/backend/evm/ir.rs
```diff
@@ -0,0 +1,1949 @@
...
+/// Block-level metadata.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct EvmIrBlockMetadata {
+    /// Estimated block hotness for future layout and scheduling decisions.
+    pub hotness: EvmIrBlockHotness,
+}
+
+/// Block hotness metadata.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub enum EvmIrBlockHotness {
```
DaniPopes, line 136: Hot/cold is enough. should be displayed only if cold, hot default

crates/codegen/src/backend/evm/ir.rs
```diff
@@ -0,0 +1,1949 @@
...
+}
+
+newtype_index! {
+    /// A unique identifier for a stack value in EVM IR.
+    pub struct EvmIrValueId;
+}
+
+/// An EVM IR module.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct EvmIrModule {
```
DaniPopes, line 35: do we need this high level hierarchy? i was thinking this is after function calls so you have just asm basic blocks, since there's no "call" instruction. and terminators are just jump or terminating instructions

DaniPopes:
can we move peepholes to a pass? can add other passes after

djolertrk:
> can we move peepholes to a pass? can add other passes after

hmm, yeah, lets see how it fits

crates/codegen/src/backend/evm/ir.rs
djolertrk: yes yes! will add these inputs to AGENTS.md as well.

djolertrk:
> > can we move peepholes to a pass? can add other passes after
> 
> hmm, yeah, lets see how it fits

I will rewrite the peepholes to operate in their own pass (`PeepholePass`), and it can be modular like whenever we want to add a pass that operates with row/very end asm, we just extend `run_assembler_passes`.

`EvmIrPass` is still at a bit higher level of abstraction, and the current peepholes are not the best match for our EVM IR, since they operate on raw asm (final DUP/SWAP/PUSH etc.). At EVM IR we could do different opts that are very similar to it, like simplifying EVM IR instructions/branches before stack scheduling...

tests/ui/codegen/evm-ir/dispatch.evmir
```diff
@@ -0,0 +1,12 @@
+bb0 (entry):
+  %selector = calldataload 0 !meta(stack=1->1, source=dispatch)
+  %is_inc = eq %selector, 0x371303c0 !meta(stack=2->1)
+  br %is_inc, bb1, bb2
+bb1:
+  %slot = push 0 !meta(stack=0->1)
```
DaniPopes: i dont think the meta is needed to be displayed on every instructions

tests/ui/codegen/evm-ir/dispatch.evmir
```diff
@@ -0,0 +1,12 @@
...
+  %value = sload %slot !meta(stack=1->1, alias=storage0)
+  jump bb3
+bb2 [cold]:
+  revert 0, 0 !meta(stack=2->0, source=fallback)
```
DaniPopes: push instruction exists but constants are still allowed as values to other instructions, should it not be one or the other?

djolertrk: agreed. fixed

DaniPopes (approved):

PR #870, opened 2026-06-21
feat(lsp): add initial language server support

0xKarl98:
Supersedes #401  , towards OSS-403

It carries forward the original #401 design: a basic `solar lsp` server, a thin VFS over files on disk, rope-backed document
contents, full-project reparse/relower on changes, and diagnostic publication back to the client.

On top of #401, this branch:
- updates the work against current `main`
  - uses the `solar lsp --stdio` path expected by editor clients
  - registers watched-file notifications for project/file changes
  - publishes parser and semantic diagnostics through `textDocument/publishDiagnostics`
  - clears stale diagnostics when errors are fixed
  - rechecks disk contents when files are closed, so saved errors stay visible while unsaved buffer-only errors are cleared
  - handles UTF-16 LSP ranges correctly when applying incremental edits, including files with non-ASCII text such as Chinese characters or emoji
  - avoids stale diagnostics from older analysis runs overwriting newer editor state
  - includes focused LSP tests for watched-file routing, diagnostic cleanup, disk-backed close handling, and UTF-16 incremental edits
  
 @DaniPopes


crates/lsp/src/global_state.rs
```diff
@@ -0,0 +1,313 @@
...
+        std::future::ready(Ok(InitializeResult {
+            capabilities,
+            server_info: Some(ServerInfo {
+                name: "solar".into(),
+                version: Some(SHORT_VERSION.into()),
+            }),
+        }))
+    }
+
+    pub(crate) fn on_initialized(&mut self, _: InitializedParams) -> NotifyResult {
```
mattsse, line 70: The server handles `workspace/didChangeWatchedFiles`, but I do not see it registering file watchers with the client. Most clients will not send those notifications just because the handler exists, so disk `.sol` edits and `foundry.toml` changes will usually never call `did_change_watched_files` / `rediscover_workspaces`.

Suggested fix: in `on_initialized`, dynamically register `workspace/didChangeWatchedFiles` for `**/*.sol` and `**/foundry.toml` via `client/registerCapability` when the client supports dynamic registration. If we want to keep this PR to initial open-buffer diagnostics only, that seems fine as a follow-up, but then we should remove/soften the PR claim that watched-file notifications are registered and track the watcher registration separately.

mattsse (commented):
I left one concern around watched-file notifications. I do not think this necessarily has to block the first LSP scaffolding if we intentionally scope disk/project-change support into a follow-up, but then we should make that explicit and avoid claiming that watched-file notifications are registered in this PR. If the PR is meant to support disk `.sol` / `foundry.toml` changes already, this should be fixed before merge.

mattsse (approved):
Approving. The initial LSP scaffolding looks good to land, and we can tackle dynamic watched-file registration for disk `.sol` / `foundry.toml` changes in a follow-up.

mattsse (approved):

DaniPopes (approved):

DaniPopes (approved):

Ari4ka (approved):

PR #880, opened 2026-06-24
feat(lsp): register watched file notifications

0xKarl98:
Towards OSS-403 .

Registers `workspace/didChangeWatchedFiles` dynamically when the client advertises support for watched-file dynamic registration.

This lets the language server ask the client to watch Solidity sources and Foundry manifests:

  - `**/*.sol`
  - `**/foundry.toml`

With these registrations in place, saved Solidity file changes can trigger diagnostics recomputation even when the file is not open in the editor, and `foundry.toml` changes can trigger workspace rediscovery before recomputing diagnostics.

Also adds focused coverage for capability negotiation, watched-file registration parameters, and the initialized LSP flow that sends the dynamic registration request.

cc @DaniPopes


mattsse (approved):
LGTM

DaniPopes (approved):

PR #881, opened 2026-06-24
  feat(lsp): load workspace config across lifecycle events

0xKarl98:
Closes OSS-403

Loads Foundry workspace configuration in the LSP and applies it to background analysis. Workspace discovery now runs during initialization, workspace folder changes, configuration changes, and watched-file updates for `foundry.toml` and Solidity files.

For each discovered `foundry.toml`, the LSP reads the default Foundry profile and converts the relevant settings into compiler options: project base path, source root, library include paths, remappings, and optional EVM version. If no manifest is found, or if a manifest fails to load, the root is kept as a naked workspace so analysis can continue instead of dropping the folder.

Analysis now builds per-workspace batches. Each batch uses the workspace's compile options, scans the configured source roots, overlays open in-memory documents on top of disk files, and assigns files to the most specific workspace when roots overlap. This lets imports resolve with the correct Foundry remappings while preserving support for unconfigured folders.

This PR also includes several LSP lifecycle and performance improvements:

- Centralize workspace refresh logic so initialization, configuration changes, workspace folder changes, and manifest file changes all rebuild the same workspace model.
- Cache discovered Solidity source files per workspace, then update that cache on watched file create/delete events instead of recursively rescanning source roots on every recompute.
- Skip stale background analysis work before it publishes diagnostics, so newer edits are not blocked by superseded runs.
- Clear stale diagnostics while avoiding repeated empty `publishDiagnostics` notifications for files that are already clean.
- Bound naked workspace discovery by skipping common heavy directories such as `.git`, `cache`, `lib`, `node_modules`, and `out`, while still analyzing a root explicitly opened with one of those names.

cc @DaniPopes


crates/lsp/src/config.rs
```diff
@@ -93,9 +157,15 @@ pub(crate) fn negotiate_capabilities(params: InitializeParams) -> (ServerCapabil
 
 #[cfg(test)]
 mod tests {
-    use lsp_types::{DidChangeWatchedFilesClientCapabilities, WorkspaceClientCapabilities};
+    use std::fs;
+
+    use lsp_types::{
+        DidChangeWatchedFilesClientCapabilities, WorkspaceClientCapabilities, WorkspaceFolder,
+    };
+    use tempfile::TempDir;
```
DaniPopes, line 157: please keep all imports in one group

crates/lsp/src/global_state.rs
```diff
@@ -283,22 +375,24 @@ mod tests {
             Some(serde_json::json!({
                 "watchers": [
                     { "globPattern": "**/*.sol", "kind": WatchKind::Create | WatchKind::Change | WatchKind::Delete },
+                    { "globPattern": "**/solar.toml", "kind": WatchKind::Create | WatchKind::Change | WatchKind::Delete },
```
DaniPopes: we don't have a stable config format, let's just keep foundry

0xKarl98: Have removed it

crates/lsp/src/workspace/manifest.rs
```diff
@@ -1,70 +1,113 @@
...
     path::{Path, PathBuf},
 };
 
-use solar_interface::data_structures::map::rustc_hash::FxHashSet;
-use tokio::io;
-
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
 pub(crate) enum ProjectManifest {
+    // todo: guarantee this to be absolute
+    Solar(PathBuf),
```
DaniPopes: please remove solar config entirely

0xKarl98: Done and rebased

DaniPopes:
windows ci

crates/lsp/src/workspace/foundry.rs
```diff
@@ -1,15 +1,91 @@
...
+
+    pub(crate) fn include_paths(&self, root: &Path) -> Vec<PathBuf> {
+        match &self.libs {
+            Some(libs) => libs.iter().map(|path| root.join(path)).collect(),
+            None => vec![root.join("lib")],
+        }
+    }
+
+    pub(crate) fn remappings(&self, root: &Path) -> Vec<ImportRemapping> {
+        let mut remappings = self.discover_lib_remappings(root);
```
mattsse: This always auto-detects lib remappings, even when the project sets `auto_detect_remappings = false` in `foundry.toml`. That means LSP import resolution can diverge from `forge` for projects that intentionally disable auto-detection. Can we parse that flag and skip `discover_lib_remappings` when it is false?

crates/lsp/src/workspace/manifest.rs
```diff
@@ -1,70 +1,102 @@
...
+    fs::read_dir,
+    io,
     path::{Path, PathBuf},
 };
 
-use solar_interface::data_structures::map::rustc_hash::FxHashSet;
-use tokio::io;
-
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
-pub(crate) enum ProjectManifest {
```
DaniPopes, line 10: this file can be mostly restored i think

0xKarl98: Yes

DaniPopes (approved):

Ari4ka (approved):

PR #888, opened 2026-06-26
feat(lsp): extract declaration symbol tables

0xKarl98:
Towards oss-404  

This copies source-level declaration data out of the compiler analysis run so later LSP request handlers can query symbols without holding HIR references. The table records declaration names, kinds, locations, name ranges, file membership, and parent symbol links for lexical scope structure.

## Follow-up work
This does not yet wire the table into user-facing LSP requests. Follow-up PRs can build on it for document symbols, workspace symbols, go to definition/declaration, references, completion, and cursor offset lookup.

cc @DaniPopes


mattsse (approved):

crates/lsp/src/global_state.rs
```diff
@@ -328,7 +346,8 @@ fn analyze(batch: AnalysisBatch) -> FxHashMap<Url, Vec<Diagnostic>> {
         Ok(())
     });
 
-    compiler.enter(|compiler| {
+    let symbol_tables = compiler.enter(|compiler| SymbolTables::build(compiler.gcx()));
+    let diagnostics = compiler.enter(|compiler| {
```
DaniPopes: you can do all this inside of the enter_mut closure above

crates/lsp/src/global_state.rs
```diff
@@ -719,4 +751,105 @@ mod tests {
 
         assert!(batch.files.is_empty());
     }
+
+    #[test]
+    fn analyze_builds_declaration_symbol_table() {
+        let path = std::env::temp_dir()
```
DaniPopes: please use tempfile crate here and everywhere else std::env::temp_dir is used if any

crates/solar/foundry_harness/mod.rs
```diff
@@ -146,26 +147,43 @@ fn foundry_solc() -> PathBuf {
                 if #[cfg(unix)] {
                     use std::os::unix::fs::PermissionsExt;
 
-                    let path = std::env::temp_dir()
-                        .join(format!("solar-zcodegen-{}.sh", std::process::id()));
+                    let file = tempfile::Builder::new()
```
DaniPopes: i fixed this on main

DaniPopes (approved):

PR #906, opened 2026-06-29
feat(lsp): add symbol providers

0xKarl98:
Towards oss-404

This adds document and workspace symbol support to the LSP server. The server now advertises both symbol provider capabilities during initialization, builds an LSP-owned declaration table from HIR analysis results, and serves nested document symbols for editor outlines and workspace symbols for project-wide symbol search.

The symbol table records source declarations, source ranges, symbol kinds, and parent links so clients can display Solidity-aware structure for contracts, interfaces, libraries, functions, variables, structs, enums, events, errors, and enum variants. Generated getter artifacts are filtered out so editor navigation reflects user-written source declarations rather than compiler-generated HIR items.
 
### Result screenshot 

document symbol 
833ccf5e9e2bf6d2c1965e67c2639926

workspace symbol :
be30ef6ddf1996a421a64c3a11bf8223

 cc @DaniPopes


crates/lsp/src/symbols.rs
```diff
@@ -175,13 +401,22 @@ fn declaration_name(gcx: Gcx<'_>, item_id: ItemId) -> Option<(String, Span)> {
...
 ///
 /// This keeps the LSP-facing kind independent from the compiler's HIR enum while preserving the
 /// source declaration category needed by document symbols, workspace symbols, and completion.
-fn declaration_kind(item_id: ItemId) -> DeclarationKind {
+fn declaration_kind(gcx: Gcx<'_>, item_id: ItemId) -> DeclarationKind {
     match item_id {
-        ItemId::Contract(_) => DeclarationKind::Contract,
+        ItemId::Contract(id) => {
+            let contract = gcx.hir.contract(id);
+            if contract.kind.is_interface() {
```
DaniPopes: match on the kind

crates/lsp/src/symbols.rs
```diff
@@ -217,16 +452,120 @@ fn is_generated_item(gcx: Gcx<'_>, item_id: ItemId) -> bool {
 /// link so LSP features can traverse the declaration scope tree without holding HIR references.
 fn parent_item(gcx: Gcx<'_>, item_id: ItemId) -> Option<ItemId> {
```
DaniPopes: we can move this to ItemId `.parent(self, hir) -> Option<Self> {hir.item(self).parent()}`, and Item `.parent(self) -> Option<ItemId>`

0xKarl98: Have applied !

DaniPopes (approved):

PR #909, opened 2026-06-30
feat(lsp): add symbol navigation index

0xKarl98:
Towards closing OSS-404

This expands the LSP symbol table from document and workspace symbols into a shared index for editor navigation. It records source declarations, lexical scopes, and resolved references from semantic analysis so the server can answer go-to definition, go-to declaration, and find references with Solidity-aware targets such as members, using directives, overloads, enum variants, and source-level declarations.

Generated compiler artifacts such as public getter functions are filtered from user-facing symbol lists so editor symbol views stay focused on declarations that exist in source.

cc @DaniPopes

### DIFF
--- a/crates/lsp/src/symbols.rs
+++ b/crates/lsp/src/symbols.rs
@@ -48,6 +48,7 @@ pub(crate) struct DeclarationSymbol {
     pub(crate) location: Location,
     pub(crate) name_range: Range,
     pub(crate) parent: Option<SymbolId>,
+    has_definition: bool,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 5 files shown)


crates/lsp/src/symbols.rs
```diff
@@ -229,14 +312,337 @@ impl SymbolTables {
...
+
+            let root = self.push_scope(uri, range, None);
+            for &item_id in source.items {
+                self.add_scope_declaration(root, item_id);
+                self.add_child_scopes(gcx, item_id, root);
+            }
+        }
+    }
+
+    fn add_child_scopes(&mut self, gcx: Gcx<'_>, item_id: ItemId, parent: ScopeId) {
```
DaniPopes: maybe better to implement as a visitor?

0xKarl98: Have built a ScopeBuilder that uses the HIR visitor to construct LSP lexical scopes

DaniPopes (approved):

PR #910, opened 2026-07-01
feat(lsp): add autocomplete

0xKarl98:
Towards closing oss-405 

cc @DaniPopes


crates/lsp/src/global_state.rs
```diff
@@ -934,6 +934,196 @@ mod tests {
         assert!(!labels.contains(&"nextValue"), "{labels:?}");
     }
 
+    #[test]
+    fn analyze_completes_imported_symbols_and_builtins_in_scope() {
+        let project = TestProject::from_fixture(
```
DaniPopes: i would like to see a test harness like in rust analyzer where you can use `$1` to indicate cursor position inside of the source string directly, and then assert the output of, for example, auto complete or go to definition with a snapshot test. the tests in rust-analyzer are very compact, like
```rust
check(
  "
source string
here.$1
",
  str![[r#"snapshot"#]]
);
```

i would also split up the tests into separate files for each request/group of requests

mattsse (commented):
Thanks for the LSP autocomplete work. I found two behavior gaps around dirty/open-buffer completions:

1. Partial member prefixes fall back to lexical completions

[`completion_items_with_line`](https://github.com/paradigmxyz/solar/blob/2b837fac6c968abfd96d721b0bf65e2a03fb6312/crates/lsp/src/symbols.rs#L425-L439) only reaches the live-line receiver fallback when [`trailing_member_receiver`](https://github.com/paradigmxyz/solar/blob/2b837fac6c968abfd96d721b0bf65e2a03fb6312/crates/lsp/src/symbols.rs#L1602-L1605) sees a line prefix ending in `.`. In a typical completion flow, after the user types `data.f|`, the live open buffer no longer ends with `.`, so the server falls back to lexical completions unless the stale analyzed AST already has a member span at that exact position. I reproduced this locally with a temporary fixture where the analyzed source was valid `data;` and the open VFS source was `data.f$1`; the response was contract/global completions instead of the struct fields.

Suggested fix: parse both `receiver` and member prefix from the live line, e.g. `member_access_prefix(line, character) -> Option<(&str, &str)>`, where the prefix may be empty. Then use the receiver to fetch the member set for both `data.$1` and `data.f$1`, and either let the client filter by prefix or prefilter labels server-side.

2. Dirty completions for builtin modules return globals

[`member_completion_set_for_res`](https://github.com/paradigmxyz/solar/blob/2b837fac6c968abfd96d721b0bf65e2a03fb6312/crates/lsp/src/symbols.rs#L674-L682) only builds receiver member sets for `Res::Item(_)` and `Res::Namespace(_)`. Scoped builtin modules like `msg`, `block`, `tx`, and `abi` are therefore added to lexical completions without a member set. When the open buffer contains `msg.|`, the fallback finds `msg` in [`member_completion_items_for_receiver`](https://github.com/paradigmxyz/solar/blob/2b837fac6c968abfd96d721b0bf65e2a03fb6312/crates/lsp/src/symbols.rs#L850-L872), hits `declaration.member_completion_ids?`, and falls back to lexical/global completions. I reproduced `msg.$1` returning globals instead of `sender`, `value`, `data`, and `sig`.

Suggested fix: include `Res::Builtin(builtin)` in member set construction when `builtin.members().is_some()`, using `gcx.member_completions_of(gcx.type_of_res(res), ...)` for those module builtins. Handle `this` and `super` separately because `Gcx::type_of_res` intentionally routes those through typeck's context-aware path rather than the generic builtin path; either skip them here or add a context-aware helper for completion receiver types.

The focused completion tests pass (`cargo test -p solar-lsp completion`). These two gaps should be covered with dirty completion fixtures for `data.f$1` and `msg.$1`.

0xKarl98:
@mattsse  Have fixed both of the cases

crates/lsp/src/global_state/tests/completion.rs
```diff
@@ -0,0 +1,530 @@
...
+    "this | function | builtin",
+    "tx | module | builtin",
+];
+
+fn completion_item_kind(kind: Option<CompletionItemKind>) -> &'static str {
+    let Some(kind) = kind else {
+        return "-";
+    };
+
+    if kind == CompletionItemKind::CLASS {
```
DaniPopes: match ?

crates/lsp/src/global_state/tests/completion.rs
```diff
@@ -0,0 +1,530 @@
...
+        //- /Members.sol open
+        contract C {
+            struct Data { uint256 field; uint256 other; }
+
+            function read(Data memory data) public pure {
+                data.$1;
+            }
+        }
+        "#,
+        |dirty_source| dirty_source.replace("data.", "data"),
```
DaniPopes, line 376: ideally we dont need to "fixup" the source, and we can still service requests even when the file doesn't parse, but this might be a larger change

crates/lsp/src/config.rs
```diff
@@ -141,7 +141,10 @@ pub(crate) fn negotiate_capabilities(params: InitializeParams) -> (ServerCapabil
 
     (
         ServerCapabilities {
-            completion_provider: Some(CompletionOptions::default()),
+            completion_provider: Some(CompletionOptions {
+                trigger_characters: Some(vec![".".into()]),
```
DaniPopes, line 145: hardhat lists other stuff https://github.com/NomicFoundation/hardhat-vscode/blob/03d942869dd9601df65db37efc65ac3bb5c21d7d/server/src/services/initialization/onInitialize.ts#L73 dont know how useful they are

crates/lsp/src/global_state.rs
DaniPopes, line 1: let's port the other tests to the new format as well. in general it's best to avoid hard coding stuff like calling "analysis" etc when we can express these tests as snapshot tests, like we do in foundry and other places in this repo too

crates/sema/src/ty/mod.rs
```diff
@@ -106,6 +106,22 @@ pub enum ResolvedMember {
...
+/// A source-level declaration visible from semantic resolution.
+#[derive(Clone, Copy, Debug)]
+pub struct ScopeDeclaration {
+    pub name: Symbol,
+    pub res: hir::Res,
+}
+
+/// A completion candidate for a member visible on a receiver type.
+#[derive(Clone, Copy, Debug)]
+pub struct MemberCompletion {
```
DaniPopes: this should be `{ member: Member, resolved: ... }`

crates/sema/src/ty/mod.rs
```diff
@@ -858,6 +874,86 @@ impl<'gcx> Gcx<'gcx> {
         native.iter().copied().chain(attached)
     }
 
+    pub fn member_completions_of(
+        self,
+        ty: Ty<'gcx>,
+        source: hir::SourceId,
+        contract: Option<hir::ContractId>,
+    ) -> Vec<MemberCompletion> {
```
DaniPopes:
```suggestion
    ) -> impl Iterator<Item = MemberCompletion> {
```

crates/lsp/src/handlers/reqs.rs
```diff
@@ -67,8 +67,18 @@ pub(crate) fn completion(
...
+        state.vfs.read().get_file_contents(&path).and_then(|contents| {
+            (params.position.line < contents.line_len() as u32)
+                .then(|| contents.line(params.position.line as usize).to_string())
+        })
+    });
+    let symbol_tables = state.symbol_tables.read();
+    let items = if let Some(line) = line.as_deref() {
+        symbol_tables.completion_items_with_line(&params.text_document.uri, params.position, line)
+    } else {
+        symbol_tables.completion_items(&params.text_document.uri, params.position)
```
DaniPopes: let's have completion_items with option param instead of two fns

crates/lsp/src/handlers/reqs.rs
```diff
@@ -67,8 +67,18 @@ pub(crate) fn completion(
     params: CompletionParams,
 ) -> impl Future<Output = Result<Option<CompletionResponse>, ResponseError>> + use<> {
     let params = params.text_document_position;
-    let items =
-        state.symbol_tables.read().completion_items(&params.text_document.uri, params.position);
+    let line = proto::vfs_path(&params.text_document.uri).and_then(|path| {
```
DaniPopes, line 70: why only line and not location?

DaniPopes (changes requested):

crates/lsp/src/config.rs
0xKarl98: Maybe we can leave these additional triggerCharacters to follow-up import auto-completions , if you think we need this kind of auto-completions

crates/lsp/src/handlers/reqs.rs
0xKarl98: uri + position already gives us the location. The extra line is not for locating the cursor, it’s for reading the text around it

0xKarl98:
Close this as this auto-complete method cann't cover all dirty cases well

PR #912, opened 2026-07-02
fix(docs): avoid rustdoc ICE on re-exports

mattsse:
Marks third-party convenience re-exports that currently trigger rustdoc to inline external crate docs as `#[doc(no_inline)]`.

`nightly-2026-07-02` ICEs while documenting `solar-compiler` when it walks through the inlined crate docs and reaches `bumpalo::core_alloc` or `smallvec::alloc`. Keeping these external re-exports as links preserves the public API and lets the component crate docs remain inlined.

### DIFF
--- a/crates/ast/src/lib.rs
+++ b/crates/ast/src/lib.rs
@@ -6,6 +6,7 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
 // Convenience re-exports.
+#[doc(no_inline)]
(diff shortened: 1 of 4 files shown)


mablr (approved):
Makes sense

0xKarl98 (approved):

DaniPopes (approved):

Ari4ka (approved):

PR #915, opened 2026-07-03
feat(lsp): add auto-completion support

0xKarl98:
Towards OSS-405 

This adds LSP completion support for Solidity member access. Completion now uses semantic receiver types, so it works beyond simple identifier receivers and accurately covers editing edge cases such as `getToken().bal`, `(this.token()).bal`, `tokens[i].bal`, `foo.`, and multiline accesses like `foo` followed by `.bal` on the next line.

The implementation also completes built-in members, empty member accesses after `.`, and partial member prefixes while filtering out unrelated global completions in member contexts. Parser and type-checking recovery were tightened so incomplete member expressions like `s.` and `s.()` still produce the original parse diagnostic without cascading into unrelated type-checking or member-not-found errors.

Also covers edge cases mentioned in #910 

cc @DaniPopes


crates/lsp/src/global_state/tests/support.rs
DaniPopes, line 1: can we make all tests non-async? we dont make use of async anyway

0xKarl98: Noticed that , have applied

DaniPopes (approved):

PR #918, opened 2026-07-05
feat(lsp): add inlay hints

0xKarl98:
Towards oss-407.

This adds inlay hints for Solidity calls in the LSP: parameter names for positional arguments and type hints for useful call results.

The hints lean on semantic call resolution, so they follow selected overloads, `using` calls, constructors, events, errors, modifiers, and `abi.encodeCall`. They also avoid noisy cases like named arguments, repeated obvious names, explicit casts, unit results, and malformed `abi.encodeCall` tuples.

cc @DaniPopes 
##### No hints for inline assembly :
截屏2026-07-07 18 02 39

##### ABI-encoded calls 
截屏2026-07-07 15 55 08

##### Normal calls
截屏2026-07-07 15 56 22

#####  Overload 
截屏2026-07-07 16 06 43

### DIFF
--- a/crates/lsp/src/inlay_hints.rs
+++ b/crates/lsp/src/inlay_hints.rs
@@ -6,7 +6,8 @@ use solar_interface::{
 };
 use solar_sema::{
     Gcx,
-    hir::{self, CallArgsKind, ExprKind, ItemId, Visit},
+    builtins::Builtin,
+    hir::{self, CallArgsKind, ExprKind, ItemId, StmtKind, Visit},
     ty::{CallableParamSource, Ty, TyKind},
 };
 use std::ops::ControlFlow;
@@ -137,8 +138,16 @@ impl<'gcx> InlayHintCollector<'gcx> {
         let (CallArgsKind::Unnamed(exprs), Some(param_source)) = (args.kind, param_source) else {
             return;
         };
+        self.push_parameter_hints_for_exprs(exprs, param_source);
+    }
+
+    fn push_parameter_hints_for_exprs(
+        &mut self,
+        exprs: impl IntoIterator<Item = &'gcx hir::Expr<'gcx>>,
+        param_source: CallableParamSource,
+    ) {
         let param_names = self.gcx.callable_param_names(param_source);
-        for (arg, param_name) in exprs.iter().zip(param_names) {
+        for (arg, param_name) in exprs.into_iter().zip(param_names) {
             let Some(param_name) = param_name else {
                 continue;
             };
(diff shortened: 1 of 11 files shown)


DaniPopes (approved):
screenshot plz?

decofe:
Expanded screenshot with a few more inlay hint cases from this PR: struct constructor fields, regular call parameters, multi-return type hints, modifier arguments, event args, and custom error args.

Solar LSP inlay hints

DaniPopes:
the `struct Transfer` part looks weird pls look, check also what hardhat does

0xKarl98:
截屏2026-07-07 15 42 25
Have fixed the lints return by struct here , aligned with hardhat @DaniPopes

0xKarl98:
Also have fixed cargo deny check CI in this pr together

0xKarl98:
> screenshot plz?

Have added some in description

DaniPopes:
I would rather not display a hint for function calls tbh, ig if hardhat does it it's fine

pls revert [`6e197fe` (this PR)](https://github.com/paradigmxyz/solar/pull/918/commits/6e197fe51e80cb3723b0a61a8fd789b615382e23) our format looks better

0xKarl98:
> I would rather not display a hint for function calls tbh, ig if hardhat does it it's fine

Yeah i was follow the format that hardhat does , but we can surely revert it

crates/lsp/src/inlay_hints.rs
```diff
@@ -0,0 +1,400 @@
...
+        match arguments.kind {
+            ExprKind::Tuple(components) => {
+                if components.iter().any(|component| component.is_none()) {
+                    return;
+                }
+                self.push_parameter_hints_for_exprs_and_names(
+                    components.iter().filter_map(|component| *component),
+                    param_names,
+                );
+            }
```
mattsse, line 198: Malformed `abi.encodeCall` tuples can still get partial parameter hints here: `(1, user, 3)` zips the first two components with the target parameter names even though type checking rejects the call. I think we should suppress hints unless the tuple arity matches the target signature.

```suggestion
            ExprKind::Tuple(components) => {
                if components.len() != param_names.len()
                    || components.iter().any(|component| component.is_none())
                {
                    return;
                }
                self.push_parameter_hints_for_exprs_and_names(
                    components.iter().filter_map(|component| *component),
                    param_names,
                );
            }
```

mattsse (commented):
One small correctness suggestion for malformed `abi.encodeCall` tuples.

0xKarl98:
@mattsse  Have applied this mismatch case , where  :
截屏2026-07-07 22 00 49

0xKarl98:
@DaniPopes  You said struct Transfer is weird earlier and then request to  revert the changes made , so i just let it not show hints here , i.e.

截屏2026-07-07 22 06 37

DaniPopes (approved):

PR #921, opened 2026-07-06
feat(codegen)!: phased MIR lowering and a leaner backend

djolertrk:
Mostly focused to resolving the OSS-433, by addressing:

1. No values for void insts
2. No raw Res slices
3. Error sentinel
4. Less special-casing
5. MIR phases

MIR is now a phased IR. A module carries a `MirPhase` (`built` -> `optimized` ->  `abi` -> `dispatch` -> `evm-shaped`) and behavior that used to happen ad hoc during lowering or inside the backend became explicit passes: `lower-abi` turns each word-sized external function into a self-decoding wrapper that calls the body as an internal function, `lower-dispatch` synthesizes the selector switch as an ordinary MIR entry routing through the new `tail_call` terminator, and the `EVM-shape` pass converts non-returning `internal calls` into `tail calls`. The pipeline is on by default (`-Zno-mir-dispatch` opts out), and the phase round-trips through the MIR text format with UI tests pinning each transition.

As mentioned above, also the remaining OSS-433 subissues are resolved alongside: void instructions (`mstore`, `logs`, `void builtins`) no longer allocate values; lowering resolves callees and members through typeck results instead of raw `&[Res]` slices, so overloading is handled uniformly; error handling collapses onto a `Value::Error(ErrorGuaranteed)` sentinel instead of the previous mix of zeros, Nones and panics; and constant folding moved out of lowering into the MIR passes (`keccak256("literal")` folds there; folding it through a memory variable needs the content tracking from OSS-413).

Also, on top of the phased pipeline sit the code-size changes (since I noticed that a bunch of contracts from our corpus were not deployable due to code size): `static frame` overlays give non-recursive functions compile-time frame addresses, return addresses travel on the EVM stack, `--libraries` links public library calls as delegatecalls (like solc), repeated `revert/return/storage-string` sequences share one synthesized helper per module, and the assembler dedupes identical terminal spans across functions. Measured end to end:
- nitro OneStepProofEntry 23,292 to 16,462 bytes
- aave BorrowLogic 37,825 → 21,636; and now every deployable aave artifact fits EIP-170 (`Pool` goes 92,764 inlined to 21,329 linked)


crates/codegen/testdata/arithmetic/cache-solar/solidity-files-cache.json
DaniPopes, line 1: please rm `testdata/**/{cache,out}*`, the test runner should generate temp dirs for these already, and not inside of the repo

DaniPopes (changes requested):

crates/codegen/testdata/arithmetic/cache-solar/solidity-files-cache.json
djolertrk: yeah, done

crates/codegen/src/lower/mod.rs
```diff
@@ -551,6 +581,94 @@ impl<'gcx> Lowerer<'gcx> {
...
+
+    /// Returns the module's shared storage-`bytes`/`string` load helper,
+    /// synthesizing it on first use: takes the slot, decodes the packed
+    /// short/long form, and returns a fresh `[length][data...]` memory copy.
+    /// Marked `no_inline` — the whole point is existing once per module.
+    pub(super) fn ensure_load_storage_bytes_helper(&mut self) -> FunctionId {
+        if let Some(id) = self.storage_bytes_helper {
+            return id;
+        }
+        let name = Ident::new(Symbol::intern("__load_storage_bytes"), Span::DUMMY);
```
DaniPopes: please add values to the pre-interned set of symbols, not call Symbol::intern inline. add this as an AGENTS.md rule

tests/ui/cli/Zhelp.stdout
```diff
@@ -50,6 +50,9 @@ Options:
           
           Off by default: MIR and bytecode output is only produced when this is set. Codegen is a work in progress and not yet part of the compiler's stable, solc-compatible behavior.
 
+      -Zno-mir-dispatch
+          Disable MIR-phase dispatch lowering: generate the dispatcher inside the EVM backend instead of lowering dispatch and ABI handling as MIR phases. By default the phases run and the backend consumes them, falling back to the backend dispatcher per module when it cannot be fully lowered
```
DaniPopes: why would a module not be able to be lowered

crates/codegen/src/mir/module.rs
```diff
@@ -17,6 +17,70 @@ use solar_interface::Ident;
...
+    /// The wrapper keeps its selector but takes no MIR arguments.
+    Abi,
+    /// The selector switch has been materialized as an ordinary MIR `entry`
+    /// function that routes to the ABI wrappers, instead of being generated
+    /// inside the backend.
+    Dispatch,
+    /// Builtin-level operations are expanded and functions take the shape the
+    /// backend expects. Reserved: nothing lowers to this phase yet, since
+    /// builtins already lower to concrete `InstKind`s.
+    EvmShaped,
```
DaniPopes, line 55: maybe abi + dispatch should be done together? basically going from public functions to the concrete entry fn with the switch; i doubt there is much to gain in separating these?

crates/config/src/opts.rs
```diff
@@ -185,6 +196,14 @@ impl CompileOpts {
         !matches!(self.optimization, OptimizationMode::None)
     }
 
+    /// Returns whether codegen should favor bytecode size over runtime gas
+    /// (`-O size`): multi-use helpers are shared through calls rather than
+    /// inlined at every site.
+    #[inline]
+    pub const fn optimize_mir_for_size(&self) -> bool {
```
DaniPopes: please just do this check inline where needed, or as a method on OptimizationMode itself, not on CompileOpts. this is slop, same for fn optimize_mir above

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -671,16 +698,28 @@ impl EvmCodegen {
...
+        // tail-call edges as MIR. Each pass bails without advancing the phase
+        // when the module is outside its scope, in which case runtime
+        // generation falls back to the backend dispatcher.
+        if self.mir_dispatch {
+            crate::pass::run_pass(module, &crate::pass::LOWER_ABI_PASS);
+            crate::pass::run_pass(module, &crate::pass::LOWER_DISPATCH_PASS);
+            // `lower-evm-shaped` stays out of this pipeline for now: its
+            // argument-carrying tail calls need callee frame setup the backend
+            // does not perform yet, while the wrapper's internal call to a
+            // fused body is fully supported.
```
DaniPopes: so it's unimplemented?

DaniPopes (changes requested):
mostly makes sense
1. please try to split this pr up. obviously a few things are conflated: couple frontend fixes, couple codegen fixes, phases refactor, etc. I'm sure some of these can be standalone PRs with little effort, e.g. the Ty printer fix with a targeted test
2. i think we want to move the main passes to after all the lowering passes are done right? and only run higher level stuff like keccak+abi encode detection at the higher level. we can worry about passes later of course.

djolertrk:
I am splitting it to:

1. https://github.com/paradigmxyz/solar/pull/926
2. https://github.com/paradigmxyz/solar/pull/927
3. https://github.com/paradigmxyz/solar/pull/928

and then I will update this one. Thanks!

crates/codegen/src/backend/evm/codegen.rs
djolertrk: half way there, we can use it with `mir-opt`, but the backend consumption needs to be added (like Callee needs to do frame setup). will add that as a separate commit now actually in this PR. thanks!

crates/codegen/src/mir/module.rs
djolertrk: hmm you are right, noting they are back-to-back and should go in one shot, adding it

crates/codegen/src/backend/evm/codegen.rs
djolertrk: added

tests/ui/cli/Zhelp.stdout
djolertrk: well, the wrappers only decode arguments. they don't do returndata encoding yet, so any module with a returning external function (getters included) falls back to the backend dispatcher. the other case is nothing to wrap at all (only receive/fallback). reworded the help text to say exactly that.

tests/ui/codegen/revert_payloads.stdout
```diff
@@ -47,63 +47,24 @@ fn @require_message(arg0: bool) {
...
-    v12 = add 68, v6
-    revert 0, v12
+    mstore 36, arg0 !metadata(memory=scratch)
+    mstore 68, arg1 !metadata(memory=scratch)
+    revert 0, 100
+}
+
+fn @revert_message() {
+  bb0 (entry):
+    internal_call fn2, 0, 3, 0x6261640000000000000000000000000000000000000000000000000000000000
```
DaniPopes: should display the name of the function

tests/ui/codegen/mir/phase_pipeline_transition.mir
```diff
@@ -0,0 +1,10 @@
+//@compile-flags: --pipeline-default
+// The canonical pipeline is a phase transition: the module comes out in
+// `optimized` and the header says so.
```
DaniPopes, line 5: id like to see CHECKs for these like in --emit tests, I added support for filecheck recently

tests/ui/codegen/abi_packed_calldata_bytes.stdout
```diff
@@ -0,0 +1 @@
...168061012052156101205160208282820302905001806101005290508061010052602081018060e0528060e0528180610100528110905090506078575b60405160e0518060e0528180610140520160405260a0518060a05281526020810160206101005103818061016052015f81525060246004350160a0518060a05281838061016052375060405182806101405251925082806101805282806101605282806101a0525e90508180610180528101915060243582806101c05252602082019150806101a052810390506101a0512060805260206080f35b506102206040523660048103905060408112905015610188575b5f80fd5b6024355f1960601c818061012052168114905015610184575b5f608052600460043501358060a0528060a052601f81018060e0528060e052818060a052811090509050156101e5575b634e487b7160e01b5f52604160045260245ffd5b601f1960e0511680610140521561014051602082828203029050018060c05290508060c0526020810180610100528061010052818060c0528110905090506101d1575b6040516101005180610100528180610160520160405260a0518060a052815260208101602060c05103818061018052015f81525060246004350160a0518060a0528183806101805237506040516270726560e81b81806101a052526003810183806101605251935083806101c05283806101805282806101e0525e915082806101c05282019150915060243560601b8180610200525260148101905081806101a052810390508120905060805260206080f35b5f5ffd"}},"version":"VERSION"}
```
DaniPopes, line 1: these --emit bin tests aren't very useful, other than asserting that it succeeds? what can we do about it? print evm ir/opcodes?

tests/ui/codegen/abi_decode_dynamic_tuple.stdout
```diff
@@ -558,30 +558,36 @@ fn @decodeBytes(arg0: memptr) {
...
-    v52 = add v34, 32
-    v53 = add v48, v47
-    mcopy v48, v52, v44
-    v54 = sub v53, v40
-    returndata v40, v54
+    internal_call fn3, 0, v34
+    invalid
+}
+
+fn @__ret_bytes(arg0: memptr) {
```
DaniPopes, line 565: shouldn't these be inlined if they have only one caller? i would think that for 1 caller callees its always profitable but maybe im wrong

crates/config/src/lib.rs
```diff
@@ -161,6 +161,16 @@ str_enum! {
     }
 }
 
+impl OptimizationMode {
+    /// Returns whether codegen should favor bytecode size over runtime gas
+    /// (`-O size`): multi-use helpers are shared through calls rather than
+    /// inlined at every site.
```
DaniPopes:
```suggestion
    /// (`-O size`).
```

crates/codegen/src/transform/dce.rs
```diff
@@ -255,6 +256,11 @@ impl DeadCodeEliminator {
             Terminator::SelfDestruct { recipient } => {
                 used.insert(*recipient);
             }
+            Terminator::TailCall { args, .. } => {
```
DaniPopes: these terminator matches can be simplified to visit values like for instructions, there's a few of these "xxx_terminator" fns that could use this

crates/codegen/src/mir/function.rs
```diff
@@ -336,6 +336,9 @@ pub struct FunctionAttributes {
     pub is_fallback: bool,
     /// Whether this is a receive function.
     pub is_receive: bool,
+    /// Never inline this function (synthesized shared helpers whose whole
+    /// point is existing once per module).
+    pub no_inline: bool,
```
DaniPopes, line 342: what about `cold` also?

crates/codegen/src/lower/mod.rs
```diff
@@ -563,6 +581,94 @@ impl<'gcx> Lowerer<'gcx> {
...
+    ///
+    /// The helper takes the message length (1..=32) and its bytes left-aligned
+    /// in one word, and reverts with the standard `Error(string)` encoding:
+    /// selector, head offset, length, one padded data word — 100 bytes of
+    /// revert data, matching what the generic in-line path produces for short
+    /// messages. Sharing this cold path saves the string materialization and
+    /// ABI-encode boilerplate (~60-90 bytes) at every `require`/`revert` site
+    /// with a constant short message.
+    pub(super) fn ensure_revert_error_helper(&mut self) -> FunctionId {
+        if let Some(id) = self.revert_error_helper {
```
DaniPopes: these can use self.X.get_or_insert_with

crates/config/src/opts.rs
```diff
@@ -99,6 +99,17 @@ pub struct CompileOpts {
...
+    /// Library addresses for linking, as `LibraryName=0xADDRESS` (an optional
+    /// `path.sol:` prefix on the name is accepted and ignored). A `public`/
+    /// `external` library function whose library has a linked address is
+    /// called through `DELEGATECALL` at that address instead of being inlined
+    /// into the caller.
+    #[cfg_attr(
+        feature = "clap",
+        arg(long = "libraries", value_name = "NAME=ADDRESS", value_delimiter = ',')
+    )]
+    pub libraries: Vec<String>,
```
DaniPopes: let's parse it here like we do with import remappings

crates/config/src/opts.rs
```diff
@@ -99,6 +99,17 @@ pub struct CompileOpts {
     #[cfg_attr(feature = "clap", arg(short = 'O', long = "optimize", value_enum, default_value_t))]
     pub optimization: OptimizationMode,
 
+    /// Library addresses for linking, as `LibraryName=0xADDRESS` (an optional
```
DaniPopes: args should have a one liner description followed by an empty line, see the other fields. you can see that the entire description shows up in the short help, so it's not short

DaniPopes (changes requested):

crates/codegen/src/transform/dce.rs
djolertrk: yes!

tests/ui/codegen/abi_packed_calldata_bytes.stdout
djolertrk: hmmm, I agree. The better fix is an `--emit=asm` that disassembles the final program: mnemonic + offset + immediate per instruction. The backend already lowers to an `EvmAsmProgram` of `AsmInsts` and both the `EVM IR` and `asm` layers have `mnemonic()`, so it's mostly wiring an emit option to a printer, not new machinery, and it can be easily checked against `filecheck`

tests/ui/codegen/abi_decode_dynamic_tuple.stdout
djolertrk: You're right, and I relaxed the gate

crates/codegen/src/mir/function.rs
djolertrk: `cold` already handled at the block level, and I experimented with an attribute on function back then, but could not manage to produce any benefits so left it on the side for now

djolertrk:
@DaniPopes thanks!

djolertrk:
hey @DaniPopes, is this ready to be merged? it seems some weird things with CI env caused the CI to fail

DaniPopes (approved):

PR #926, opened 2026-07-07
fix(sema): encode library function signatures like solc

djolertrk:
A `library` (unlike a contract) may expose `public`/`external` functions that take `mapping` or `storage` reference parameters and refer to structs by canonical name, and solc encodes their signatures — and hence 4-byte selectors — specially:

  * structs are referred to by canonical name (`DataTypes.Reserve`), not flattened into an ABI tuple;
  * `mapping` types are printed structurally as `mapping(key => value)`;
  * `storage` reference parameters carry a trailing ` storage` suffix (`memory`/`calldata` do not).

`TyAbiPrinter` previously flattened structs (yielding wrong selectors for storage-struct parameters) and had no case for mappings or storage locations, so computing the selector of a mapping-parameter function hit the "printing unsupported type as ABI" assertion (seen on aave-v3-core GenericLogic, whose library functions take
`mapping(address => DataTypes.ReserveData) storage` parameters).

Add a `structs_by_name` mode to `TyAbiPrinter`, enabled for library functions in `item_signature`, that reproduces solc's encoding. Verified selector-for-selector against solc for mapping/struct/array/ nested storage parameters. The `--emit=hashes` UI test pins `total(S storage)` -> 33ad6f28 (previously flattened to `total((uint256,uint256))`, a wrong selector) and that contract signatures still tuple-encode. Mapping-parameter functions remain excluded from the interface by the `interfaceType` TODO in `interface_functions`, so they cannot appear in `hashes` yet; the printer already encodes them like solc when that lands.


tests/ui/abi/library_signature_encoding.sol
```diff
@@ -0,0 +1,39 @@
...
+//
+// Contract function signatures are unaffected: structs still flatten into ABI
+// tuples and carry no location suffix.
+//
+// `get` is dropped from the interface (and hence from `hashes`) because its
+// mapping parameter is not considered exportable yet (see the `interfaceType`
+// TODO in `interface_functions`); solc lists it as
+// `2aed1630: get(mapping(address => S) storage,address)`, which this printer
+// already produces.
+
```
DaniPopes, line 17: let's add filechecks and a // TODO CHECK for the todo

tests/ui/abi/library_signature_encoding.sol
```diff
@@ -0,0 +1,39 @@
...
+library L {
+    function get(mapping(address => S) storage m, address k) public view returns (uint256) {
+        return m[k].a;
+    }
+
+    // keccak256("total(S storage)")[..4] == 0x33ad6f28, matching solc.
+    function total(S storage s) external view returns (uint256) {
+        return s.a + s.b;
+    }
```
DaniPopes, line 99: can you add more where the struct or enum is defined inside of the library, and inside another contract

djolertrk:
@DaniPopes I agree, added

DaniPopes (approved):

DaniPopes (approved):

PR #929, opened 2026-07-07
feat: several code size improvements

djolertrk:
This PR reduces runtime bytecode size through MIR and EVM optimizations covering improvements to stack scheduling, spill reduction, memory/store elimination, some initial outlining, deduplication, dispatcher slimming, and static allocation of local heap objects. It also adds conservative correctness checks around the new transformations.

On the eight-contract corpus, the compiler moves from beating solc to 5/8, including ERC-20 at 1,799 vs 1,864 bytes.


crates/codegen/src/lower/mod.rs
```diff
@@ -118,6 +132,12 @@ pub struct Lowerer<'gcx> {
     storage_bytes_helper: Option<FunctionId>,
     /// Guards helper synthesis against routing through itself.
     synthesizing_helper: bool,
+    /// Per-function memo of computed mapping element slots, scoped like the
+    /// structured lowering so a memoized slot always dominates its reuse:
+    /// entries created inside a branch arm die with that arm's scope, loops
+    /// forget everything on entry, and memory writes/internal calls/assembly
+    /// invalidate all scopes (a key's bytes may have changed).
+    mapping_slot_memo: MappingSlotMemo,
```
DaniPopes: this is pretty hacky, especially the invalidation. could this be implemented as a builtin function and handled by existing optimization passes?

crates/codegen/src/transform/mod.rs
```diff
@@ -52,11 +54,56 @@ pub use lower_abi::{LowerAbiPass, LowerAbiStats};
...
+pub use static_alloc::StaticAllocPass;
 pub use storage_dse::{StorageDsePass, StorageStoreEliminator};
 pub use storage_load_cse::{StorageLoadCse, StorageLoadCsePass};
 pub use storage_promotion::{
     StoragePromotionStats, StorageScalarPromoter, StorageScalarPromotionPass,
 };
+
+/// Whether an external entry must reject nonzero callvalue, mirroring the
+/// backend dispatcher's rule.
+pub(crate) fn rejects_callvalue(func: &crate::mir::Function) -> bool {
```
DaniPopes: utils module?

crates/codegen/src/transform/outline_reverts.rs
DaniPopes, line 1: isn't this handled by block deduplication?

crates/codegen/src/transform/mod.rs
djolertrk: yep, it will work

crates/codegen/src/lower/mod.rs
djolertrk: agreed. I replaced the lowerer-local memoization with explicit MIR mapping-slot builtins. Existing CSE now handles reuse and memory invalidation, and a dedicated lowering pass expands them to the concrete memory and keccak256 operations afterward.

crates/codegen/src/transform/outline_reverts.rs
djolertrk: partially, but they operate at different levels. Block dedup only merges byte-identical terminal spans after lowering, while this pass canonicalizes equivalent revert shapes at MIR level (e.g. in LLVM, we have outliner at IR level as well as at Machine IR level). Also, removing it regressed the corpus code size overall

DaniPopes (approved):

PR #930, opened 2026-07-08
feat(lsp): add flychecks

0xKarl98:
Towards OSS-402

Adds on-save LSP flychecks for external Solidity diagnostics, with default forge lint --json discovery for Foundry workspaces and configurable custom commands from the VS Code extension. Flycheck diagnostics are parsed from solc/forge-style JSON, merged with compiler diagnostics by owner, cancelled when stale, timed out when hung, and republished only for affected files so unrelated diagnostics stay stable.

 Optimization points:

  - Auto-detects forge lint --json for Foundry workspaces.
  - Supports configurable flycheck commands via solarLsp.flychecks.
  - Parses solc JSON, forge lint JSON, and rustc-style diagnostic streams.
  - Converts byte offsets to LSP UTF-16 ranges with cached line indexes.
  - Cancels stale flychecks and kills timed-out child processes.
  - Tracks diagnostics by owner, so compiler and flycheck messages do not overwrite each other.
  - Publishes only affected URIs and clears stale/removed flycheck diagnostics.
  
  
### Original screenshots
截屏2026-07-08 19 09 56

### Modified a bit and save 
截屏2026-07-08 19 10 47

### multiple-flychecks 
截屏2026-07-08 19 17 07

Verified multiple configured flychecks can publish diagnostics for the same file simultaneously without one owner overwriting another. 

截屏2026-07-08 19 26 57

### DIFF
--- a/crates/lsp/src/flycheck/parser.rs
+++ b/crates/lsp/src/flycheck/parser.rs
@@ -1,7 +1,15 @@
 use crate::{diagnostics::DiagnosticMap, flycheck::config::FlycheckOutput};
-use lsp_types::{Diagnostic, DiagnosticSeverity, NumberOrString, Position, Range, Url};
-use serde_json::Value;
-use solar_interface::{data_structures::map::FxHashMap, source_map::SourceMap};
+use lsp_types::{
+    Diagnostic as LspDiagnostic, DiagnosticSeverity, NumberOrString, Position, Range, Url,
+};
+use serde::Deserialize;
+use solar_interface::{
+    data_structures::map::FxHashMap,
+    diagnostics::{
+        JsonDiagnostic, JsonDiagnosticMessage, JsonDiagnosticSpan, Severity, SolcDiagnostic,
+    },
+    source_map::SourceMap,
+};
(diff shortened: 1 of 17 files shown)


crates/lsp/src/flycheck/parser.rs
DaniPopes, line 1: this is parsing `forge lint --json` right? It uses solar's `JsonEmitter` so just use those data structures with serde?

0xKarl98: Have reused JsonEmitter's diagnostics and switched flycheck parsing to deserialize them with serde

mattsse (commented):
Two small things to tighten up:

- Scope flycheck cancellation/versioning to the affected owner/workspace. Right now saving an unrelated file, or one with no matching flycheck, can cancel another workspace's in-flight check without replacing it.
- Preserve Foundry lint severity: `forge lint --json` emits info/gas/code-size as `level: "note"`, so map `note`/`gas`/`code-size` to `INFORMATION` instead of warning.

DaniPopes (approved):

PR #932, opened 2026-07-09
fix(sema): allow events and errors named builtins

### LINKED ISSUE #216: [ICE]: identifier `this` already declared

Osraka:
Fixes #216.

Allows contract-level events and errors named `this` or `super` to coexist with Solar's synthetic builtin declarations.

The change keeps the scope narrow:

- declaration conflict checks ignore synthetic `this` / `super` only when the new declaration is an event or error
- other declarations, such as structs or enums named `this` / `super`, still report conflicts
- value-position resolution still prefers the builtin when the candidates are only `this` / `super` plus event/error declarations, so `address(this)` and `super.g()` continue to resolve as builtins

The regression test covers `event this`, overloaded `event this(uint256)`, `event super`, `error this`, `error super(uint256)`, and the corresponding `emit` / `revert` paths alongside normal builtin usage.

Tested:

- `cargo build -p solar-compiler --bin solar`
- `env TESTER_MODE=ui cargo test -p solar-compiler --test tests -- shadow_builtin_identifiers`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/sema/src/ast_lowering/resolve.rs
+++ b/crates/sema/src/ast_lowering/resolve.rs
@@ -2318,12 +2318,30 @@ impl Declarations {
         decl: Declaration,
         declarations: &[Declaration],
     ) -> Option<Declaration> {
+        use crate::builtins::Builtin as BuiltinKind;
         use Res::*;
         use hir::ItemId::*;
 
         if declarations.is_empty() {
             return None;
         }
+
+        let filtered;
+        let declarations = if matches!(decl.res, Item(Event(_) | Error(_))) {
+            filtered = declarations
+                .iter()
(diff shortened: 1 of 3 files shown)


DaniPopes:
Thanks but this is not worth it

PR #935, opened 2026-07-11
feat(ci): report compiler peak RSS

DaniPopes:
Add peak RSS summaries to the codegen benchmark report, including the average and maximum peak for every compiler and expandable per-benchmark Solar versus solc results. The report consumes `peak_rss_bytes` emitted by the benchmark runner and remains compatible with older result files that do not contain memory measurements.

This depends on walnuthq/solidity-compiler-benchmarks#2. Once that PR merges, this PR can pin `BENCHMARK_REF` to the upstream commit and exercise the report in CI before being marked ready.


djolertrk (approved):
LGTM, thanks!

mattsse (approved):
lgtm

PR #941, opened 2026-07-12
feat(lsp): add signature help

0xKarl98:
Closes OSS-497 

Adds capability-aware Solidity signature help to the LSP, backed by semantic analysis so the resolved overload appears first and signatures stay accurate across functions, constructors, events, errors, call options, and named arguments.

It also recovers incomplete calls while someone is typing without changing strict compiler parsing. Failed re-analysis can reuse still-valid call sites, stale callees are rejected, and responses respect client support for UTF-16 label offsets, Markdown NatSpec docs, and active parameters. The index shares rendered signatures and avoids redundant allocations to keep the request path lean.


crates/lsp/src/signature_help.rs
```diff
@@ -0,0 +1,1142 @@
...
+                .map(|call| call.range.start)
+                .collect::<FxHashSet<_>>();
+            for previous_call in previous_calls {
+                if !current_starts.insert(previous_call.range.start) {
+                    continue;
+                }
+                let mut call = previous_call.clone();
+                for signature in &mut call.signatures {
+                    *signature = self.intern_shared_signature(signature.clone());
+                }
```
mattsse: The callee-token check at request time only proves that the call text is unchanged; it does not prove that the declaration still exists. For example, rename `function foo(uint256)` to `bar` without changing an existing `foo(1)` call. The new analysis cannot resolve `foo`, this block restores the old call site, and signature help advertises the removed declaration. Filter the cached signatures through the newly built index before restoring the site (or otherwise retain and validate declaration identity).

```suggestion
                let mut call = previous_call.clone();
                call.signatures.retain(|signature| {
                    self.signatures_by_label
                        .get(&signature.information.label)
                        .is_some_and(|candidates| candidates.contains(signature))
                });
                if call.signatures.is_empty() {
                    continue;
                }
                for signature in &mut call.signatures {
                    *signature = self.intern_shared_signature(signature.clone());
                }
```

mattsse (changes requested):
The signature-help implementation is well covered, but cached call sites can outlive the declarations they were resolved against. I reproduced the issue by renaming `function foo` to `bar` while leaving an unresolved `foo(...)` call: failed re-analysis restores the old call site and still returns the removed `foo` signature. The inline suggestion prevents restoring signatures that are absent from the new semantic index.

crates/lsp/src/proto.rs
```diff
@@ -11,6 +11,37 @@ pub(crate) fn vfs_path(url: &lsp_types::Url) -> Option<vfs::VfsPath> {
     url.to_file_path().map(VfsPath::from).ok()
 }
 
+/// Converts an LSP UTF-16 position to a byte offset.
+///
+/// Out-of-range lines and columns are clamped as required by LSP. Positions that split a UTF-16
+/// surrogate pair return `None`.
+pub(crate) fn text_offset(rope: &Rope, position: lsp_types::Position) -> Option<usize> {
```
DaniPopes: why is this different from the range fn? like isnt there a method for this on rope

crates/lsp/src/signature_help.rs
0xKarl98: Have fixed stale signature helper

crates/parse/src/parser/mod.rs
```diff
@@ -666,37 +688,45 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
...
 
-        if !allow_empty {
-            v.push(f(self)?);
-            first = false;
-        }
-
-        while !self.check(ket) {
-            if self.token.kind == TokenKind::Eof {
+        loop {
+            let required_first = first && !allow_empty;
```
DaniPopes, line 685: lets extract the parser changes to a separate pr and add a test that enables this bool it thru a -Z flag with two revisions demonstrating what it does

0xKarl98: Extracted the parser recovery into #942. It adds `-Zrecover-incomplete-input` and a `strict`/`recover` revisioned UI test showing that recovery retains the incomplete call, continues semantic analysis, and recovers EOF-delimited items. I will rebase this PR to remove the parser diff after #942 lands.

crates/lsp/src/proto.rs
0xKarl98: Have removed it and keep range fn

DaniPopes:
looks good



image

DaniPopes (approved):

mattsse (approved):

PR #942, opened 2026-07-13
feat(parser): recover incomplete input

0xKarl98:
Should be merged before #941  

This extracts incomplete-input recovery from https://github.com/paradigmxyz/solar/pull/941 into a compiler-facing parser change. The new `-Zrecover-incomplete-input` mode keeps strict parsing as the default while allowing editor-style callers to retain a partial AST after incomplete argument lists, missing semicolons, and EOF-delimited item lists.

A revisioned UI test runs the same source in strict and recovery modes. The recovery revision demonstrates that the incomplete call remains available to semantic analysis and that parsing continues into later items.

### DIFF
--- a/crates/parse/src/parser/item.rs
+++ b/crates/parse/src/parser/item.rs
@@ -50,7 +50,12 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
         }
         if !self.eat(end) {
             let (msg, note) = get_msg_note(self);
-            return Err(self.dcx().err(msg).span(self.token.span).note(note));
+            let err = self.dcx().err(msg).span(self.token.span).note(note);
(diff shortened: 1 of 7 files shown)


crates/parse/src/parser/item.rs
```diff
@@ -50,7 +50,12 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
         }
         if !self.eat(end) {
             let (msg, note) = get_msg_note(self);
-            return Err(self.dcx().err(msg).span(self.token.span).note(note));
+            let err = self.dcx().err(msg).span(self.token.span).note(note);
```
DaniPopes: we can probably always emit here instead of returning

DaniPopes (approved):

PR #949, opened 2026-07-14
fix: address some correctness issues

djolertrk:
address: cold-path lowering defects, ternary operand preservation, same-block use-before-definition validation, and safe multi-return scratch handling. Adds targeted regression and runtime differential coverage.

Depends on https://github.com/paradigmxyz/solar/pull/929


djolertrk:
The regression will go away after we migrate to the improved memory model.

crates/codegen/src/lower/call.rs
```diff
@@ -1162,24 +1162,20 @@ impl<'gcx> Lowerer<'gcx> {
...
-        // For structs, lower_expr returns the memory pointer.
-        let arg_vals: Vec<ValueId> =
-            arg_infos.iter().map(|(arg, _)| self.lower_expr(builder, arg)).collect();
+        // Use the recursive ABI encoder for every high-level call. The former
+        // shallow struct loop copied nested memory pointers as calldata words
+        // and treated dynamic bytes pointers as their encoded value.
+        let arg_exprs: Vec<_> = args.exprs().collect();
+        let selector_word = builder.imm_u256(U256::from(selector) << 224);
+        let Some((calldata_start, calldata_size)) =
+            self.abi_encode_call_payload(builder, Some(selector_word), &arg_exprs)
```
mattsse, line 1171: [P1] Materialize calldata references before memory ABI encoding.

This recursive encoder expects dynamic values in the memory `[length][data...]` representation, but a calldata parameter still lowers to its ABI head offset. Minimal repro:

```solidity
interface Sink { function consume(bytes calldata data) external; }
contract C {
    function forward(bytes calldata data, Sink sink) external {
        sink.consume(data);
    }
}
```

The resulting MIR performs `mload arg0` and copies from `arg0 + 32`; `arg0` is relative to calldata, not a memory pointer, so the outbound payload is read from unrelated memory. Before calling the tuple encoder, detect `calldata_dyn_head` and either encode directly from calldata or materialize `bytes`/`string` and dynamic arrays into their normal memory representation. Reusing `materialize_calldata_bytes` / `materialize_calldata_dyn_array` (or the corresponding `lower_expr_as_memory_*` helpers) would keep the encoder's input contract consistent. A regression should check the forwarded bytes at runtime, not only that a `call` instruction is emitted.

crates/codegen/src/lower/expr.rs
```diff
@@ -1154,23 +1212,22 @@ impl<'gcx> Lowerer<'gcx> {
...
+                .iter()
+                .filter_map(|elem| elem.map(|elem| self.lower_expr(builder, elem)))
+                .collect::<Vec<_>>()
         } else {
-            // Not a tuple - just store single value at offset 0
-            let val = self.lower_expr(builder, expr);
-            let zero = builder.imm_u64(0);
-            builder.mstore(zero, val);
-        }
+            vec![self.lower_expr(builder, expr)]
```
mattsse: [P1] Preserve the tail of a non-literal tuple branch before restaging it.

Minimal repro:

```solidity
interface PairSource {
    function pair() external returns (uint256, uint256);
}
contract C {
    function choose(PairSource source, bool condition)
        external returns (uint256, uint256)
    {
        return condition ? source.pair() : (3, 4);
    }
}
```

`source.pair()` writes both return words at its return buffer, but this branch collects only the first MIR value. `stage_multi_return_values` then publishes a new base at the current free-memory pointer, abandoning (and in this case overwriting) the second word. The true branch returns `(first, 0)` instead of the callee's pair. For a non-literal tuple-producing expression, snapshot all values using the known tuple arity and the producer's `MULTI_RETURN_BUFFER_PTR_SLOT` before publishing a new buffer, or carry the producer's buffer through the merge without restaging only the first value. Please cover both ternary arms with a runtime differential test.

crates/codegen/src/lower/mod.rs
```diff
@@ -776,6 +780,7 @@ impl<'gcx> Lowerer<'gcx> {
 
         self.locals.clear();
         self.local_memory_slots.clear();
+        self.materialized_calldata_params.clear();
```
mattsse: [P1] Restore this per-function state after lowering a nested callee.

Minimal repro (with `helper` not yet lowered):

```solidity
function forward(bytes calldata data, uint256 start)
    external returns (uint256)
{
    data = data[start:];
    helper(1);
    return data.length;
}
function helper(uint256 x) internal pure returns (uint256) {
    if (x != 0) return x;
    return 0;
}
```

The slice rebind inserts `data` here, but `ensure_function_lowered` enters `lower_function(helper)`, which clears the set. Unlike `locals`, `local_memory_slots`, and `assigned_vars`, this set is not saved/restored. Back in `forward`, `data.length` is lowered as `calldataload(4 + <memory pointer>)`. Save/take/restore `materialized_calldata_params` in both `ensure_function_lowered` and `ensure_internal_mir_function` (and any other re-entrant function-lowering path) alongside the other per-function maps.

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -3188,6 +3200,10 @@ impl EvmCodegen {
             self.asm.emit_push_deferred(addr);
             return;
         }
+        if !self.in_internal_function && !self.in_constructor {
```
mattsse: [P1] Constructor bodies also need absolute addresses for cloned inline frames.

Minimal repro:

```solidity
contract C {
    uint256 public value;
    constructor(uint256 input, uint256 poison) {
        value = helper(input);
    }
    function helper(uint256 input) internal pure returns (uint256 result) {
        if (input == 0) result = 7;
        else result = input + 1;
    }
}
```

The MIR inliner clones the helper's `internal_frame_addr` operations into the constructor. While emitting the constructor, `in_constructor` is true and `in_internal_function` is false, so this condition falls through to `emit_current_internal_frame_addr`. No constructor frame pointer is initialized; moreover `0xa0` contains the second copied constructor argument, so cloned locals are addressed at `poison + offset`, causing wrong state or OOG. The absolute path should apply whenever `!in_internal_function`, including the constructor body. Internal callees emitted for the constructor still set `in_internal_function = true`, so they retain frame-pointer addressing. Please add a deployment test with a nonzero second argument and an inlined helper that uses a local/named return.

mattsse (changes requested):
Thanks for the fixes here. I found four reproducible correctness issues that still produce invalid runtime behavior. I included minimal repros and repair directions inline. The focused `solar-codegen` suite passes all 189 tests, so each case needs a regression test.

djolertrk:
> Thanks for the fixes here. I found four reproducible correctness issues that still produce invalid runtime behavior. I included minimal repros and repair directions inline. The focused `solar-codegen` suite passes all 189 tests, so each case needs a regression test.

@mattsse thanks!!!

crates/codegen/src/lower/abi_encode.rs
```diff
@@ -520,13 +520,20 @@ impl<'gcx> Lowerer<'gcx> {
...
-                .map(|(arg, ty)| (self.lower_return_value_for_ty(builder, arg, ty), ty))
-                .collect(),
-        )
+        let mut items = Vec::with_capacity(arg_exprs.len());
+        for (arg, ty) in arg_exprs.iter().zip(tys) {
+            let value = if let Some((head, is_bytes)) = self.calldata_dyn_head(arg) {
+                if is_bytes {
+                    self.materialize_calldata_bytes(builder, head)
+                } else {
+                    self.materialize_calldata_dyn_array(builder, head)
```
mattsse: [P1] Restrict this materializer to word-element arrays.

`calldata_dyn_head` classifies every dynamic array as non-bytes, but `materialize_calldata_dyn_array` only copies `length * 32` raw words. For `bytes[]` or arrays of structs, those words are ABI offsets or multiword heads rather than memory pointers/values, so forwarding `sink.consume(data)` produces malformed calldata. I reproduced this with a raw sink: a direct memory-array call passed, while forwarding the same `bytes[] calldata` failed only under Solar. Gate this helper on `abi_is_word_element`; recursively materialize composite elements or encode them directly from calldata, and add a runtime differential test.

crates/codegen/src/lower/expr.rs
```diff
@@ -434,6 +410,22 @@ impl<'gcx> Lowerer<'gcx> {
...
+                    let var = self.gcx.hir.variable(var_id);
+                    if var.data_location == Some(solar_ast::DataLocation::Calldata)
+                        && matches!(
+                            var.ty.kind,
+                            hir::TypeKind::Elementary(
+                                hir::ElementaryType::Bytes | hir::ElementaryType::String
+                            )
+                        )
+                    {
+                        self.materialized_calldata_params.insert(var_id);
```
mattsse: [P1] Do not record a branch-local representation globally.

This set is mutated while lowering one branch, so later uses treat the parameter as memory-backed on every path even though the replacement pointer exists only on the taken path. For `if (slice) data = data[1:]; return data.length;`, the taken path works but the untaken path returns the wrong result under Solar. Decide the representation before CFG lowering and initialize a memory-backed local on entry, or merge the representations explicitly; please cover both branch outcomes with a runtime differential test.

mattsse (changes requested):
Thanks for the follow-up fixes—the original four repros now pass. I found two remaining reproducible miscompilations in nested calldata forwarding and conditional slice rebinding; suggestions are inline. The focused `solar-codegen` suite passes all 190 tests, so both cases need runtime regression coverage.

DaniPopes (approved):

PR #950, opened 2026-07-14
feat(lsp): add semantic rename support

0xKarl98:
Towards OSS-502

-`Rename Symbol` on any `pick` should be unavailable because the call is conservatively ambiguous.
截屏2026-07-14 22 10 29

- rename either `BASE_SLOT` occurrence. Both the constant declaration and `layout at` expression should change
截屏2026-07-14 21 56 03

- rename `stored` to `storageSlot` ,  the declaration and `stored.slot` should change. 

截屏2026-07-14 21 58 25

- ename `SharedNS` to `LibraryNS`. Its declaration and qualified use should change exactly once, even though two entry files import the shared source.
截屏2026-07-14 21 59 31

- rename the `owner` label before `:` to `account`. Only the mapping key declaration and named-argument label should change
截屏2026-07-14 22 05 37

- Rename `amount` to `value`; the modifier parameter, its body reference, and the named-argument label in `guarded({amount: 1})` should update together.

截屏2026-07-15 10 02 20


crates/lsp/src/rename.rs
```diff
@@ -0,0 +1,769 @@
...
+        declarations: &IndexVec<SymbolId, DeclarationSymbol>,
+    ) -> Option<RenameCandidate> {
+        let occurrence = self.occurrence_at(uri, position)?;
+        let &target = occurrence.targets.first()?;
+        let targets = match target {
+            RenameTarget::Symbol(symbol_id) => self
+                .symbol_targets
+                .iter()
+                .copied()
+                .filter(|&candidate| same_symbol_declaration(declarations, symbol_id, candidate))
```
mattsse: `candidate` currently unifies only symbols that point to the same source declaration. Please also record and union the Solidity override family when building the rename index (including functions, modifiers, and public-variable getter overrides). Renaming `Base.f` currently edits only the base declaration, leaving `Child.f() override` and child call sites unchanged, so the resulting program no longer compiles. `gcx.base_override_functions` already exposes the override edges; please add regression coverage for renaming from both the base and derived declarations.

crates/lsp/src/symbols.rs
```diff
@@ -1281,6 +1424,63 @@ impl<'gcx> ReferenceCollector<'_, 'gcx> {
         }
     }
 
+    fn push_named_arg_references(&mut self, callee: &'gcx hir::Expr<'gcx>, args: &CallArgs<'gcx>) {
```
mattsse: This helper only accepts expression callees, so named labels on `hir::Modifier` nodes are never indexed. Those nodes cover both modifier invocations and base-constructor invocations. Please generalize the parameter-label indexing so `visit_modifier` can feed its resolved callable and `CallArgsKind::Named` here as well. Otherwise renaming `amount` changes its declaration but leaves `Base({amount: 1})` or `guarded({amount: 1})` unchanged, producing an invalid edit. Please cover both invocation forms with focused tests.

mattsse (changes requested):
Requesting changes for two valid Solidity constructs where semantic rename returns incomplete workspace edits that break compilation. Focused reproduction details and suggested fixes are inline.

crates/lsp/src/rename.rs
0xKarl98: Have fixed it , now whether initiated at the base declaration, derived declaration, or call site, the result is exactly the same.

crates/lsp/src/symbols.rs
0xKarl98: Have fixed

0xKarl98:
Added two missing cases :
- Rename  'Base' to 'Back' now correctly updates explicit base contract references in fallback and receive override specifiers.
截屏2026-07-15 15 55 45

- Renaming a function parameter 'amount' to 'value' now also updates its validated @param reference in NatSpec 
截屏2026-07-15 15 58 19

- Renaming a return parameter from 'result' to 'value' now updates the matching @return name in NatSpec too 
截屏2026-07-15 18 18 30

crates/lsp/src/rename.rs
```diff
@@ -0,0 +1,941 @@
...
+            return;
+        }
+        self.push_span_occurrence(gcx, final_ident.span, final_targets);
+        if identifiers.is_empty() {
+            return;
+        }
+
+        let mut current_symbols = symbols.to_vec();
+        for ident in identifiers.into_iter().rev() {
+            let mut parents = current_symbols
```
mattsse: `current_symbols` follows lexical declaration ownership, but a valid source path does not have to use that owner. For example:

```solidity
contract Base { struct S { uint256 field; } }
contract Child is Base {}
contract Use { Child.S value; }
```

Renaming `Child` currently edits only its declaration and leaves `Child.S` unchanged, so the resulting program no longer compiles. Please resolve/index the actual source qualifier components instead of deriving them solely from `DeclarationSymbol::parent` (for example, retain the resolver result for every identifier in the qualified path). Please add coverage that initiates rename from both the `Child` declaration and the `Child` qualifier.

crates/lsp/src/handlers/reqs.rs
```diff
@@ -64,6 +67,114 @@ pub(crate) fn references(
...
+    }
+
+    let mut changes = HashMap::<Url, Vec<TextEdit>>::new();
+    for location in candidate.locations {
+        changes
+            .entry(location.uri)
+            .or_default()
+            .push(TextEdit::new(location.range, new_name.clone()));
+    }
+    Ok(WorkspaceEdit { changes: Some(changes), document_changes: None, change_annotations: None })
```
mattsse: These edits are unversioned, so the content checks above do not close the race for open documents. `rename` runs asynchronously while notifications can update the VFS; a `didChange` arriving after a `Rope` is cloned/validated but before this response is applied leaves the client with stale ranges that `WorkspaceEdit::changes` cannot reject.

Suggested direction: retain the versions from `didOpen`/`didChange`, negotiate `workspace.workspaceEdit.documentChanges`, and return `TextDocumentEdit`s with `OptionalVersionedTextDocumentIdentifier` when supported (keeping `changes` only as the compatibility fallback). A regression should interleave a document change with an in-flight rename.

mattsse (changes requested):
Requesting changes for one reproducible incomplete semantic rename and one stale-edit race. Both have focused fix directions inline.

mattsse (approved):
lgtm

crates/sema/src/ty/mod.rs
```diff
@@ -1761,6 +1787,13 @@ fn is_value_ns(id: hir::ItemId) -> bool {
     )
 }
 
+impl Gcx<'_> {
+    /// Resolves a contract name within a source's scope for NatSpec `@inheritdoc`.
+    pub fn natspec_contract(self, name: Symbol, source: hir::SourceId) -> Option<hir::ContractId> {
```
DaniPopes: move to the main `impl Gcx` block

DaniPopes (approved):

PR #954, opened 2026-07-15
feat(sema): add lint query helpers

mattsse:
Adds a compact query layer for lint and analysis consumers: full-call target resolution, unambiguous variable extraction, function and builtin resolution projections, and an address type predicate.

Full-call resolution preserves attached `using for` metadata, function-typed variable targets, and the type checker's error resolution. Callable struct fields now carry their variable ID through the existing member resolution, while the existing resolved-member query retains the exact struct-field and enum-variant identity needed by codegen and language tooling.


crates/sema/src/ty/mod.rs
```diff
@@ -529,12 +529,31 @@ impl<'gcx> Gcx<'gcx> {
...
 
+    /// Returns the target selected for a full call expression.
+    ///
+    /// This preserves whether a member call is attached through `using for` and may resolve to a
+    /// function-typed variable rather than a function declaration. Returns `None` if overload
+    /// resolution failed.
+    #[inline]
+    pub fn resolved_call(self, expr: &hir::Expr<'gcx>) -> Option<ResolvedCallee> {
+        let hir::ExprKind::Call(callee, ..) = expr.peel_parens().kind else { return None };
+        self.resolved_callee(callee.id).filter(|resolved| !resolved.res.is_err())
```
DaniPopes: this shouldn't filter

crates/sema/src/ty/mod.rs
```diff
@@ -529,12 +529,31 @@ impl<'gcx> Gcx<'gcx> {
         self.typeck_results.get()?.type_of_expr(id)
     }
 
+    /// Returns the type inferred for the given expression, ignoring parentheses.
+    ///
+    /// Expression types are populated by the type checker.
+    #[inline]
+    pub fn expr_ty(self, expr: &hir::Expr<'gcx>) -> Option<Ty<'gcx>> {
```
DaniPopes: why is this necessary?

crates/sema/src/typeck/checker.rs
```diff
@@ -1541,7 +1541,16 @@ impl<'gcx> TypeChecker<'gcx> {
         let ty = match self.select_member_call_overload(receiver_ty, &possible_members, args) {
             Ok(member) => {
                 self.check_library_self_call(member, ident.span);
-                if let Some(res) = member.res {
+                let resolved = self.gcx.resolve_member_target(receiver_ty, member.name, member.res);
```
DaniPopes: this shouldn't be needed, you should add the res here https://github.com/paradigmxyz/solar/blob/0bd1fd8fd0ebeb8665e927263fda14445d799f64/crates/sema/src/builtins/members.rs#L192 and https://github.com/paradigmxyz/solar/blob/0bd1fd8fd0ebeb8665e927263fda14445d799f64/crates/sema/src/builtins/members.rs#L234

crates/sema/src/ty/mod.rs
mattsse: Removed the filter. `resolved_call` now preserves the type checker error resolution, with regression coverage.

crates/sema/src/ty/mod.rs
mattsse: Removed the redundant `expr_ty` wrapper. The test now uses the existing `type_of_expr(receiver.id)` query directly.

crates/sema/src/typeck/checker.rs
mattsse: Moved resolution to member construction. Members now carry the callable HIR resolution plus the exact `ResolvedMember`; struct fields retain their variable target and field identity, enum variants retain their variant identity, and the checker no longer reconstructs either from the receiver type and name.

crates/sema/src/builtins/members.rs
```diff
@@ -230,9 +258,19 @@ fn type_type<'gcx>(gcx: Gcx<'gcx>, ty: Ty<'gcx>) -> MemberListOwned<'gcx> {
...
-            gcx.hir.enumm(id).variants.iter().map(|v| Member::new(v.name, ty)).collect()
-        }
+        TyKind::Enum(id) => gcx
+            .hir
+            .enumm(id)
+            .variants
+            .iter()
+            .enumerate()
+            .map(|(variant_index, v)| {
+                Member::new(v.name, ty).with_resolved_member(ResolvedMember::EnumVariant {
```
DaniPopes: I was thinking Res is enough, since it can just be varid.

mattsse: Agreed for struct fields: I removed the parallel `Member.resolved` state and now store the field `VariableId` in `Member.res`. Enum variants cannot use the same representation yet because HIR stores them as `&[Ident]`, so there is no variant `VariableId`/`Res` to attach; they stay on the existing `resolve_member_target` path to preserve exact enum identity for codegen and LSP.

DaniPopes (approved):

PR #955, opened 2026-07-15
feat(sema): add call argument query

mattsse:
Adds `Gcx::call_arg`, a semantic query that returns the source argument bound to a visible parameter index. It handles positional and reordered named arguments, selected overloads, attached `using for` receivers, constructors, events, errors, variadic builtins, explicit conversions, function-typed variables, and function-typed struct fields.

Parameter-source discovery is centralized in sema and shared with the LSP, replacing its parallel constructor and function-field lookup. Special `abi.decode` parameter names now use the same binding path in type checking, lint queries, and inlay hints, while unresolved or genuinely ambiguous calls remain conservative.

### DIFF
--- a/crates/lsp/src/signature_help.rs
+++ b/crates/lsp/src/signature_help.rs
@@ -651,7 +651,7 @@ fn signature_declaration_parts<'gcx>(
             );
             doc_id = Some(variable.doc);
         }
-        None => {
+        Some(CallableParamSource::Builtin(_)) | None => {
             let fallback_name = fallback_name.or_else(|| match res {
                 Some(Res::Builtin(builtin)) => Some(builtin.name().to_string()),
                 _ => None,
--- a/crates/sema/src/ty/mod.rs
+++ b/crates/sema/src/ty/mod.rs
@@ -20,6 +20,7 @@ use solar_interface::{
     config::CompilerStage,
     diagnostics::{DiagCtxt, ErrorGuaranteed},
     source_map::{FileName, SourceFile},
+    sym,
 };
 use std::{
     fmt,
@@ -141,6 +142,8 @@ pub enum CallableParamSource {
     Event(hir::EventId),
     /// An error invocation.
     Error(hir::ErrorId),
+    /// A builtin with named parameters.
+    Builtin(Builtin),
 }
 
 /// The visible signature of a callable expression.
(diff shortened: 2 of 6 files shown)


crates/sema/src/ty/mod.rs
```diff
@@ -933,6 +1024,10 @@ impl<'gcx> Gcx<'gcx> {
             CallableParamSource::Struct(id) => self.param_names(self.hir.strukt(id).fields),
             CallableParamSource::Event(id) => self.param_names(self.hir.event(id).parameters),
             CallableParamSource::Error(id) => self.param_names(self.hir.error(id).parameters),
+            CallableParamSource::Builtin(Builtin::AbiDecode) => {
```
DaniPopes, line 1028: is this the only builtin with named params?

crates/lsp/src/signature_help.rs
```diff
@@ -506,7 +511,10 @@ fn render_res(gcx: Gcx<'_>, res: Res) -> Option<CallSignature> {
     if let Res::Item(item_id) = res {
         return render_item(gcx, item_id);
     }
-    let callable = gcx.callable_signature_of_ty(gcx.type_of_res(res))?;
+    let mut callable = gcx.callable_signature_of_ty(gcx.type_of_res(res))?;
+    if let Res::Builtin(builtin @ Builtin::AbiDecode) = res {
```
DaniPopes: these changes should not be done here

crates/sema/src/ty/mod.rs
mattsse: Yes — abi.decode is the only builtin we currently accept named arguments for (data and types). Variadic builtins reject named arguments, and the remaining fixed-arity builtins do not expose parameter names. This just moves the existing abi_decode_arg_kind mapping into the shared parameter-source query; I kept the Builtin payload so the source remains extensible if another named builtin is added.

crates/lsp/src/signature_help.rs
mattsse: Removed the `abi.decode` signature-help behavior and its regression fixture. The only remaining `signature_help.rs` diff is the exhaustive handling of the new `CallableParamSource::Builtin` variant through the existing fallback path. Focused sema call-argument tests, all 45 signature-help tests, nightly formatting, and sema/LSP clippy pass.

DaniPopes (approved):

PR #956, opened 2026-07-16
feat(lsp): add document formatting

0xKarl98:
Towards OSS-498 , 

This adds standard textDocument/formatting support backed by forge fmt, allowing editors to use the compiler’s LSP formatting provider instead of maintaining client-specific integrations.

Formatting uses the latest VFS contents so unsaved edits are preserved, while the real document path and nearest Foundry project root are retained for configuration discovery and [fmt].ignore handling. The server returns UTF-16-aware whole-document edits, rejects stale results when the document changes during formatting, and handles cancellation, timeouts, and Forge failures without mutating files directly.

- Regular Solidity file: Format the code currently open in the editor using the project’s Foundry settings 
截屏2026-07-16 13 08 24

- File ignored by foundry.toml: Leave the file untouched when its path matches [fmt].ignore. Will not be formatted 
截屏2026-07-16 13 09 18
  
   - With [fmt].ignore = ["src/*.sol"], Solar should format src/nested/Nested.sol because a single * must not match across directory separators , so this case should be solar formatted 
  
截屏2026-07-16 16 35 52

  
- File outside workspace will be formatted using nearest Foundry project configuration
截屏2026-07-16 13 15 26


crates/lsp/src/formatter.rs
```diff
@@ -0,0 +1,281 @@
...
+    let Ok(document) = load_foundry_document(&root.join("foundry.toml")) else {
+        return false;
+    };
+    let root = root.to_path_buf().normalize();
+    let path = path.to_path_buf().normalize();
+    let options = MatchOptions { require_literal_separator: true, ..MatchOptions::new() };
+
+    document.formatter_ignores().iter().any(|ignore| {
+        let ignore = ignore.trim_end_matches(['/', '\\']);
+        Pattern::new(&root.join(ignore).to_string_lossy()).is_ok_and(|pattern| {
```
mattsse: Normalize the joined ignore pattern too. Forge accepts patterns such as `./src/Test.sol` (and paths containing `..`), but the candidate path is normalized while this pattern is not, so the LSP formats a file Forge ignores.

```suggestion
        let ignore = root.join(ignore.trim_end_matches(['/', '\\'])).normalize();
        Pattern::new(&ignore.to_string_lossy()).is_ok_and(|pattern| {
```

mattsse (commented):
One remaining ignore-pattern mismatch:

crates/lsp/src/formatter.rs
0xKarl98: Have fixed !

editors/vscode/src/extension.ts
```diff
@@ -28,26 +28,22 @@ export function activate(context: vscode.ExtensionContext) {
         return;
       }
 
-      const edit = await formatDocument(editor.document);
-      if (edit) {
-        await editor.edit((builder) => {
-          builder.replace(edit.range, edit.newText);
-        });
-      }
+      await vscode.commands.executeCommand("editor.action.formatDocument");
```
mattsse: Please keep the direct `forge fmt` path as a fallback when the server does not advertise document formatting. The extension accepts arbitrary installed `solar` and `forge lsp` binaries; pre-956 binaries have no formatter, so this command and legacy format-on-save otherwise become no-ops.

crates/lsp/src/formatter.rs
```diff
@@ -0,0 +1,283 @@
...
+use tokio::{io::AsyncWriteExt, process::Command, time};
+
+const FORMATTER_TIMEOUT: Duration = Duration::from_secs(30);
+
+pub(crate) async fn run(forge: &Path, root: &Path, source: &str) -> Result<String, FormatterError> {
+    run_with_timeout(forge, root, source, FORMATTER_TIMEOUT).await
+}
+
+pub(crate) fn is_ignored(path: &Path, root: &Path) -> bool {
+    let Ok(document) = load_foundry_document(&root.join("foundry.toml")) else {
```
mattsse: Please use Forge's resolved formatter config (for example via `forge config --json`) instead of only the local top-level `[fmt]` table. This misses profile-scoped, inherited/global, `FOUNDRY_CONFIG`, and environment-provided ignores, causing files Forge considers ignored to be formatted.

crates/lsp/src/formatter.rs
```diff
@@ -0,0 +1,283 @@
...
+        return false;
+    };
+    let root = root.normalize();
+    let path = path.normalize();
```
mattsse: Please canonicalize existing candidate and ignore paths, with lexical normalization as the fallback for unsaved or nonexistent files. Forge canonicalizes both; opening a file through a symlink otherwise bypasses an ignore for the target path.

mattsse (changes requested):
A few compatibility gaps remain:

crates/lsp/src/formatter.rs
0xKarl98: Have applied these missing ignore cases

mattsse (approved):
LGTM

crates/lsp/src/handlers/reqs.rs
DaniPopes, line 1: tests in a different file please

DaniPopes (approved):

PR #958, opened 2026-07-16
perf(codegen): use dense sets for IDs

DaniPopes:
Replace hash sets for compact MIR, HIR, AST, source, contract, and graph-node indices with bitsets across analyses, lowering, backend state, optimization paths, semantic checking, and CLI orchestration. Stable domains use fixed dense sets, while parsing, rewriting, expression results, changing declaration scopes, and dynamically allocated graph nodes use growable sets. Composite and semantic keys continue using hash sets.

Bitsets now accept both generated index newtypes and plain `usize` keys, and dense, chunked, mixed, and growable bitsets support direct borrowed iteration. Ordered bitset traversal makes generated layouts deterministic; the affected codegen snapshots are refreshed. The audit also exposed and fixes a DCE bounds assumption around stale argument values after phase lowering.


crates/codegen/src/transform/dce.rs
```diff
@@ -158,15 +158,15 @@ impl DeadCodeEliminator {
...
+        let mut used_args = vec![false; func.params.len()];
 
         // Collect from all values used in instructions
         for (_, block) in func.blocks.iter_enumerated() {
             for &inst_id in &block.instructions {
                 let inst = &func.instructions[inst_id];
                 for val_id in inst.kind.operands() {
                     if let Value::Arg { index, .. } = &func.values[val_id] {
-                        used_args.insert(*index);
+                        used_args[*index as usize] = true;
```
mattsse: **[P1] Avoid indexing by a stale argument ID.** `func.params` can already have been shortened while MIR values still contain a `Value::Arg` with the old index. The existing `phase_pass_bounds.mir` test reaches this line with argument 0 and an empty `used_args`, so this becomes an out-of-bounds panic; the former hash set tolerated that intermediate shape. Please bounds-check this write (and the equivalent terminator path below) or retain set semantics.

mattsse (changes requested):
Requesting changes for two blocking issues:

- **[P1] DCE can panic on valid intermediate MIR.** The existing `phase_pass_bounds.mir` case reaches `find_unused_parameters` after a lowering pass removes a parameter while a `Value::Arg` still refers to its old index. The new `Vec<bool>` then indexes outside its domain; see the inline comment.
- **[P2] Codegen snapshots were not updated.** Dense-set iteration changes output order, and 16 existing UI tests now produce different bytecode. Please review and bless those outputs, or preserve the previous ordering if the changes are not intended.

On the earlier head, the exact `mir-opt` reproduction exited 101, while the codegen/data-structures library tests passed. The latest commit only adds iterator support and leaves both the DCE indexing and snapshot set unchanged.

crates/codegen/src/transform/dce.rs
DaniPopes: Fixed in 3c2297a2. DCE now tracks used arguments with DenseBitSet<usize> and ignores stale Value::Arg indices outside the current parameter domain in both instruction and terminator operands. The phase_pass_bounds.mir reproduction passes again.

DaniPopes:
Addressed both review blockers in 3c2297a2: DCE bounds-checks stale argument indices, and the 16 deterministic codegen snapshots have been reviewed and refreshed. All 255 codegen UI tests pass; the non-Foundry workspace suite is 498/498, and workspace clippy and formatting pass.

mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #959, opened 2026-07-16
feat(codegen): emit final EVM IR

DaniPopes:
Codegen UI tests currently snapshot opaque bytecode hex, which makes backend changes difficult to review. This adds plain-text creation and runtime EVM IR emit modes captured after assembler rewrites and immediately before byte emission, without running additional EVM IR passes. Creation output preserves constructor, deployment postlude, and runtime segments in byte order, while canonical opcode mnemonics keep the machine-level output readable. Existing bytecode snapshot fixtures now use these emit modes instead.

### DIFF
--- a/crates/cli/src/commands/evm_opt.rs
+++ b/crates/cli/src/commands/evm_opt.rs
@@ -80,28 +80,15 @@ fn process_evmir(args: &EvmOptArgs) -> Result<(), String> {
         .load_file(Path::new(&args.input))
         .map_err(|e| format!("failed to read {}: {e}", args.input))?;
     let text = source.src.as_str();
-    let mut result: Result<(), String> = Ok(());
     sess.enter(|| {
         let input_name = Ident::with_dummy_span(Symbol::intern(&args.input)).to_string();
         for (name, section) in evm_ir_sections(text, &input_name) {
-            let mut module = match parse_evm_ir_module(section) {
-                Ok(module) => module,
-                Err(err) => {
-                    result = Err(format!("{err}"));
-                    return;
-                }
-            };
-            if let Err(err) = verify_evm_ir_module(&module) {
-                result = Err(format!("{err}"));
-                return;
-            }
-            if let Err(err) = run_pipeline(&mut module, name, args) {
-                result = Err(err);
-                return;
-            }
+            let mut module = parse_evm_ir_module(section).map_err(|err| format!("{err}"))?;
+            verify_evm_ir_module(&module).map_err(|err| format!("{err}"))?;
+            run_pipeline(&mut module, name, args)?;
         }
-    });
-    result
+        Ok(())
+    })
 }
 
 fn evm_ir_sections<'a>(input: &'a str, input_name: &'a str) -> Vec<(&'a str, &'a str)> {
--- a/crates/cli/src/commands/evm_opt.rs
+++ b/crates/cli/src/commands/evm_opt.rs
@@ -82,21 +82,55 @@ fn process_evmir(args: &EvmOptArgs) -> Result<(), String> {
     let text = source.src.as_str();
     let mut result: Result<(), String> = Ok(());
     sess.enter(|| {
-        let mut module = match parse_evm_ir_module(text) {
-            Ok(m) => m,
-            Err(e) => {
-                result = Err(format!("{e}"));
+        let input_name = Ident::with_dummy_span(Symbol::intern(&args.input)).to_string();
+        for (name, section) in evm_ir_sections(text, &input_name) {
+            let mut module = match parse_evm_ir_module(section) {
+                Ok(module) => module,
+                Err(err) => {
+                    result = Err(format!("{err}"));
+                    return;
+                }
+            };
+            if let Err(err) = verify_evm_ir_module(&module) {
+                result = Err(format!("{err}"));
                 return;
             }
-        };
-        if let Err(e) = verify_evm_ir_module(&module) {
-            result = Err(format!("{e}"));
-            return;
+            if let Err(err) = run_pipeline(&mut module, name, args) {
+                result = Err(err);
+                return;
+            }
         }
-        let name = Ident::with_dummy_span(Symbol::intern(&args.input)).to_string();
-        result = run_pipeline(&mut module, &name, args);
     });
     result
+}
+
+fn evm_ir_sections<'a>(input: &'a str, input_name: &'a str) -> Vec<(&'a str, &'a str)> {
+    let mut sections = Vec::new();
+    let mut name = input_name;
+    let mut offset = 0;
+    let mut section_start = 0;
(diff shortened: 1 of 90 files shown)


crates/cli/src/emit.rs
```diff
@@ -205,20 +218,46 @@ fn generate_contract_bytecodes(gcx: Gcx<'_>) -> Result<FxHashMap<ContractId, Gen
...
         bytecodes.insert(
             id,
             GeneratedBytecodes {
-                deployment: alloy_primitives::hex::encode(artifact.deployment),
-                runtime: alloy_primitives::hex::encode(artifact.runtime),
+                deployment: artifact.deployment.into(),
+                runtime: artifact.runtime.into(),
+                deployment_evm_ir: capture_evm_ir
+                    .then(|| format_deployment_evm_ir(&artifact.deployment_evm_ir)),
+                runtime_evm_ir: artifact.runtime_evm_ir.map(|ir| ir.to_text().to_string()),
```
mattsse, line 231: The runtime emit does not currently round-trip through `evm-opt`. For example, emitting `tests/ui/codegen/emit_abi_bin.sol` with `--emit=evm-ir-runtime` and feeding `combined.evmir` to `solar evm-opt` fails with `operand-cleared instruction `mstore` must declare an explicit stack effect`. The captured module has explicit `stack=0->0` metadata, but `to_text()` suppresses it as a default and the parser reconstructs `None`. Could we preserve required metadata when printing operand-cleared operations and add a parse + verify round-trip test for this emit mode?

crates/cli/src/emit.rs
```diff
@@ -205,20 +218,46 @@ fn generate_contract_bytecodes(gcx: Gcx<'_>) -> Result<FxHashMap<ContractId, Gen
...
+pub(crate) fn format_deployment_evm_ir(modules: &[EvmIrModule]) -> String {
+    use std::fmt::Write;
+
+    let mut output = String::new();
+    for (index, module) in modules.iter().enumerate() {
+        if index != 0 {
+            output.push('\n');
+        }
+        writeln!(output, "// === {} ===", module.name).unwrap();
+        write!(output, "{}", module.to_text()).unwrap();
```
mattsse, line 256: Concatenating these independent modules produces creation output that is not valid as a single EVM IR input: every segment restarts at `bb0`, so `solar evm-opt combined.evmir` fails on the deployment segment with `duplicate block `bb0``. Constructor segments are also left without a terminator after their trailing `STOP` is removed. Could we either merge/relabel the segments into one valid module with the constructor fallthrough represented, or emit them as separate parseable units? An `evm-opt` round-trip test would pin this down.

tests/ui/codegen/emit_abi_bin.sol
```diff
@@ -1,5 +1,5 @@
 //@ignore-host: windows
-//@compile-flags: -Zcodegen --emit=abi,bin,bin-runtime --pretty-json
+//@compile-flags: -Zcodegen --emit=abi,evm-ir,evm-ir-runtime --pretty-json
```
mattsse: This was the last UI test exercising `--emit=bin`/`bin-runtime`, and the PR now also refactors combined JSON emission and byte storage. Could we keep a revision or a small dedicated fixture for `abi,bin,bin-runtime` so the JSON field names, hex serialization, and bytecode outputs remain covered while the other snapshots move to readable EVM IR?

mattsse (commented):
Suggestions from a second pass. Differential checks show the generated bytecode remains unchanged; the inline comments focus on emitted EVM IR validity and retaining coverage for the refactored bytecode/combined-JSON output.

crates/cli/src/emit.rs
DaniPopes: Fixed in ef0cb1bf. Operand-cleared operations now retain their explicit stack effect in text, and the emitted-output regression test parses and verifies every EVM IR section.

crates/cli/src/emit.rs
DaniPopes: Fixed in ef0cb1bf without merging modules. Constructor fallthrough is represented explicitly as `fallthrough_next`, and `evm-opt` processes the emitted section markers as separate parseable/verifiable modules.

tests/ui/codegen/emit_abi_bin.sol
DaniPopes: Fixed in ef0cb1bf. `emit_abi_bin.sol` now has a `bin` revision covering `abi,bin,bin-runtime`, including kebab-case field names and hex serialization.

crates/cli/src/commands/evm_opt.rs
```diff
@@ -80,23 +80,44 @@ fn process_evmir(args: &EvmOptArgs) -> Result<(), String> {
...
+
+fn evm_ir_sections<'a>(input: &'a str, input_name: &'a str) -> Vec<(&'a str, &'a str)> {
+    let mut sections = Vec::new();
+    let mut name = input_name;
+    let mut offset = 0;
+    let mut section_start = 0;
+    for line in input.split_inclusive('\n') {
+        let end = offset + line.len();
+        if let Some(next_name) =
+            line.trim().strip_prefix("// === ").and_then(|line| line.strip_suffix(" ==="))
```
mattsse: Non-blocking: this treats any `// === ... ===` comment as a section boundary, including an ordinary comment inside a valid single module. For example, placing such a comment between `bb0` (which jumps to `bb1`) and `bb1` makes `evm-opt` report `unknown block bb1`, while the base branch accepts the input. It may be worth first trying to parse the complete input as one module, and only applying section splitting when that fails.

crates/codegen/src/backend/evm/ir.rs
```diff
@@ -413,10 +415,18 @@ pub(super) fn default_instruction_stack_effect(inst: &EvmIrInstruction) -> EvmIr
...
-        EvmIrInstructionKind::Operation(_) => EvmIrStackEffect::new(
-            inst.operands.len().try_into().unwrap_or(u16::MAX),
-            u16::from(inst.result.is_some()),
-        ),
+        EvmIrInstructionKind::Operation(mnemonic) => {
+            if let Some(opcode) = super::assembler::op::from_mnemonic(mnemonic)
+                && let Some((inputs, outputs)) = super::assembler::op::stack_io(opcode)
+            {
+                EvmIrStackEffect::new(inputs, outputs)
+            } else {
```
mattsse, line 419: Non-blocking: for an unregistered mnemonic this fallback makes an operand-cleared unknown/pseudo operation with no result implicitly `0 -> 0`. Consequently `mystery` without stack metadata now passes verification, whereas the base branch required an explicit effect. Consider retaining the explicit stack-effect requirement for unknown operations while deriving effects only for registered opcodes.

mattsse (approved):
lgtm

PR #962, opened 2026-07-16
feat(codegen): add pass timing flag

DaniPopes:
Adds a global `-Ztime-passes` flag that reports elapsed time and change status for every MIR and EVM IR pass. The option is threaded through normal codegen, `mir-opt`, `evm-opt`, and the assembler bridge, while timing only the transform itself so validation and printing do not distort pass measurements.


mattsse (approved):
lgtm

PR #964, opened 2026-07-16
fix(bench): manage codegen PR comments

DaniPopes:
Post the codegen benchmark report only when a pull request changes benchmark output. Later changed runs create or update the sticky report normally, while unchanged runs use the action’s update-only mode: they do nothing when no report exists and add an unchanged-from-main note when one does. This also removes the separate REST API lookup whose transient HTML 503 responses caused benchmark jobs to fail, as seen on #961.


mattsse (approved):
lgtm

PR #967, opened 2026-07-17
refactor(codegen): require EVM IR lowering

DaniPopes:
Route all bytecode generation through MIR → EVM IR → assembly and remove the direct structured-assembly path. The assembler now constructs EVM IR with raw opcode instructions, symbols for names, and shared opcode metadata instead of string mnemonics or a separate structured assembly abstraction. Optimized builds run the canonical terminal-deduplication and block-layout passes by default, while unoptimized builds preserve their input layout. Runtime and deployment code reuse prepared EVM IR across offset stabilization, including a single constructor and deployment-postlude control-flow graph. Bytecode unit expectations use snapbox snapshots of a compact disassembly with one uppercase opcode and optional immediate per line.

Most of the large UI snapshot diff comes from correcting what `--emit=evm-ir` observes. It previously reconstructed EVM IR from linear assembly after assembler peepholes; it now captures the actual EVM IR immediately before lowering to assembly. The new snapshots therefore retain explicit CFG fallthrough edges and instructions that later assembler cleanup removes, and block layout renumbers blocks and moves cold revert paths. This is primarily representational churn rather than equivalent growth in generated bytecode: the bytecode snapshots change only where the default layout pipeline removes jumps or improves fallthrough, producing slightly smaller output.


crates/codegen/src/backend/evm/assembler/program.rs
```diff
@@ -354,74 +318,63 @@ impl StructuredAsmProgram {
...
-                let label = label_for_block(module, *target, labels, context);
+                let label = label_for_block(module, *target, labels, assembler);
                 program.push(AsmInst::push_label(label));
                 program.push(AsmInst::op(op::JUMP));
             }
             ir::TerminatorKind::RawOpcode(opcode) => program.push(AsmInst::op(*opcode)),
-            ir::TerminatorKind::Stop => program.push(AsmInst::op(op::STOP)),
+            // Falling off the end of bytecode is an implicit stop, so the
+            // bridge's synthetic final `Stop` does not need an opcode.
+            ir::TerminatorKind::Stop => {}
```
mattsse: When block layout moves this synthetic EOF block before another block, dropping the terminator changes the implicit stop into fallthrough. I reproduced this with an entry jump to an EOF tail followed by a cold revert block: the result executes the revert after the tail. Preserve an explicit `STOP` whenever the block is no longer last.

```suggestion
            ir::TerminatorKind::Stop => {
                if next_block(module, block_id).is_some() {
                    program.push(AsmInst::op(op::STOP));
                }
            }
```

mattsse (changes requested):
Found three merge-blocking regressions: block layout can turn an implicit EOF stop into fallthrough, `evm-opt` ICEs on valid identifiers longer than 31 bytes, and the verifier no longer rejects typed metadata in the untyped EVM IR. The EVM-IR UI suite is also currently failing because `stack_verify_missing_effect.stderr` is missing the final blank line.

djolertrk:
The fix for CI should be trivial, just update stack_verify_missing_effect.stderr with the final blank line.

Edit: I did not refresh and see the @mattsse already commented.

crates/codegen/src/backend/evm/assembler.rs
```diff
@@ -759,9 +736,7 @@ impl Assembler {
             }
             AsmInst::push(push_values.intern(value))
         });
-        if self.config.evm_ir_stack_schedule || self.config.evm_ir_layout_passes {
-            ir_program.optimize_with_evm_ir(self);
-        }
+        ir_program.lower_through_evm_ir(self);
```
djolertrk: There is still a constructor-specific issue here. The suggested fix for preserving STOP after layout fixes ordinary assembler semantics, but constructors later remove STOP only when it is the final byte (right?). If layout places a cold revert block after the constructor’s successful block, that STOP remains and execution never reaches the separately appended deployment postlude. You can reproduce this with a branchy constructor: deployment succeeds with zero bytes of runtime code. So, could constructor success jump to a shared deployment continuation, or could constructor and postlude be assembled as one CFG?

crates/codegen/src/backend/evm/assembler/program.rs
```diff
@@ -277,56 +254,43 @@ impl StructuredAsmProgram {
...
         labels: &mut Vec<Option<Label>>,
-        context: &mut C,
-    ) -> Self
-    where
-        C: StructuredAsmContext,
-    {
+        assembler: &mut Assembler,
+    ) -> Self {
         let mut program = Self::default();
         for (block_id, block) in module.blocks.iter_enumerated() {
```
djolertrk: Can we allocate labels for all referenced blocks before rebuilding the assembler program? Terminal `dedup` can make a later block jump to the initially unlabeled entry block. `label_for_block` then creates the entry label after that block has already been emitted, leaving a `PushLabel` with no matching definition.

You can reproduce `Assembler::assemble()` panicking with label `Label(1)` was never defined.

crates/codegen/src/backend/evm/ir/parse.rs
```diff
@@ -776,11 +794,15 @@ fn value_id(
...
         return value;
     }
-    let value = module.add_value(name.to_string());
+    let value = module.add_value(name);
     value_labels.insert(name.to_string(), value);
     value
 }
 
+fn block_label_index(label: &str) -> u32 {
+    label.strip_prefix("bb").and_then(|value| value.parse().ok()).expect("parsed block label")
```
djolertrk: This still ICEs on a lexically valid block label that does not fit in `u32`, for example `bb4294967296`. Could this return a `ParseError` instead of using `expect`?

djolertrk (changes requested):

mattsse (approved):
lgtm

PR #969, opened 2026-07-17
feat(lsp): add document links

0xKarl98:
Add support for `textDocument/documentLink` so editors can navigate directly from Solidity import paths to their resolved source files.

### DIFF
--- a/crates/lsp/src/document_links.rs
+++ b/crates/lsp/src/document_links.rs
@@ -4,13 +4,13 @@ use solar_interface::{
     data_structures::map::{FxHashMap, FxHashSet},
 };
 use solar_sema::{Gcx, ast};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::proto;
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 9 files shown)


mattsse (approved):
lgtm

crates/lsp/src/document_links.rs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            let source_path = source_path.to_path_buf();
+            let Some(ast) = &source.ast else { continue };
+            for &(item_id, target_source_id) in &source.imports {
+                let ast::ItemKind::Import(import) = &ast.items[item_id].kind else { continue };
+                let span = import.path.span;
+                if span.hi().to_u32().saturating_sub(span.lo().to_u32()) < 2 {
+                    continue;
+                }
+                let span = Span::new(span.lo() + 1, span.hi() - 1);
```
DaniPopes: why is this truncating? shouldn't it just be import.path.span ?

0xKarl98: Makes sense , have applied  !

DaniPopes (approved):

PR #972, opened 2026-07-17
fix(bench): clarify PR benchmark reports

DaniPopes:
Warn when benchmark results come from a branch that is behind main because those comparisons may be inaccurate. Behind or unchanged reports are collapsed while changed results from an up-to-date branch remain expanded, and gas deltas now use the same conventional lower-is-better sign as size deltas.


mattsse (approved):
lgtm

PR #975, opened 2026-07-17
fix(typeck): validate return conversions

DaniPopes:
Type-checks Solidity return expressions against their declared return types, including zero, single, tuple, and storage-pointer returns. It also preserves integer-literal semantics through negated compound expressions so invalid address returns are diagnosed consistently.


mattsse (approved):
lgtm

PR #976, opened 2026-07-17
fix(typeck): reject non-mobile functions

DaniPopes:
Models builtins as special function values and applies Solidity mobile-type rules in conditional expressions. Declaration, attached, builtin, low-level-call, and creation functions are rejected; allowed function values are normalized by removing declaration-specific identity and selector state while preserving ordinary internal, external, and library-call semantics.


mattsse (approved):
lgtm

PR #977, opened 2026-07-17
fix(typeck): tighten explicit conversions

DaniPopes:
Aligns explicit enum and calldata-slice conversions with tracked solc behavior. Enum values convert only to unsigned integers, literal enum casts are checked with bounded exact rational arithmetic and solc-compatible precision limits, and generic calldata array slices convert like their underlying arrays while preserving calldata view locations.


mattsse (approved):
lgtm

PR #980, opened 2026-07-18
feat: view/pure checker

DaniPopes:
This implements view and pure checking after type checking, covering state access, function calls, modifiers, events, inline assembly, and mutability restriction suggestions. Modifier and Yul function effects are inferred lazily through a shared concurrent function cache during the single parallel function traversal, while lexical Yul definition markers preserve solc-compatible checking and diagnostic order. Diagnostics are buffered for deterministic source-order emission.


mattsse (approved):
lgtm

PR #981, opened 2026-07-18
feat(lsp): add go to implementation

0xKarl98:
Towards OSS-499 , 

Editors can navigate from interface functions, virtual declarations, overrides, and call sites to every concrete implementation in the workspace. 

The lookup uses a shared override-family index, traverses multi-level and multiple inheritance directionally, keeps overloads separate, supports public getter and modifier overrides and named import aliases, and returns stable, deduplicated locations across overlapping analysis batches.

截屏2026-07-18 18 53 40


mattsse (approved):
lgtm

DaniPopes (approved):

PR #983, opened 2026-07-18
feat(lsp): add go to type definition

0xKarl98:
Towards OSS-500 

- Adds Go to Type Definition support for user-defined Solidity types, variables, function returns, getters, and cross-file
references. 
 
- It waits for the latest analysis so results stay fresh, and stores symbol IDs instead of full locations to keep the index
lightweight.

Works well under multi-root workspace 
截屏2026-07-18 21 05 43


mattsse (approved):
lgtm

DaniPopes (approved):

PR #986, opened 2026-07-19
feat(lsp): add document highlights

0xKarl98:
Add document highlights for Solidity symbols. Reads and writes are reported separately so editors can display them with different highlight colors. Declarations, assignments, compound assignments, deletes, and increment or decrement targets are classified as  writes, while ordinary value uses are classified as reads. 

 Document highlights showing separate read and write colors
  
 requests use the latest analysis so unsaved edits do not produce stale highlights.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #987, opened 2026-07-20
 feat(lsp): add hover support

0xKarl98:
Towards OSS-503 

- Hover configure at its call site. It should show the final override signature and inherited NatSpec remapped to the current parameter and return names.
image

- Hover configure at both its declaration and reference. Both locations should produce identical hover content. 
image
image

- Hover newLimit and newOwner. Each should show its type and inherited documentation using the current name rather than the base declaration’s name.
image
image

- Hover previousLimit and previousOwner. Each should show its type and the correctly mapped inherited return documentation.
image
image

- Hover the public struct variable record. It should show the variable declaration and inherited return documentation using the current field names value and owner. 
image

- Hover balances at its declaration and reference. Both should show the mapping declaration, inherited notice, and return documentation. 
image
image

- Hover mixedReturns. The unnamed and named return documentation should appear in declaration order. 
image

- Hover the attached library call bump. It should show the complete underlying library function signature and documentation,including self, increment, and result.   
image


tests/ui/natspec/tag_returns.sol
```diff
@@ -27,3 +27,11 @@ contract InvalidReturnNames {
     //~^ ERROR: tag `@return` does not contain the name of its return parameter
     function missingName() public returns (uint result) {}
 }
+
+contract InvalidReturnOrder {
+    /// @return second The second return value is documented first
+    //~^ ERROR: tag `@return` does not contain the name of its return parameter
```
DaniPopes: can we make the error message a bit better here?

crates/sema/src/ty/mod.rs
```diff
@@ -1518,9 +1520,32 @@ pub fn base_override_items(gcx: _, item: hir::ItemId) -> &'gcx [hir::ItemId] {
...
+/// compiler-generated getter. Other variables have no positional documentation.
+pub fn natspec_view(gcx: _, item: hir::ItemId) -> NatSpecView<'gcx> {
+    crate::natspec::resolve_view(gcx, item)
+}
+
 /// Returns the resolved NatSpec doc comments for the given doc ID.
+///
+/// This preserves the flat representation used by compiler outputs. Use [`Gcx::natspec_view`] when
+/// documentation must be associated with the current item's parameter or return positions.
 pub fn natspec_doc_comments(gcx: _, id: hir::DocId) -> &'gcx [hir::NatSpecItem] {
```
DaniPopes: now this should move into a normal impl block, not cached

0xKarl98:
Implemented in commit 7816e183: the new position index reduces burst-hover/256 latency from 1.607 ms to 65.82 µs, a 95.91% reduction and approximately a 24.4× speedup. 

The benchmark shows no measurable regression in analysis-build performance.

mattsse (approved):
lgtm

DaniPopes:
let's add the benches to codspeed CI and make a framework aroudn it, we probably want to see what other LSPs do and add some project-wide test cases

0xKarl98:
@DaniPopes  Have added the benches  to CI  , but think we can make the reusable framework to follow-up pr ?

DaniPopes (approved):

PR #992, opened 2026-07-20
chore(deps): update analyze dependencies

DaniPopes:
Raise the benchmark analysis requirements to the latest stable matplotlib, NumPy, and tabulate releases, and refresh the uv lockfile so the direct and transitive Python dependencies are current.


mattsse (approved):
lgtm

PR #993, opened 2026-07-21
feat(lsp): add project benchmark framework

0xKarl98:
This PR follows up on Dani's comment in #987:
> let's add the benches to codspeed CI and make a framework around it, we probably want to see what other LSPs do and add some project-wide test cases

It extends the existing CodSpeed coverage with a project-wide LSP benchmark framework and realistic `unifap-v2` workloads for full analysis, analysis after a source edit, and common LSP requests. Fixture preparation is kept outside the measured path to provide a stable baseline for future performance tracking and comparisons with other LSP implementations.

### Run it  :
cargo bench -p solar-lsp --bench lsp --features bench

### Open it :
open target/criterion/report/index.html 

### Result :
image

Benchmark result also be shown in codSpeed

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -14,7 +14,7 @@ const UNIFAP_FACTORY: &str = "src/UnifapV2Factory.sol";
 
 struct BenchmarkSource {
     project: BenchmarkProject,
-    hover_requests: Vec<BenchmarkRequest>,
+    hover_positions: Vec<(u32, u32)>,
 }
 
 struct SourceBuilder {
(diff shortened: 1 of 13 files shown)


mattsse (approved):
lgtm

DaniPopes:
let's make all the benches have the same scheme in naming, you can see there's currently 3 benches with just the project name in the codspeed comment

0xKarl98:
@DaniPopes  Have clarified it

crates/lsp/benches/lsp.rs
```diff
@@ -121,5 +132,188 @@ fn burst_hover(c: &mut Criterion) {
...
+        group.bench_with_input(BenchmarkId::from_parameter(name), request, |b, request| {
+            b.iter(|| black_box(analysis.execute(black_box(request))));
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(
+    benches,
+    analysis_build,
```
DaniPopes: can we merge these into one function so we can share the setup, would also help with understanding what benches are being run at a first glance, like in the normal solar benchmarks

0xKarl98: Have unified

DaniPopes:
this is still an issue

image

0xKarl98:
@DaniPopes Have fixed project-analysis pic

DaniPopes (approved):

PR #997, opened 2026-07-21
feat(lsp): add NatSpec completions

0xKarl98:
This adds declaration-aware NatSpec completion for empty line and block comments across contracts, functions, events, errors, structs, enums, and state variables. Templates include matching `@param` and `@return` tags, public getter returns, and signature-compatible `@inheritdoc` candidates, with client-negotiated snippets and UTF-16-safe edits.

Completion stays non-blocking while users type. It derives syntax-only templates from current VFS text and uses indexed semantic details only when source and workspace context are fresh, preventing edits, cache lifecycle changes, or failed and cancelled analysis from exposing stale getter or inheritance data.


crates/lsp/src/symbols.rs
```diff
@@ -288,6 +290,7 @@ impl SymbolTables {
         tables.build_references(gcx, &item_symbols);
         tables.document_links = DocumentLinkIndex::build(gcx, document_link_sources);
         tables.inlay_hints = InlayHintIndex::build(gcx);
+        tables.natspec_completion = NatSpecCompletionIndex::build(gcx);
```
mattsse, line 293: Building the NatSpec index here adds a full HIR/AST/source-token pass to every LSP analysis. CodSpeed reports about an 11% regression in both `analysis-build` cases, and same-machine base/head runs also showed a 4–6% slowdown. Could we collect these entries and source fingerprints during the existing symbol-table traversal, or build the index lazily, instead of rescanning the workspace on each edit?

0xKarl98: Sure , working on the optimizations atm

mattsse (approved):
lgtm

0xKarl98:
@mattsse  Just fixed a dead-lock , now they work as normal
截屏2026-07-22 13 03 53

DaniPopes (approved):

PR #1003, opened 2026-07-21
feat(codegen): support narrow immutables

DaniPopes:
Immutable references previously carried byte offsets through MIR and were forced into PUSH32 placeholders. This change gives each immutable a stable `ImmutableId`, records its `MirType` at the module boundary, prints named declarations under `immutables:`, and represents constructor assignments with a semantic MIR `StoreImmutable` instruction. Staging addresses and word storage now belong exclusively to the EVM backend; `storeimmutable` is not modeled as a Yul or EVM builtin.

Runtime lowering carries each immutable's `TypeSize` through untyped EVM IR and assembly relocation metadata. Deployment patches PUSH1 through PUSH32 immediates without overwriting neighboring runtime bytes. Signed integers are sign-extended after narrow pushes, fixed bytes are restored to their left-aligned MIR representation, and constructor staging, spills, heap allocations, and copied runtime code use disjoint backend-owned memory regions. Byte-counted ABI packing remains unchanged.

Immutable reads and writes have explicit effects so CSE, GVN, LICM, inlining, validation, and stack rematerialization preserve constructor read-before-write and read-after-write semantics. Coverage includes MIR parsing and round trips, optimizer clobber behavior, named lowering snapshots, and deployment/runtime execution for unsigned, signed, fixed-bytes, address, bool, default-width, UDVT, internal-function-pointer, and reassigned immutables.


DaniPopes:
waiting on cyclops

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -924,21 +952,87 @@ impl<'gcx> EvmCodegen<'gcx> {
...
+        if byte_width < 32 {
+            let trailing_bits = usize::from(32 - byte_width) * 8;
+            match encoding {
+                ImmutableEncoding::LeftAligned(_) => {
+                    self.asm.emit_push(U256::MAX << trailing_bits);
+                    self.asm.emit_op(op::AND);
+                }
+                ImmutableEncoding::Unsigned(_) | ImmutableEncoding::Signed(_) => {
+                    self.asm.emit_push(U256::from(trailing_bits));
+                    self.asm.emit_op(op::SHL);
```
DaniPopes, line 1063: Fixed in 39247c78. The error was that backend-synthesized deployment patching and fixed-bytes normalization emitted SHL without checking the selected fork. When has_bitwise_shifting is false, narrow immutable references now use a full-width TypeSize and PUSH32 placeholder. Deployment copies the complete word and runtime loads need no shift or sign/left-alignment repair. The Byzantium EVM IR check pins 32-byte references, and the run-call suite executes unsigned, signed, fixed-bytes, address, bool, UDVT, and function-pointer immutables on Byzantium.

crates/codegen/src/immutable.rs
```diff
@@ -0,0 +1,21 @@
+//! Deployment-time immutable staging layout.
+//!
+//! Immutable assignments are lowered to ordinary MIR memory stores, while the
+//! deployment postlude reads the same words to patch runtime `PUSH<N>`
+//! placeholders. This module keeps both sides on one layout.
+
+use crate::mir::ImmutableId;
+
+/// First constructor-memory word reserved for immutable values.
+pub(crate) const IMMUTABLE_STAGING_BASE: u64 = 0x2000;
```
DaniPopes: Fixed in 62aac3a4, with the exact argument-boundary regression in 004fa759. The failure was a real alias: fixed staging at 0x2000 overlapped constructor argument word 252 and sufficiently large local frames. Staging now begins after the actual constructor local region and reserved frame-pointer word, immutable words and exact spills follow contiguously, and the complete ABI blob is copied at a deferred ConstructorArgsBase after that fixed region. The free-memory pointer is advanced past the aligned blob. The old 0x2000 staging, 0x4000 heap floor, and fixed spill base are gone. Coverage includes a 0x3000 local frame and a run-call where an immutable initializer precedes reading argument word 252.

crates/codegen/src/mir/display.rs
```diff
@@ -400,14 +406,33 @@ fn display_inst_kind<'a>(
...
+    fmt::from_fn(move |f| {
+        if let Some(module) = module
+            && let Some(immutable) = module.get_immutable(id)
+        {
+            if module.iter_immutables().filter(|(_, other)| other.name == immutable.name).count()
+                == 1
+            {
+                write!(f, "{}", immutable.name)
+            } else {
+                write!(f, "{}{}", immutable.name, id.index())
```
DaniPopes: Fixed in 3bf50c0a. The error was that per-name numeric suffixing did not reserve source identifiers, so a generated alias such as value0 could collide with a real immutable named value0. Display names are now allocated module-wide: every source name is reserved first, duplicate aliases are generated deterministically, and underscores are appended until the alias is unused. immutable_name_conflicts.sol covers duplicate value declarations together with a real value0 and checks every declaration and reference.

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -2833,17 +2936,11 @@ impl<'gcx> EvmCodegen<'gcx> {
...
-                    // The running constructor's own placeholders are never
-                    // patched; read the staged scratch word instead.
-                    self.asm.emit_push(U256::from(
-                        EvmMemoryLayout::IMMUTABLE_SCRATCH_BASE + u64::from(*offset),
-                    ));
-                    self.asm.emit_op(op::MLOAD);
-                } else {
-                    self.asm.emit_push_immutable(*offset);
-                }
+            InstKind::StoreImmutable(..) => {
```
DaniPopes, line 3272: Fixed in ed67e760. The error was that a custom MIR pipeline could skip lower-immutables and hand StoreImmutable to the backend path that treated it as unreachable. lower-evm-shaped now refuses to advance while an immutable assignment survives, phase validation reports the invariant violation, and backend unsupported-instruction checking provides a recoverable source diagnostic as a final guard. The custom-pipeline UI regression checks the diagnostic and confirms there is no panic.

crates/codegen/src/mir/inst.rs
```diff
@@ -691,12 +701,15 @@ pub(crate) enum InstKind {
     ExtCodeCopy(ValueId, ValueId, ValueId, ValueId),
     /// Get external code hash: `extcodehash(addr)`
     ExtCodeHash(ValueId),
-    /// Read an immutable word identified by its byte offset: `loadimmutable <offset>`
+    /// Assign an immutable during construction: `storeimmutable <name>, value`.
+    /// Lowered to constructor staging memory after MIR optimization.
+    StoreImmutable(ImmutableId, ValueId),
```
DaniPopes, line 708: Fixed in 62522d3e. The error was that EffectKind described immutable effects but generic ModRef queries still returned no reads or writes, leaving current and future ModRef consumers unsound. Alias analysis now has an Immutable address space and exact Location::Immutable(ImmutableId) entries for loads and stores. Internal-call summaries propagate the fourth address space, and CSE derives Specific versus All immutable clobbers from ModRef instead of a separate manual path. A unit test pins the exact read/write locations.

DaniPopes:
cyclops audit

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -999,31 +1115,42 @@ impl<'gcx> EvmCodegen<'gcx> {
...
             self.constructor_param_count = ctor.params.len() as u32;
 
-            // If constructor has parameters, copy the full ABI-encoded argument blob to memory.
-            // Constructor args are appended after generated deployment bytecode, so the copy size
-            // is `CODESIZE - constructor_arg_offset`.
+            // Constructor args are appended after generated deployment bytecode.
+            // Copy the complete blob above every fixed compiler-owned region,
+            // then place the free-memory pointer after its word-aligned end.
             if let Some(arg_offset) = constructor_arg_offset {
+                self.constructor_args_base_const = Some(constructor_fixed_memory_end);
```
DaniPopes, line 1187: Fixed in ce490089. The bug had two parts: constructor lowering never rejected an ABI blob shorter than its semantic head, and memory-backed ABI decoding had no source bound. The deployment prefix now publishes the exact copied-blob end, and the constructor MIR prologue checks `end - base >= head_size` before any argument decode, retains that exact end, and then aligns the free-memory pointer. `AbiSource::Memory { end }` carries the bound through relative-offset resolution, word loads, byte copies, padded bytes tails, and nested-array heads; each range also rejects arithmetic overflow. The MIR/FileCheck fixture pins the head and word bounds, and run-call coverage exercises valid dynamic `bytes` and array constructor inputs.

crates/codegen/src/lower/mod.rs
```diff
@@ -576,16 +559,19 @@ impl<'gcx> Lowerer<'gcx> {
         {
             let mut builder = FunctionBuilder::new(&mut mir_func);
             let saved_lowering_constructor = self.lowering_constructor;
+            let saved_constructor_args_base = self.constructor_args_base;
             let saved_lowering_internal_function = self.lowering_internal_function;
             let saved_in_unchecked_block = self.in_unchecked_block;
             let saved_current_return_tys = std::mem::take(&mut self.current_return_tys);
             self.lowering_constructor = true;
+            self.constructor_args_base = None;
```
DaniPopes, line 574: Fixed in ce490089. Synthetic constructor generation was reusing ambient lowering state and never converted its local-memory high-water mark into `Function::internal_frame_size`, so the immutable layout legitimately saw a zero-sized frame. It now gets an isolated function-local lowering context, starts `next_local_memory_offset` at `HEAP_START`, records the resulting frame size after inlined base-constructor lowering, and restores the outer context afterward. Since immutable staging is derived from that recorded frame boundary, it can no longer overlap those locals. The regression test uses three mutable base-constructor locals so the old third slot lands exactly on the first immutable staging word, then verifies the derived immutable remains intact.

DaniPopes:
cyclops audit

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -1007,31 +1140,42 @@ impl<'gcx> EvmCodegen<'gcx> {
...
                 self.asm.emit_push_deferred(arg_offset);
                 self.asm.emit_op(op::CODESIZE);
                 self.asm.emit_op(op::SUB); // size = CODESIZE - arg_offset
+                self.asm.emit_op(op::dup(1));
```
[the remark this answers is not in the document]
DaniPopes: Confirmed as real generic constructor-ABI behavior, but it is not introduced by this PR. `origin/main` already computes `CODESIZE - constructor_arg_offset` and `CODECOPY`s the result without minimum-head or dynamic-tail bounds; this diff only moves the destination from the fixed heap start to the deferred backend-owned constructor argument base so it cannot overlap immutable staging or spills. We previously separated and reverted constructor ABI hardening from this immutable-sizing PR, so the head/tail validation belongs in that separate change.

PR #1004, opened 2026-07-22
feat: introduce new memory model

djolertrk:
- model slices and memory allocation in MIR
- fix codegen for Solady and enable it in benches
- fix a few more minor codegen issues


crates/codegen/src/analysis/memory_summary.rs
```diff
@@ -0,0 +1,311 @@
...
+use crate::mir::{Function, FunctionId, InstKind, Module, Terminator, Value, ValueId};
+use solar_data_structures::{index::IndexVec, map::FxHashSet};
+
+/// Conservative memory effects and pointer captures for one MIR function.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct FunctionMemorySummary {
+    reads: [bool; 3],
+    writes: [bool; 3],
+    may_reset_fmp: bool,
+    captures: Vec<bool>,
```
DaniPopes: use bitsets

tests/ui/erc7201/codegen.stdout
```diff
@@ -17,25 +17,23 @@ fn @zeroLastByteLiteral() {
...
-    v9 = and v8, 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00
-    mstore 128, v9
+    v0 = alloc memorybytes, exact, uninitialized, infallible, 64
+    set_memory_object_len memorybytes, v0, 12
+    v1 = memory_object_data memorybytes, v0
+    v2 = add v1, 0
+    mstore v2, 0x6578616d706c652e6d61696e0000000000000000000000000000000000000000
+    v3 = memory_object_len memorybytes, v0
+    v4 = memory_object_data memorybytes, v0
+    v5 = keccak256 v4, v3
```
DaniPopes: maybe pass slice directly to keccak and other instructions that expect (ptr,len) ?

DaniPopes (commented):
plz look at codspeed for performance, check all `Vec<bool>`

djolertrk:
> plz look at codspeed for performance, check all `Vec<bool>`

Yeah, I am on it

tests/ui/erc7201/codegen.stdout
djolertrk: yes, good idea!

DaniPopes: hmm actually we would still need the normal keccak instruction anyway, is this worth it?

djolertrk: yeah but we keep the physical keccak regardless. imo it is still consistent with how the rest of the memory model works, and the CSE/alias get one whole object read instead of the dance. so bytecode is identical either way, it is just ir. but if you'd rather not grow InstKind for this, I can drop it

djolertrk:
> > plz look at codspeed for performance, check all `Vec<bool>`
> 
> Yeah, I am on it

I am still on this

DaniPopes (approved):

tests/ui/erc7201/codegen.stdout
DaniPopes: it's fine then

PR #1005, opened 2026-07-22
perf(codegen): improve stack scheduling

DaniPopes:
MIR names SSA values without fixing their physical placement, while EVM instructions can directly reach only the top 16 stack words. Preparing operands independently duplicates consumable last uses, extends live ranges, and spills values that could remain on the stack. This PR adds target-aware operand and edge scheduling at the MIR-to-EVM lowering boundary. Late MIR instruction ordering is split into the stacked follow-up #1045.

The implementation prepares an instruction's complete operand head as one plan, consumes dead resident values in place, preserves live copies, rematerializes cheap values, and spills only when required. Selected loop and decoded-argument edges can carry a verified canonical stack layout; other edges retain the conservative spill-and-reload path. Local search remains disabled under `-O none`.

The [stack module documentation](https://github.com/paradigmxyz/solar/blob/dani/stack-scheduler/crates/codegen/src/backend/evm/stack/mod.rs) is the architectural reference for planner tiers, cost ordering, edge policies, spills, correctness boundaries, tradeoffs, and the pinned Plank, solc, Sonatina, and Fe sources evaluated or adapted. This is an original implementation rather than vendored allocator code.


mattsse (approved):
lgtm

DaniPopes:
waiting on cyclops setup

djolertrk (approved):
lgtm, nice!

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1006, opened 2026-07-22
refactor: use typed index vectors

DaniPopes:
Several compiler analyses and EVM transforms stored dense tables in `Vec` while indexing them with MIR or EVM index newtypes. This switches those tables to `IndexVec`, including local tables and helper signatures, so indexing remains type-checked end to end.

The repository guidance now requires `IndexVec` whenever a collection is keyed by a typed index, including local variables.


mattsse (approved):
lgtm

PR #1010, opened 2026-07-23
feat(lsp): add pull diagnostics

0xKarl98:
Towards OSS-511 

 - The diagnostic store now caches only non-empty merged reports and uses one stable result ID for empty reports, avoiding per-URI cache entries for documents without diagnostics. 
 
- Report updates are folded directly into diagnostic batch publication, removing the redundant published-URI set and intermediate updates allocation while preserving existing push behavior for absent, empty, and cleared diagnostics. Since pull requests no longer mutate the cache, they use a read lock and can execute concurrently.  
 
c6bf28598895c7b44039e43f6afdd6f3


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1012, opened 2026-07-23
feat(lsp): add selection range support

0xKarl98:
Towards OSS-508  

- This adds support for the LSP `textDocument/selectionRange` request, making editor selection expansion syntax-aware for Solidity and inline Yul. The server parses the current document on demand, collects nested ranges for identifiers, expressions, statements, blocks, declarations, types, parameters, and call arguments, and returns a strictly nested parent chain ending at the full document.

- Requests prefer unsaved VFS contents and fall back to disk for closed files, so selection ranges do not depend on background semantic analysis being ready. The parser recovers useful ranges from incomplete source where possible; comments, whitespace, and unparseable input fall back to the cursor and full-document ranges. Multiple requested positions retain their original order.

- Position conversion follows LSP UTF-16 semantics across LF, CRLF, and standalone CR line endings. Columns past a logical line end are clamped, while nonexistent lines and positions inside surrogate pairs are rejected. Parsing runs on the blocking pool so it does not block the async LSP runtime.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1016, opened 2026-07-24
feat(lsp): report workspace indexing progress

0xKarl98:
towards OSS-520  

This makes long workspace indexing runs visible in editors instead of leaving users guessing. The language server now reports the read, analyze, and publish stages, stays quiet for quick runs, and keeps one continuous progress update when newer changes replace work already in flight.

Clearing the analysis cache ends the progress immediately, while stale workers are stopped from restoring outdated diagnostics or navigation data.

Analyzing : 
截屏2026-07-24 20 18 03

Ready : 
截屏2026-07-24 20 25 23


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1025, opened 2026-07-24
feat(lint): add extensible lint framework

mablr:
Migrates foundry's lint infra to solar-lint, allowing downstream users to register early, late, and project-wide passes with fresh state and configurable policies. The design contains no Foundry-specific concepts and and will allow to preserve existing Forge lint behavior and diagnostics.

Cancellation handling is deferred until an LSP integration needs it, where it can be checked before starting visitors without affecting lint passes or traversal callbacks.


crates/lint/src/context.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub fn get_span_indentation(&self, span: Span) -> usize {
+        if !span.is_dummy() {
+            let loc = self.sess.source_map().lookup_char_pos(span.lo());
+            if let Some(line_text) = loc.file.get_line(loc.line) {
+                let col_offset = loc.col.to_usize();
+                if col_offset <= line_text.len() {
+                    let previous = &line_text[..col_offset];
+                    return previous.len() - previous.trim().len();
+                }
+            }
```
mattsse: `Loc::line` is 1-based while `SourceFile::get_line` expects a 0-based index, so spans on the final source line incorrectly return zero indentation. Also, `Loc::col` is a character offset but is currently used as a byte index, which can panic when valid UTF-8 precedes the span. Using `trim_start` also avoids counting whitespace immediately before an inline item as indentation.

```suggestion
            if let Some(line_index) = loc.line.checked_sub(1)
                && let Some(line_text) = loc.file.get_line(line_index)
            {
                let col_offset = loc.col.to_usize();
                let byte_offset = line_text
                    .char_indices()
                    .nth(col_offset)
                    .map_or(line_text.len(), |(offset, _)| offset);
                let previous = &line_text[..byte_offset];
                return previous.len() - previous.trim_start().len();
            }
```

mattsse (changes requested):
The framework looks solid overall, but the indentation helper currently misreads source locations and can panic on valid UTF-8 input. Please address the inline finding.

crates/lint/src/context.rs
mablr: pre-existing in foundry

mablr: fixed here https://github.com/paradigmxyz/solar/pull/1025/commits/b904198a61d128a1ea3b667e94de6df63b9ba26d

mattsse (approved):
lgtm

PR #1035, opened 2026-07-26
feat(lsp): add call hierarchy support

0xKarl98:
Adds call hierarchy support to the LSP, so you can inspect incoming and outgoing calls directly in the editor. It covers functions, modifiers, and explicit constructor calls, while avoiding misleading partial results when workspace snapshots conflict.

- In a multi-root workspace, conflicting snapshots of the same caller must not produce a partial incoming call hierarchy. The server should return null , rather than showing only unaffected callers as a complete result. 
6592ad1e542be4b1d35a96fd0ec0bae0

### DIFF
--- a/crates/lsp/src/call_hierarchy.rs
+++ b/crates/lsp/src/call_hierarchy.rs
@@ -1,0 +1,532 @@
+//! Call hierarchy indexing.
+
+use crate::{
+    proto,
+    symbols::{DeclarationSymbol, SymbolId, remap_symbol_id},
+};
+use lsp_types::{
+    CallHierarchyIncomingCall, CallHierarchyItem, CallHierarchyOutgoingCall, Position, Range, Url,
+};
+use serde::{Deserialize, Serialize};
+use solar_interface::{
+    Span,
+    data_structures::{
+        Never,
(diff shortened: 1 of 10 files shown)


DaniPopes (approved):

crates/lsp/src/call_hierarchy.rs
```diff
@@ -0,0 +1,626 @@
...
+        position: Position,
+        declaration: Option<SymbolId>,
+    ) -> Option<Vec<CallHierarchyItem>> {
+        if let Some(key) = self.call_site_key(uri, position) {
+            return Some(vec![self.item(key)?.clone()]);
+        }
+        if let Some(key) = declaration.and_then(|symbol| self.key_by_symbol.get(&symbol)) {
+            return Some(vec![self.item(key)?.clone()]);
+        }
+        let key = self.enclosing_body_key(uri, position)?;
```
mattsse, line 373: Could we retain dropped call-site ranges and return `None` before this fallback? When a callee is noncanonical, preparing on its call site currently falls through here and returns the enclosing caller instead.

mattsse (commented):
One correctness edge case.

PR #1036, opened 2026-07-26
refactor(codegen): stop creating dummy values

DaniPopes:
Codegen now represents unit-valued expressions as `None` instead of allocating placeholder MIR values. Expression and call lowering use `Option<ValueId>`, void internal and function-pointer calls emit value-less call instructions, and zero-return external and library calls no longer reserve return data.

Unsupported value-producing paths now emit an error value instead of silently returning zero. Error detection happens at the expression-lowering boundary, builtin arguments are validated through one `Result`-based helper, and Yul builtins have separate unit- and value-returning paths.

Assembler relocations now carry their deferred label instead of a dummy immediate, and SSA locals synthesize Solidity's zero default only when read. Remaining zero immediates represent real values such as default initialization, memory offsets, sizes, and call operands.


mattsse:
this now rejects the function-pointer cast/invocation used by `testdata/console.sol`. `main` compiles it, but this head exits with `codegen does not support this call expression yet`; the codspeed failure reproduces the same regression (and this affects the Seaport idiom too). can we preserve the value-call fallback without reintroducing dummy unit values?

but perhaps this is expected for now?

DaniPopes:
it's unhandled on main too by silently returning a dummy value, need to fix

PR #1038, opened 2026-07-26
refactor(codegen): centralize contract codegen

DaniPopes:
Move contract bytecode dependency orchestration from the CLI and Standard JSON paths into `solar-codegen`. Callers pass `ContractSelection::Specific` or `All` to `emit_requested`, and the generator recursively builds required creation-bytecode dependencies while caching each artifact. Standard JSON now generates only contracts whose selected outputs need bytecode.

Keep EVM IR modules in the artifacts instead of serializing them early. Module names are `Symbol`s, deployment IR stores only the deployment prefix, runtime IR stays in its existing field, and CLI dumps compose both without cloning the runtime module. UI coverage pins selected-contract generation and recursive contract-creation diagnostics.


mattsse (approved):
lgtm

PR #1046, opened 2026-07-27
feat(codegen): lower clz and callcode

DaniPopes:
Lower the Yul `clz` and `callcode` builtins through MIR into their native EVM opcodes. `clz` is gated to Osaka-compatible targets, matching solc, while `callcode` carries the same conservative memory and state effects as other mutating external calls.

The `extcall`, `extdelegatecall`, and `extstaticcall` names remain unsupported because their EOF opcodes are not part of solc's ordinary EVM Yul dialect.


mattsse (approved):
lgtm

PR #1048, opened 2026-07-27
feat(codegen): simplify clz expressions

DaniPopes:
Adds local MIR simplifications for `clz` based on its exact EVM result range of 0 through 256. Predicate, comparison, shift, byte, mask, division, modulo, and sign-extension consumers now collapse to cheaper operations or constants, and `clz` folds to zero when its operand is known to have the sign bit set.

The transforms stay in MIR, allowing the existing optimization pipeline to remove obsolete `clz` instructions before stack scheduling.


mattsse (approved):
lgtm

PR #1050, opened 2026-07-27
feat(lsp): add CodeLens support

0xKarl98:
Add semantic CodeLens annotations for ABI selectors, reference counts, and direct contract inheritance. The language server derives them from merged semantic analysis, deduplicates references across analysis batches, and suppresses results when source snapshots or compile-context-dependent facts disagree. It also negotiates refresh support so active lenses update after analysis is published or the cache is cleared.
截屏2026-07-27 20 09 20

- Click selector above and it will copy 4-byte selector automatically 

- Click reference above :
截屏2026-07-27 20 19 54

- Click 'base contract ' above : 
截屏2026-07-27 20 22 16

- Click 'derive contract 'above :

截屏2026-07-27 20 22 48

### DIFF
--- a/crates/lsp/src/symbols.rs
+++ b/crates/lsp/src/symbols.rs
@@ -315,7 +315,7 @@ impl SymbolTables {
 
         tables.build_type_definitions(gcx, &item_symbols);
         tables.call_hierarchy = CallHierarchyIndex::build(gcx, &item_symbols, &tables.declarations);
-        tables.code_lens = CodeLensIndex::build(gcx, &item_symbols, &tables.declarations);
+        tables.code_lens = CodeLensIndex::build(gcx, &item_symbols);
 
         // Public state-variable getters are compiler-generated functions, but source member calls
         // such as `this.value()` still name the state variable. Point getter resolutions back to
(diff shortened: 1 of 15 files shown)


crates/lsp/src/symbols.rs
```diff
@@ -1096,7 +1194,42 @@ impl SymbolTables {
...
+    ) -> Option<ReferenceTargets> {
+        if !self.has_merged_batches {
+            // Duplicate declarations and conflicting snapshots can only come from aggregation.
+            debug_assert!(self.rename.conflicting_contents().is_empty());
+            return self.symbol_ids_at_position(uri, position);
+        }
+        if self.rename.conflicting_contents().contains(uri) {
+            return None;
+        }
+        if let Some(reference) = self.reference_at_position(uri, position) {
```
mattsse, line 1210: Please merge the targets from all identical reference occurrences at this URI/range before looking up references, and add a two-batch shared-dependency regression test. Today this selects one batch’s `SymbolId`, so Find References from a shared dependency silently omits usages from other workspace roots.

mattsse (changes requested):
One merged-analysis issue to address.

crates/lsp/src/symbols.rs
0xKarl98: Have fixed

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1053, opened 2026-07-27
fix(lsp): debounce workspace analysis

DaniPopes:
Frequent document changes currently start a full-workspace analysis for every edit, allowing blocking workers to accumulate and saturate the CPU. This makes analysis single-flight, replaces stale pending work with the newest snapshot, and waits 250 ms after source changes before starting another run.

Work-done progress starts only after the current analysis acquires the scheduler and stays silent unless that analysis is still active after 100 ms. Replacing an invisible analysis resets the progress delay, so quick or superseded work does not flash in the editor.

The LSP runtime timings now live in its existing `Config` and are passed to analysis, progress, formatting, and flycheck. The workspace scanner also treats Foundry's `target` directory as expensive so initialization avoids traversing generated artifacts.

Forge lint parsing now accepts non-diagnostic JSON records, collects diagnostics from both output streams, and ignores plain Foundry warnings mixed with JSON output.

### DIFF
--- a/crates/interface/src/session.rs
+++ b/crates/interface/src/session.rs
@@ -175,7 +175,7 @@ impl SessionBuilder {
             thread_pool: OnceLock::new(),
         };
         sess.reconfigure();
-        debug!(version = %solar_config::version::SHORT_VERSION, "created new session");
+        debug!(version = %solar_config::version::SEMVER_VERSION, "created new session");
         sess
     }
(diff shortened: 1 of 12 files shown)


DaniPopes:
cc @0xKarl98

mattsse (approved):
lgtm

PR #1057, opened 2026-07-27
fix(codegen): pack sub-word storage like solc

Ayush7614:
State variables such as `uint8`, `address`, and `enum` were each given a full storage slot even though the sema storage-layout emitter already packs them the way solc does. That diverges layout for any contract with consecutive sub-word scalars and breaks slot-sensitive code.

This expands right-aligned packing in codegen to those types, keeps left-aligned `bytesN`/external function pointers on the full-slot path until load/store can shift them correctly, and sign-extends packed `intN` loads so ABI and arithmetic see a proper signed word. Runtime `run-call` coverage checks RMW neighbor preservation, full-word spill alignment, signed packing, and `address`+`uint96` sharing a slot.

### DIFF
--- a/crates/codegen/src/lower/expr.rs
+++ b/crates/codegen/src/lower/expr.rs
@@ -665,8 +665,17 @@ impl<'gcx> Lowerer<'gcx> {
                             return self.materialize_storage_bytes(builder, slot_val);
                         }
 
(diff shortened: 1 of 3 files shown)


decofe:
Superseded by #1062, which carries this change forward with CI fixes.

PR #1058, opened 2026-07-27
fix(sema): include library mapping params in hashes

Ayush7614:
Library functions that take `mapping(...) storage` parameters were silently dropped from the external interface, so `--emit=hashes` omitted selectors that solc emits. Signature printing already knew how to render those types; `interface_functions` was the only gate.

Libraries may expose mapping storage parameters (`interfaceType(true)` in solc). Keep those functions in the interface when the only non-exportable piece is a mapping, and pin the existing `get(mapping(address => S) storage,address)` selector in the library signature encoding UI test.


decofe:
Superseded by #1061, which carries this change forward with CI fixes.

PR #1059, opened 2026-07-27
fix(sema): reject internal types in external fn types

Ayush7614:
Solc rejects internal types in the parameter or return list of an external function type (TypeError 2582). We previously accepted programs such as `function(function () internal) external`, which cannot exist at the ABI boundary.

The type checker now walks external `FunctionTypeName` parameters and returns and errors when `can_be_exported` fails, covering nested external function types as well.


decofe:
Superseded by #1063, which carries this change forward with CI fixes.

PR #1060, opened 2026-07-27
fix(ast): reject unchecked as bare loop/if body

Ayush7614:
Solc requires `unchecked` blocks to appear inside a regular block. We accepted them as the sole body of `while`/`do`/`for`/`if`, which is invalid Solidity and was a skipped solc fixture (`unchecked_while_body`).

AST validation now rejects that shape with the same diagnostic family as bare variable declarations, and the corresponding solc skip is removed.


decofe:
Superseded by #1064, which carries this change forward with CI fixes.

PR #1064, opened 2026-07-28
fix(ast): reject unchecked as bare loop/if body

decofe:
Solc requires `unchecked` blocks to appear inside a regular block. Solar enforces the same rule during AST validation, covering bare `while`/`do`/`for`/`if`/`else` bodies.

The parse-only behavior intentionally differs: Solar's parser accepts the statement grammar and AST validation rejects the invalid placement, while solc rejects it during parsing. This keeps the upstream `unchecked_while_body` fixture excluded from parse-only parity testing, removes the stale TODO, and documents the phase difference in `docs/SOLC_DIVERGENCE.md`.

Replaces #1060.

Prompted by: @DaniPopes

### DIFF
--- a/crates/sema/src/ast_passes.rs
+++ b/crates/sema/src/ast_passes.rs
@@ -54,6 +54,10 @@ impl<'sess> AstValidator<'sess, '_> {
                 .help("wrap the statement in a block (`{ ... }`)")
                 .emit();
         }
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1067, opened 2026-07-28
fix(codegen): order base constructors

DaniPopes:
Split constructor lowering into the same semantic phases as solc: state initialization, inheritance-argument evaluation in the derived constructor’s context, and constructor-body execution from base to derived. This preserves argument side effects and values across indirect base constructors instead of substituting zero or running bodies in argument-evaluation order.

The runtime cases cover the upstream side-effect order, arguments propagated through an intermediate constructor, dynamic strings, and legacy state-initializer ordering.


mattsse (approved):
lgtm

PR #1069, opened 2026-07-28
fix(codegen): rebind storage references

DaniPopes:
Represent reassigned storage-reference locals and parameters as mutable slot values across control-flow joins. Assignments lower the right-hand lvalue to its slot rather than loading its contents.

A semantic definite-assignment pass now rejects reads of unbound storage references instead of letting codegen acquire slot zero. It follows branch intersections, short-circuit expressions, ternaries, loops, try/catch, switches, Yul `.slot` assignments, and named storage returns while suppressing follow-on diagnostics for expressions and types already known to be erroneous. Runtime and diagnostic cases cover branch assignment, reassignment, local reads, assembly assignment, and storage return variables.


mattsse (commented):
Two storage-reference rebinding cases still need fixes:

1. Mutable storage-reference parameters are stored via `local_memory_addr` before all parameters/returns are registered, so the address can differ from later loads. A `Status storage status, bool second` parameter retains slot zero when the branch does not rebind it. Please defer these stores until the frame prefix is final (or compute from final counts), and add a runtime test with the storage reference before another parameter.

2. Tuple assignment returns before the new slot-lowering path; `lower_tuple_assign` evaluates `statuses[2]` with `lower_expr` and stores its contents as the pointer. `(status,) = (statuses[2], 0)` therefore silently writes through the wrong slot. Please lower RHS elements destined for storage-reference lvalues with `lower_lvalue_slot` and add tuple-rebinding coverage.

PR #1076, opened 2026-07-28
feat(lsp): handle workspace file operations

0xKarl98:
Add LSP support for workspace create, rename, and delete operations. Keep imports, workspace roots, VFS buffers, diagnostics, and file-watcher events synchronized across file moves, including Unicode and platform-specific paths


crates/lsp/src/file_operations.rs
```diff
@@ -0,0 +1,1066 @@
...
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.moves.is_empty()
+    }
+
+    pub(crate) fn map_path(&self, path: &Path) -> Option<(FileMoveId, PathBuf)> {
+        let path = path.normalize();
+        self.moves.iter().enumerate().find_map(|(index, file_move)| {
+            let suffix = path.strip_prefix(&file_move.old_path).ok()?;
+            Some((FileMoveId(index), file_move.new_path.join(suffix)))
```
mattsse: `join`ing the empty suffix for an exact move appends a separator. For example, `C:\src\Target.sol -> D:\contracts\Renamed.sol` currently rewrites the import as `"D:/contracts/Renamed.sol/"`, which no longer resolves. Return the destination unchanged when `suffix.as_os_str().is_empty()` and apply the same fix in `reverse_map_path`. A focused regression test is `assert_eq!(batch.map_path(&old).unwrap().1, new)` plus the reverse assertion for exact file paths; the existing cross-drive/UNC tests should then pass.

mattsse (changes requested):
Windows CI is also failing in `aggregator_preserves_document_links_without_declarations`: construct the fixture URI with `Url::from_file_path(temp_dir().join(...))` instead of hard-coding `file:///workspace/...`.

crates/lsp/src/file_operations.rs
0xKarl98: Have fixed

mattsse (approved):
lgtm

crates/interface/src/source_map/file_resolver.rs
```diff
@@ -415,6 +374,52 @@ impl<'a> FileResolver<'a> {
     }
 }
 
+/// Applies `remappings` to `path` using the same selection rules as [`FileResolver`].
+pub fn apply_import_remappings<'a>(
```
DaniPopes, line 378: please don't modify this file and use file resolver directly

DaniPopes (changes requested):

DaniPopes (approved):

PR #1077, opened 2026-07-28
feat(codegen): add disassembly dumps

DaniPopes:
Adds `-Zdump=disasm-deploy` and `-Zdump=disasm-runtime` so codegen output can be checked at the opcode level without raw byte snapshots. Deployment disassembly stops before the appended runtime bytecode.

Source compilation and standalone EVM IR inputs share the production formatter. Standalone input runs and validates through the normal unified IR pipeline before assembly, diagnoses unresolved deferred pushes, and normalizes sparse textual block labels.

Moves the assembler disassembly cases into `tests/ui/codegen/disasm` as Solidity and EVM IR UI fixtures with FileCheck coverage for push selection, label relaxation, block layout, deduplication, and deploy/runtime separation. Retained assembler unit coverage keeps its exact disassembly snapshot assertions.


mattsse (approved):
lgtm

PR #1083, opened 2026-07-29
feat(lsp): refresh pull results

0xKarl98:
Towards OSS-597

Pull diagnostics and inlay hints are normally requested again after a document edit, but analysis results can also change when watched files, workspace configuration, flychecks, or the analysis cache change. The existing push-diagnostic path and edit-triggered pulls can mask this, while clients may keep stale pull result IDs or cached hints when the document itself is unchanged.

This change records the client's independent diagnostic and inlay-hint refresh capabilities and sends each refresh only when its corresponding result actually changes. Refresh intent is coalesced across superseded analyses, survives a current-worker failure, and is completed when a later successful analysis publishes. Clearing the analysis cache refreshes the empty state, and any invalidated recovery refreshes the restored diagnostics and hints.

Diagnostic result IDs remain stable for unchanged reports, flycheck-owner removal is batched, and inlay-hint comparisons are skipped unless a supported refresh may be needed. The initialization compatibility shim also recognizes the plural workspace diagnostics capability used on the LSP wire.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1084, opened 2026-07-29
feat(fuzz): add symbolic compiler differentials

figtracer:
Adds an opt-in `fuzz/bin/solsymdiff` command that symbolically compares one explicitly selected Solc/Solar function compiled from the same closed Standard JSON input.

The focused runner supports pure functions by default and explicit zero-state view or single-call nonpayable modes. It preserves project imports/remappings, optimizer/via-IR/EVM settings, dynamic input shapes, search limits, exact success/revert data, and stateful target storage/log comparison. A finding is reported only after Foundry concretely replays the counterexample; the generated project, shared compiler input, result, and replay artifact are retained.

Automatic function discovery and campaign scheduling, independent Anvil replay, multi-call or nonzero initial state, payable calls, and external-contract side-effect comparison are intentionally outside this primitive. A caller such as Derek can select targets and matrix settings without duplicating the differential oracle or imposing a campaign policy here.

This PR was developed with AI assistance; the final scope, code, tests, and real-engine results were reviewed and executed by the author.


mattsse (approved):
lgtm

fuzz/fandango/README.md
```diff
@@ -282,6 +282,60 @@ bytecode blobs into one Foundry test, installs them at fixed addresses with
 `vm.etch`, and lets Foundry fuzz calldata/caller/environment values while the
 test compares return bytes, revert behavior, logs, and normalized state diffs.
 
+### Symbolic Solc-vs-Solar differential
```
DaniPopes, line 285: can you also add a small link to agents.md, and also to use it when comparing semantics to solc?

figtracer: 89d2e16f6290a84720ad7dd30a653440fc5fe9a6 added here

DaniPopes (approved):

PR #1081, opened 2026-07-28
feat: enable more codegen benches

djolertrk:
- Enable multi-file projects as benches
- Fix performance along the way
- Fix bugs in both sema and codegen

PR #1089, opened 2026-07-29
feat: enable more codegen benches

DaniPopes:
Builds on #1088 and replaces the codegen and benchmark portion of #1081. It adds whole-project Foundry inputs and the large-contract runtime corpus, then expands codegen support for the Solidity features those projects exercise, including aggregate ABI decoding, typed try/catch, dynamic storage copies, calldata slices, user-defined operators, and pre-Cancun memory copying.

The remaining commits improve MIR and backend performance, bound the CodSpeed workload, and make the benchmark workflow cancel stale runs and stream large deployments.


DaniPopes:
Superseded by #1091 and #1092. The project benchmark harness is now an independently mergeable base; the codegen changes and runtime gate sit above it.

PR #1102, opened 2026-07-30
feat(cli): expand standard JSON codegen support

DaniPopes:
This follows #1003 by retaining assembled immutable references in contract artifacts and exposing their source identifiers, byte offsets, and typed widths.

Standard JSON output now includes solc-compatible opcode strings, empty link-reference maps for fully linked bytecode, and deployed-bytecode immutable reference maps. Fine-grained output selection emits only the requested fields without cloning or serializing the bytecode object when it is not selected.

Standard JSON settings now accept typed library addresses and the supported optimizer fields. Library lookup preserves source qualifiers, handles same-name libraries in different source units, and falls back to unqualified CLI entries. Library addresses use `Address` throughout parsing, configuration, and codegen.


mattsse (commented):
Found one issue with the Standard JSON optimizer mapping.

crates/cli/src/standard_json/compile.rs
```diff
@@ -152,14 +154,24 @@ fn compile(
...
-    // override when the input explicitly disables the optimizer, leaving the
-    // CLI-driven default otherwise. `runs` and `details` have no analogue in the
-    // MIR optimizer yet, so they're parsed but unused.
-    if let Some(Optimizer { enabled: false, runs: _runs }) = optimizer {
-        opts.optimization = OptimizationMode::None;
+    if let Some(Optimizer { enabled, runs }) = optimizer {
+        // 200 runs is the default value if unspecified in solc.
+        // Treat lower values as Size.
+        opts.optimization = if enabled {
+            if runs >= Some(200) { OptimizationMode::Gas } else { OptimizationMode::Size }
```
mattsse, line 161: `runs` is an `Option<u64>`, so `None >= Some(200)` is false. As a result, `{"optimizer":{"enabled":true}}` selects `OptimizationMode::Size`, even though omitted `runs` defaults to 200 in solc and should follow the same path as explicit `"runs": 200` (Gas in this approximation). This is observable in generated bytecode—for example, a signed immutable produces different output for omitted versus explicit 200 runs—while solc produces identical output. Please use the default before comparing (for example, `runs.unwrap_or(200) >= 200`) and add a regression test for omitted `runs`.

DaniPopes: thx boss

DaniPopes: d2906f570532982a9463bbb741e7ac00af44c508

PR #1111, opened 2026-07-31
feat(lsp): add workspace diagnostics

0xKarl98:
Towards OSS-595  

### Workspace Diagnostic Reports

- When a workspace is opened, the client sends a workspace/diagnostic request, and the server returns reports for every analyzed Solidity file, including clean files and imported dependencies.

- When the client supplies valid previousResultIds, unchanged files return unchanged reports with the same result IDs, while changed files return new full reports.

### Workspace Membership Changes

- When New.sol is added, the server sends workspace/diagnostic/refresh, and the following workspace pull includes the new file.

- When New.sol is removed, the server sends another refresh, returns one empty full report to clear its cached diagnostics, and omits the file after that result is acknowledged.

- When a workspace folder is removed, diagnostics belonging only to that folder are cleared immediately, while files covered by a nested workspace remain available.

 ### Streaming, Progress, and Cancellation

- When a request includes a partialResultToken, the server streams up to 64 reports per $/progress notification and returns an empty final items array after all batches are sent.

- When a request includes a workDoneToken, the server sends begin and end progress notifications titled Workspace diagnostics.

- When the client cancels a workspace diagnostic request, the server stops sending further batches, ends work-done progress, and returns RequestCancelled.

### Analysis Snapshot Consistency

- When an open document receives a new version without content changes, the server avoids redundant analysis and refresh requests while reporting the latest version in the next workspace pull.

- When document contents change but the new analysis fails, the server preserves the last successful diagnostics with their original document version instead of relabeling stale results.

- When workspace analysis is still running, diagnostic requests wait for the latest published result, and concurrent requests share that result without starting duplicate analyses.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1113, opened 2026-07-31
fix(capi): build missing test cdylib

0xKarl98:
Current nightly Cargo no longer exposes package `cdylib` outputs through the integration-test loader path, causing the `solar-capi` C smoke test to fail in a clean target directory even though stable Cargo still provides the legacy path.

This keeps the existing loader lookup for stable builds, limits it to the active target directory, and falls back to building the C API library with the active `nightly` and `asm` features. The smoke test therefore uses the matching public debug artifact instead of depending on stale loader state.

This unblocks the nightly CI failure observed on #1111 without coupling the fix to its LSP changes.

### DIFF
--- a/crates/capi/tests/it/ctest.rs
+++ b/crates/capi/tests/it/ctest.rs
@@ -17,9 +17,7 @@ fn c_api_smoke_test() {
         eprintln!("skipping C API smoke test because no C compiler was found");
         return;
     };
-    let lib_path = dynamic_library().unwrap_or_else(|| {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

djolertrk:
lgtm! I needed this for https://github.com/paradigmxyz/solar/pull/1112 as well.

DaniPopes:
the cdylib target should be built before the test is, can you please just tweak the path? you shouldn't need to "cargo build" inside of a cargo test

DaniPopes:
https://github.com/paradigmxyz/solar/pull/1114

PR #1116, opened 2026-07-31
Perf(SIMD): add SIMD acceleration for long whitespace and identifier runs

malik672:
image
image

Pre and Post for uniswap and seaport 
well thanks to Tanner Gooding for the lexer SIMD explanation
We can do this for x86 too and even more effectively


DaniPopes:
not worth, we can do in the future with std simd

PR #1119, opened 2026-07-31
test: run foundry comparisons in ci

DaniPopes:
Foundry comparisons run as a separate `#[test]` in `solar-tester`, so the existing `cargo nextest run --workspace` CI command discovers them without an extra workflow step. The runner scans `tests/foundry` for every `foundry.toml` project, runs each Solar/solc comparison, and keeps the Solar-only stack-depth suite.

The broader suite also fixes dynamic ABI return and constructor encoding, modifier expansion, aggregate storage assignment and deletion, and compiler-specific assumptions in the Unifap fixtures.


DaniPopes:
Superseded by the stacked PRs #1126 and #1127.

PR #1124, opened 2026-08-01
feat(lsp): add protocol tracing

0xKarl98:
This adds LSP protocol tracing that honors the trace level supplied during initialization and later `$/setTrace` updates. When tracing is enabled for a client request, the server emits one `$/logTrace` notification after that request completes, including successful, failed, and cancelled requests. Message mode reports the method and outcome, while verbose mode also reports server processing time. Client notifications and request-start lifecycle events are intentionally not traced.

Trace messages omit parameters, results, error details, request IDs, and sequence numbers. Method names that are not bounded ASCII LSP-style names are redacted. Each completion trace is queued before its corresponding response so clients observe deterministic wire ordering.

A successful request completion:

Successful request completion trace

A failed request completion:

Failed request completion trace

A cancelled request completion:

Cancelled request completion trace


mattsse (approved):
lgtm

editors/vscode/package.json
```diff
@@ -83,6 +83,15 @@
           "default": true,
           "description": "Format files on save using Solar LSP"
         },
+        "solarLsp.trace.server": {
```
DaniPopes, line 86: what about zed?

crates/lsp/src/tests/router.rs
```diff
@@ -952,3 +956,6 @@ async fn initialized_registers_watched_files_when_client_supports_dynamic_regist
     assert!(server_main.await.unwrap().is_ok());
     assert!(matches!(client_main.await.unwrap(), Err(async_lsp::Error::Eof)));
 }
+
+#[path = "protocol_trace.rs"]
+mod protocol_trace_tests;
```
DaniPopes: why is this not declared together with the other modules in mod.rs?

editors/vscode/package.json
0xKarl98: Unlike vscode , zed has its built-in trace_level_menu here :
https://github.com/zed-industries/zed/blob/v1.13.1/crates/language_tools/src/lsp_log_view.rs#L1167-L1181

when users switch level , Zed sends $/setTrace directly thru :
https://github.com/zed-industries/zed/blob/v1.13.1/crates/language_tools/src/lsp_log_view.rs#L610-L631

crates/lsp/src/tests/router.rs
0xKarl98: Was thinking to make smallest code changes  , but have centralized it now

DaniPopes (approved):

PR #1125, opened 2026-08-01
feat(test): remove external bench

djolertrk:


testdata/codegen-runtime/README.md
DaniPopes, line 1: let's move to benches/ dir and import the projects in the same format as the other, in the json.gz

DaniPopes (approved):

PR #1130, opened 2026-08-03
feat(lsp): add diagnostic quick fixes

0xKarl98:
Adds `textDocument/codeAction` support for compiler, flycheck, and Forge lint diagnostics. Quick fixes preserve structured suggestions, provide guarded fallbacks for common Solidity diagnostics, and validate edits against the current document before applying them.

- Converts a payable `fallback` function to `receive` when no receive function is defined.

Quick fix to convert a payable fallback function to receive

- Marks contracts with unimplemented functions as `abstract`.

Quick fix to mark a contract as abstract

- Removes deprecated names from function type return parameters.

Quick fix to remove a named function type return parameter

- Adds a missing `override` specifier to an overriding function.

Quick fix to add an override specifier

- Removes an unused named import reported by Forge lint.

Quick fix to remove an unused import

- Converts mutable variable names to Solidity `mixedCase` style.

Quick fix to convert a variable name to mixedCase

- Adds missing SPDX license and compiler pragma headers.

Quick fixes to add SPDX and Solidity pragma headers

- Restricts eligible functions to `view` based on flycheck diagnostics.

Quick fix to change function state mutability to view

### DIFF
--- a/crates/interface/src/diagnostics/emitter/json.rs
+++ b/crates/interface/src/diagnostics/emitter/json.rs
@@ -105,7 +105,12 @@ impl JsonEmitter {
             .children
             .iter()
             .map(|sub| self.sub_diagnostic(sub))
-            .chain(diagnostic.suggestions.iter().map(|sugg| self.suggestion_to_diagnostic(sugg)))
+            .chain(
+                diagnostic
+                    .suggestions
+                    .iter()
+                    .flat_map(|suggestion| self.suggestion_to_diagnostics(suggestion)),
+            )
             .collect();
 
         JsonDiagnostic {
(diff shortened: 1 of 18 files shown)


mattsse (changes requested):
Two issues need addressing before merge:

- **[P2] Snapshot every file covered by the flycheck.** `crates/lsp/src/global_state.rs:1124` builds the trusted snapshot from `Workspace::source_files()` plus the saved file, but a Foundry workspace only populates that cache from its configured `src` root. Saving `src/A.sol` runs the default unscoped `forge lint --json`, which can also report an actionable diagnostic in `test/B.t.sol` or `script/Deploy.s.sol`. Those files are absent from the snapshot, so their diagnostics get no `data`, and `code_actions::plans` rejects them. The existing snapshot test covers only a tracked `src` file and the saved file.

- **[P2] Preserve the public `SourceLocation` offset types.** `crates/interface/src/diagnostics/emitter/json.rs:483` changes the publicly re-exported `start` and `end` fields from `u32` to `i64`. Downstream code constructing `SourceLocation` from typed `u32` offsets will stop compiling after this update. The current tests use untyped integer literals, so they do not catch the source-compatibility break. Please handle solc's `-1` sentinel through a private/input adapter or another approach that keeps this public emitter API stable.

0xKarl98:
@mattsse  Have fixed !

mattsse (approved):
lgtm

crates/interface/src/diagnostics/emitter/json.rs
DaniPopes, line 1: fyi this file is adapted from rustc, so I would first check if/how rust-analyzer handles these or if this file was not integrated correctly

0xKarl98: I checked both current rustc and rust-analyzer. rustc still flattens substitutions, while rust-analyzer treats each child as one atomic fix, so the ambiguity remains upstream. Our internal Substitution already represents one complete alternative, and the end-to-end test verifies that preserving this boundary produces two valid multipart actions. 

This is therefore an intentional Solar-specific divergence from rustc’s current JSON shape, not a claim that [#53934](https://github.com/rust-lang/rust/issues/53934) has been adopted upstream.

DaniPopes (approved):

PR #1134, opened 2026-08-03
perf(codegen): return static calls on stack

djolertrk:
Return one-word results from non-recursive static callees on the EVM stack and remove their unused frame word. Keep narrowly eligible single-use stack arguments in place as well.


djolertrk:
Depends on https://github.com/paradigmxyz/solar/pull/1133

PR #1135, opened 2026-08-03
perf(codegen): inline tiny leaf helpers

djolertrk:
Inline gas-mode leaf helpers with one block, no nested call or control flow, at most four MIR instructions, and at most eight estimated bytes of growth. This isolates the measured profitable subset from the regressive general inliner.


djolertrk:
depends on https://github.com/paradigmxyz/solar/pull/1134

PR #1136, opened 2026-08-03
perf(codegen): compact static frames

djolertrk:
Runtime static frames never read the dynamic frame header, so omit its two words and reuse the low-memory pointer slot when a module has no dynamic internal calls. Preserve constructor and dynamic frame layouts.


djolertrk:
depends on https://github.com/paradigmxyz/solar/pull/1135

PR #1137, opened 2026-08-03
fix(codegen): panic on zero modular builtin

figtracer:
Solidity’s high-level `addmod` and `mulmod` builtins must revert with `Panic(0x12)` when their modulus is zero. The backend currently lowers them directly to the raw EVM opcodes, whose native zero-modulus behavior is returning zero, so generated contracts diverge from Solc.

This adds the missing high-level zero guard through the existing panic path and covers the MIR and runtime behavior under each optimizer objective. Raw Yul `addmod` and `mulmod` remain unchanged and continue returning zero for a zero modulus.

The mismatch was found by the symbolic Solc-versus-Solar differential work in #1084 and is split out here so the compiler correction can be reviewed independently.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1139, opened 2026-08-04
feat(lsp): bound workspace indexing

0xKarl98:
Towards OSS-516 

Bounds workspace discovery and file watching to project sources, skipping generated, dependency, hidden, and nested-repository trees by default while loading imported files on demand. Adds configurable indexing exclusions and reduces unnecessary indexing overhead in large workspaces.


crates/lsp/src/workspace/mod.rs
```diff
@@ -168,58 +201,123 @@ impl<'a> WorkspacePathIndex<'a> {
...
-
-fn is_ignored_naked_directory(root: &Path, directory: &Path) -> bool {
-    directory != root && (is_heavy_dir(directory) || directory.join(".git").exists())
+impl SourceFileCollector<'_> {
+    fn collect(&mut self, path: &Path) -> bool {
+        if self.cancellation.is_cancelled() {
+            return false;
+        }
+        self.metrics.visited += 1;
+        if self.import_only_roots.iter().any(|root| path.starts_with(root)) {
```
mattsse: An explicit source root can overlap an include path. For example, `src = "lib"` also gets the default `libs = ["lib"]`, so this prunes the source root itself and indexes zero closed files. Please let explicit source roots win over import-only ancestors and add an overlap regression test.

crates/lsp/src/config.rs
```diff
@@ -219,8 +254,99 @@ impl Config {
         &self.workspace_roots
     }
 
+    pub(crate) fn watched_file_specs(&self) -> Vec<WatchedFileSpec> {
+        let mut specs = BTreeSet::new();
+        for root in &self.workspace_roots {
+            for pattern in ["**/*.sol", "**/foundry.toml", "**/remappings.txt"] {
```
mattsse: These workspace-root globs still make the client watch excluded trees such as `node_modules`, build output, hidden directories, and nested repositories. Server-side event filtering does not avoid watcher traversal. Please register Solidity watchers from the bounded source roots (and targeted config paths) instead, with coverage showing excluded trees are not included.

mattsse (changes requested):
Two edge cases still need tightening before merge.

crates/lsp/src/config.rs
0xKarl98: Have resolved these two cases

crates/lsp/src/tests/mod.rs
DaniPopes, line 1: what's happening in this file?

0xKarl98: Have refactored to make it clean

mattsse:
oops @0xKarl98 this now has a few conflicts unfortunately

DaniPopes:
cyclops audit

mattsse:
cyclops audit

mattsse (approved):
The outstanding indexing touchups are addressed at 8c8058c0: default exclusions apply beneath explicit roots, approved missing import parents are watched, and didCreate directory echo scanning is bounded and policy-aware. The full CI and benchmark matrix is green.

PR #1145, opened 2026-08-05
fix(codegen): truncate narrow left shifts

figtracer:
High-level left shifts currently preserve the full 256-bit EVM `SHL` result even when the expression type is `uintN` or `intN`. Bits outside the declared width can therefore affect later control flow and escape through ABI output.

This restores the typed-value invariant after `SHL`: unsigned sub-word results are masked and signed results are truncated and sign-extended. Full-width shifts, right shifts, and raw Yul retain their existing EVM semantics.

The benchmark corpus has one expected output-size change: Nitro's runtime grows from 14,341 to 14,357 bytes and deployment gas grows by 3,446. Disassembly accounts for all 16 bytes as the three required width-normalization sequences for its `uint16`, `uint32`, and `uint64` `ret <<= 8` loops. Hot runtime gas, all 149 CodSpeed cases, and all 25 Gungraun instruction benchmarks are unchanged.

### How this was found

The automatic differential in #1084 compiled the same `pure` fixture with solc and this compiler, then used Foundry's symbolic EVM to search for calldata that changed the exact result. It found `uint8ShiftControl(1, 8)`, which returned `true` under solc and `false` under this compiler; Foundry replay and a separate fresh-Anvil replay both confirmed the witness.

### DIFF
--- a/crates/codegen/src/lower/checked_arith.rs
+++ b/crates/codegen/src/lower/checked_arith.rs
@@ -177,7 +177,10 @@ impl<'gcx> Lowerer<'gcx> {
             BinOpKind::BitAnd => builder.and(lhs, rhs),
             BinOpKind::BitOr => builder.or(lhs, rhs),
             BinOpKind::BitXor => builder.xor(lhs, rhs),
-            BinOpKind::Shl => builder.shl(rhs, lhs),
+            BinOpKind::Shl => {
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1146, opened 2026-08-05
fix(codegen): clean narrow bitwise not

figtracer:
Unary bitwise-not currently leaves a full-word complement for `uintN` and `bytesN`. That makes expressions such as `~uint8(0) == type(uint8).max` evaluate incorrectly, rather than only producing a noncanonical ABI word.

This cleans the result according to the expression type while preserving signed integers, untyped integer literals such as `~0`, full-width values, and raw Yul behavior.

### How this was found

The automatic differential in #1084 compiled the same `pure` fixture with solc and this compiler, then used Foundry's symbolic EVM to search for calldata that changed the exact result. It found `uint8NotControl(0)`, which returned `true` under solc and `false` under this compiler; Foundry replay and a separate fresh-Anvil replay both confirmed the witness.

### DIFF
--- a/crates/codegen/src/lower/expr.rs
+++ b/crates/codegen/src/lower/expr.rs
@@ -255,29 +255,7 @@ impl<'gcx> Lowerer<'gcx> {
                             let guar = self.emit_unsupported_udvt_operator(expr.span);
                             return builder.error_value(guar);
                         }
-                        let result =
-                            self.lower_unary_op(builder, *op, operand_val, int_info, expr.span);
-                        if op.kind == UnOpKind::BitNot {
-                            if let Some(width) = self.fixed_bytes_width_of_expr(expr) {
-                                return self.clean_fixed_bytes(
-                                    builder,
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

crates/codegen/src/lower/expr.rs
```diff
@@ -255,7 +255,29 @@ impl<'gcx> Lowerer<'gcx> {
                             let guar = self.emit_unsupported_udvt_operator(expr.span);
                             return builder.error_value(guar);
                         }
-                        self.lower_unary_op(builder, *op, operand_val, int_info, expr.span)
+                        let result =
+                            self.lower_unary_op(builder, *op, operand_val, int_info, expr.span);
+                        if op.kind == UnOpKind::BitNot {
```
DaniPopes: please do this inside of lower_unary_op

DaniPopes (approved):

PR #1147, opened 2026-08-05
fix(codegen): lower function modifiers

figtracer:
Function modifier invocations are present in HIR, but codegen currently lowers only the function body and treats each `_` placeholder as a no-op. Guards, prefixes, and suffixes can therefore be silently skipped while compilation still succeeds.

This expands modifier invocations as nested MIR continuations, preserving argument timing, lexical bindings across repeated substitutions, explicit-return joins, conditional and multiple placeholders, constructors, virtual overrides, public internal copies, and calldata-slice returns.

CodSpeed flags only the synthetic `many_modifiers_small/codegen` case; the other 148 measured benchmarks are unchanged. Exact local Criterion measurement is 124.83 µs on the base and 255.72 µs on this branch. The base emits only 90 bytes because it omits all ten `require` paths, while this branch emits 368 bytes; writing the same checks inline emits byte-for-byte identical code in 233.89 µs. Raw source-to-MIR work moves only from 56.83 to 60.41 µs, attributing almost all of the increase to optimizing the restored control flow rather than modifier bookkeeping. The runtime corpus's size increases are likewise modifier-bearing paths that were previously absent, and all 15 measured hot-gas workloads are unchanged.

### How this was found

The automatic differential in #1084 compiled the same `pure` fixture with solc and this compiler, then used Foundry's symbolic EVM to search for a success/revert or exact-result difference. It found `one(0)`, which reverted with `Error("zero")` under solc but returned `0` under this compiler because the modifier guard was skipped; Foundry replay and a separate fresh-Anvil replay both confirmed the witness.

### DIFF
--- a/crates/codegen/src/lower/call.rs
+++ b/crates/codegen/src/lower/call.rs
@@ -2081,7 +2081,7 @@ impl<'gcx> Lowerer<'gcx> {
             && !self.function_is_recursive(func_id)
             && self.try_enter_inline(func_id)
         {
-            let values = self.inline_slice_return_body(builder, func, &body, &arg_vals);
+            let values = self.inline_slice_return_body(builder, func, body, &arg_vals);
             self.exit_inline();
             let Some(first) = values.first().copied() else {
                 return self.err_value(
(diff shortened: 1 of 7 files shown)


mattsse (approved):
lgtm

figtracer:
https://github.com/paradigmxyz/solar/pull/1161 should be solving this properly

PR #1151, opened 2026-08-06
fix(codegen): mask packed signed fields

figtracer:
Static `abi.encodePacked` runs are coalesced into one word before being stored or hashed. Signed sub-word values arrive sign-extended; shifting them without first masking to their declared width lets the extension overwrite bytes belonging to earlier fields. For example, `abi.encodePacked(uint8(0), int16(-1), bytes3(0))` produced `ffffff000000` instead of `00ffff000000`, and the same contamination changed packed hashes.

This tracks signed packed arguments and masks only those terms before coalescing, leaving unsigned packing unchanged. The runtime regressions cover int8, int16, and int32 materialization and the in-place hash path.

The bug was found by the symbolic Solc-versus-Solar differential in #1084. Foundry produced a concrete witness, then its durable replay and an independent fresh-Anvil replay both confirmed the exact byte difference. Re-running the affected symbolic campaigns on this branch completed 21 of 21 functions with bounded agreement.

This PR was developed with AI assistance; the implementation, minimized reproduction, symbolic replays, and regression coverage were reviewed and executed by the author.

### DIFF
--- a/crates/codegen/src/lower/abi_packed.rs
+++ b/crates/codegen/src/lower/abi_packed.rs
@@ -6,7 +6,7 @@ use crate::{
     mir::{FunctionBuilder, MemoryObjectKind, Value, ValueId},
 };
 use alloy_primitives::U256;
-use solar_ast::{ElementaryType, LitKind};
+use solar_ast::{ElementaryType, LitKind, TypeSize};
 use solar_interface::{Symbol, diagnostics::ErrorGuaranteed, sym};
 use solar_sema::{
     builtins::Builtin,
@@ -18,7 +18,7 @@ enum PackedAbiArg {
     /// Compile-time literal bytes, packed without padding.
     Bytes(Vec<u8>),
     /// A single value occupying the top `size` bytes of its word.
-    Value { value: ValueId, size: usize, left_aligned: bool },
+    Value { value: ValueId, size: usize, left_aligned: bool, signed: bool },
     /// A memory `bytes`/`string` pointer (`[length][data...]`) whose data is
     /// copied without padding.
     DynamicBytes(ValueId),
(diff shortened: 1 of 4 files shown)


mattsse (dismissed):
[P2] Preserve signedness through user-defined value types

`ty.is_signed()` at `crates/codegen/src/lower/abi_packed.rs:217` does not look through `TyKind::Udvt`, so a signed UDVT such as `type Signed16 is int16` is classified as unsigned and skips the new mask. For `abi.encodePacked(uint8(0), Signed16.wrap(int16(-1)))`, the coalesced word retains the sign extension and the first packed byte is `0xff` instead of `0x00`. The added tests only exercise primitive signed integer arguments, so they do not catch this supported case. Please derive signedness from the underlying UDVT type and add a runtime regression.

crates/codegen/src/lower/abi_packed.rs
```diff
@@ -214,8 +214,9 @@ impl<'gcx> Lowerer<'gcx> {
 
             let size = self.get_packed_size_from_expr(arg, ty);
             let left_aligned = self.expr_is_fixed_bytes(arg);
+            let signed = ty.is_signed();
```
mattsse: `Ty::is_signed()` [doesn't unwrap `TyKind::Udvt`](https://github.com/paradigmxyz/solar/blob/f8bb992ab4690506b1582b653ae19b55198048eb/crates/sema/src/ty/ty.rs#L576-L581), so signed UDVTs skip the mask. `abi.encodePacked(uint8(0), Signed16.wrap(int16(-1)))` still starts with `ff`, not `00`.

```suggestion
let signed = matches!(self.lower_type_from_ty(ty), crate::mir::MirType::Int(_));
```

mattsse (changes requested):
One remaining correctness issue.

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1152, opened 2026-08-06
feat(lsp): add import resolution

0xKarl98:
Towards OSS-515, stacked on #915

Use workspace-specific compiler settings to resolve Solidity imports for completion and definition navigation. Track Foundry configuration, remappings, include paths, nested and external workspaces, and file-operation changes so import results stay correct as projects evolve.

## Manual verification

The scenarios below were checked in a local VS Code Extension Development Host running the `solar lsp` binary from this branch. Definition navigation was also checked for relative, remapped, nested, and external-project imports.

### Relative import completion

A relative `./` import offers sibling files and nested directories from the current Foundry project.

<p align="center">
  Relative import completion suggestions
</p>

### Nested Foundry remappings

The nested Foundry workspace uses the `pkg/` remapping to resolve the inner `lib/inner` target instead of the ancestor `lib/outer` target.

<p align="center">
  Nested Foundry remapping completion
</p>

### Remapping changes are live

With `live/=lib/old/`, completion offers `live/OldOnly.sol`.

<p align="center">
  Completion before remapping update
</p>

After changing `remappings.txt` to `live/=lib/new/` while the host remains open, completion refreshes to `live/NewOnly.sol` without a restart.

<p align="center">
  Completion after remapping update
</p>

### Benchmark coverage

The path-index benchmark covers the import-resolution hot path with 16 nested workspaces and 1024 queries, plus isolated single ownership/overlay and base-path containment queries. The Criterion IDs are `16-workspaces-1024-queries`, `16-workspaces-single-query`, and `16-workspaces-containment-query`, so CodSpeed reports distinguish the workloads without changing the measured code paths.


mattsse (changes requested):
P2 — crates/lsp/src/file_operations.rs:535-539: The directory replay guard can discard a genuine later file creation. If didCreateFiles records an empty folder, no descendant watcher echo exists to consume the root guard. The first later watcher-only creation beneath that folder, such as a terminal-generated Solidity file or foundry.toml, matches the retained root and is ignored, so workspace discovery and analysis stay stale until another event. Please expire or otherwise distinguish these guards and cover the delayed-creation case.

crates/lsp/src/import_resolution.rs
```diff
@@ -0,0 +1,511 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ImportCursor<'a> {
+    path_prefix: &'a str,
+    complete_path: Option<&'a str>,
+    replacement_range: ByteRange<usize>,
+    delimiter: u8,
+}
+
+impl<'a> ImportCursor<'a> {
```
DaniPopes: why do we need this? and not just get the ast note for the import

crates/lsp/src/global_state.rs
```diff
@@ -1166,8 +1196,25 @@ impl GlobalStateSnapshot {
         let source_map = SourceMap::empty();
 
         for (path, contents, version) in vfs_files {
```
DaniPopes, line 1199: please avoid cloning the contents of every file a million times. you can store contents as Arc<String> in the vfs directly, which you can also passthrough to solar source map without cloning the raw contents

crates/lsp/src/document_links.rs
```diff
@@ -151,6 +151,26 @@ fn import_path_bytes(path: &Path) -> Cow<'_, [u8]> {
     }
 }
 
+pub(crate) fn solidity_string_contents(bytes: &[u8], delimiter: u8) -> String {
```
DaniPopes: -> impl Display, and use write! etc lazy formatting in callsites

crates/lsp/src/document_links.rs
```diff
@@ -151,6 +151,26 @@ fn import_path_bytes(path: &Path) -> Cow<'_, [u8]> {
...
+    debug_assert!(matches!(delimiter, b'\'' | b'"'));
+    const HEX: &[u8; 16] = b"0123456789ABCDEF";
+
+    let mut contents = String::with_capacity(bytes.len());
+    for &byte in bytes {
+        let is_safe =
+            byte == b' ' || (byte.is_ascii_graphic() && byte != delimiter && byte != b'\\');
```
DaniPopes: can reverse escaping too, such as `\`->`\\`, `\n` etc. look at what the lexer accepts and reverse the process

crates/interface/src/source_map/file_resolver.rs
```diff
@@ -204,26 +209,84 @@ impl<'a> FileResolver<'a> {
         }
     }
 
+    /// Returns the normalized filesystem paths that may satisfy `path`, in resolution order.
+    ///
+    /// This does not perform I/O or consult the source map. A returned path may remain relative if
+    /// the resolver has no current directory. Use [`Self::resolve_file`] for exact resolution,
+    /// including preloaded source-unit names and ambiguity detection.
+    pub fn candidate_paths(&self, path: &Path, parent: Option<&Path>) -> Vec<PathBuf> {
```
DaniPopes, line 226: try to dedup with resolve_file, perhaps with an inner function that has a closure to handle the early-returns / io

crates/lsp/src/import_resolution.rs
DaniPopes, line 1: presumably some of this exists already in foundry / foundry-core so i would also check just in case there

DaniPopes (changes requested):

crates/lsp/src/import_resolution.rs
0xKarl98: You mean AST node for the import ? If so , we can live with that

crates/interface/src/source_map/file.rs
```diff
@@ -245,6 +245,28 @@ impl SourceFile {
         })
     }
 
+    /// Creates a new `SourceFile` from shared source text.
+    pub(crate) fn new_shared(
```
DaniPopes: plz make `fn new` `Arc<String>` directly and the shrink_to_fit into a "if arc is unique" branch

crates/lsp/src/import_resolution.rs
DaniPopes: yes

mattsse (approved):

PR #1156, opened 2026-08-07
perf(codegen): compact interprocedural codegen

djolertrk:
This compacts internal calls across MIR and EVM IR. It prunes transitively unused internal arguments and discarded one-word results while preserving ABI wrappers, folds equivalent functions including direct self-recursive bodies, and calibrates single-use inlining using optimizer-run-weighted runtime and deployment costs.

The backend also regenerates repeated block-local expressions and identical pushes, preserves profitable values across static calls and control-flow joins, coalesces disjoint constant word copies into MCOPY on Cancun and later, and outlines repeated machine runs with live stack inputs and outputs under -Osize.


tests/ui/codegen/lowering/stack-too-deep/call_args.stdout
```diff
@@ -286,82 +286,10 @@ bb4:
...
-  mstore
-  push 704
-  mload
-  push 0x560
-  mstore
-  push 736
-  mload
-  push 0x580
-  mstore
+  mcopy
```
DaniPopes, line 292: we might also have lowering that generates these long copy chains, can you check if we can't emit mcopy directly?

crates/codegen/src/transform/function_compaction.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+/// Computes the least fixed point of argument liveness and removes every argument outside it.
+fn prune_unused_args(module: &mut Module) -> usize {
+    let mut called = DenseBitSet::new_empty(module.functions.len());
+    let mut live = module
+        .functions
+        .iter()
+        .map(|func| vec![false; func.params.len()])
+        .collect::<IndexVec<FunctionId, _>>();
```
DaniPopes, line 187: bitsets

crates/codegen/src/transform/function_compaction.rs
```diff
@@ -0,0 +1,740 @@
...
+
+fn has_rewritable_signature(func: &Function, is_called: bool) -> bool {
+    is_called
+        && !func.params.is_empty()
+        && func.params.len() == func.arg_indices().count()
+        && is_internal_body(func)
+}
+
+/// Computes the least fixed point of argument liveness and removes every argument outside it.
+fn prune_unused_args(module: &mut Module) -> usize {
```
DaniPopes, line 181: can this be one pass for both args and returns like llvm deadargelim?

crates/codegen/src/backend/evm/ir/passes/coalesce_copies.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    fn run_pass(&self, gcx: Gcx<'_>, module: &mut Module) -> bool {
+        coalesce_copies(gcx, module)
+    }
+}
+
+const WORD_BYTES: usize = 32;
+const COPY_INSTRUCTIONS: usize = 4;
+
+fn coalesce_copies(gcx: Gcx<'_>, module: &mut Module) -> bool {
```
DaniPopes, line 32: would this be easier to do in mir?

crates/codegen/src/backend/evm/ir/passes/block_cse.rs
```diff
@@ -0,0 +1,482 @@
...
+    id
+}
+
+fn fresh(next_expr: &mut usize) -> usize {
+    let id = *next_expr;
+    *next_expr += 1;
+    id
+}
+
+const fn expression_inputs(
```
DaniPopes, line 442: a lot of hardcoded properties in this file, please at least extract to op.rs in individual functions; opcode inputs is already present

crates/codegen/src/backend/evm/ir/passes/block_cse.rs
```diff
@@ -0,0 +1,482 @@
...
+
+fn ensure_depth(stack: &mut Vec<StackValue>, depth: usize, next_expr: &mut usize) {
+    let missing = depth.saturating_sub(stack.len());
+    if missing != 0 {
+        stack.splice(0..0, (0..missing).map(|_| StackValue { expr: fresh(next_expr), span: None }));
+    }
+}
+
+fn intern(expr: Expr, expressions: &mut FxHashMap<Expr, usize>, next_expr: &mut usize) -> usize {
+    if let Some(&id) = expressions.get(&expr) {
```
DaniPopes: .entry().or_insert_with

crates/codegen/src/backend/evm/ir/passes/block_cse.rs
```diff
@@ -0,0 +1,482 @@
...
+}
+
+fn push_fingerprint(opcode: u8, encoding: u8, value: PushValue) -> u64 {
+    let mut hasher = FxHasher::default();
+    (opcode, encoding, value).hash(&mut hasher);
+    hasher.finish()
+}
+
+fn operation_fingerprint(opcode: u8, epoch: Option<u64>, operands: &[u64]) -> u64 {
+    let mut hash = 0x9e37_79b9_7f4a_7c15u64 ^ u64::from(opcode);
```
DaniPopes: this is weird, please use fxhasher

DaniPopes (commented):
i would recommend getting into the habit of running a check over the entire pr for adherence to the agents.md rules, since they are easily missed, like `vec![bool] -> bitset`, `hashmap<index, _> -> indexvec<_>`

DaniPopes:
cyclops audit

djolertrk:
> i would recommend getting into the habit of running a check over the entire pr for adherence to the agents.md rules, since they are easily missed, like `vec![bool] -> bitset`, `hashmap<index, _> -> indexvec<_>`

Hmmm, I agree. I am wondering if we can automate this somehow.

crates/codegen/src/backend/evm/ir/passes/coalesce_copies.rs
djolertrk: Hmm, actually no. The copies this fuses are backend marshaling of spilled call operands. they don't exist as MIR level (the MIR just has internal_call @f, args…), so a MIR pass would have nothing to coalesce. The copies that are explicit in MIR already emit `MCOPY` at lowering.

tests/ui/codegen/lowering/stack-too-deep/call_args.stdout
djolertrk: these particular chains are backend marshaling of stack-too-deep call operands that got spilled to memory, not lowering, so there's no MIR-level copy to turn into an mcopy. Where lowering does emit contiguous copies, we already emit `MCOPY` directly. and, the `coalesce_copies` is what turns the marshaling spills into the single `mcopy` you see here.

crates/codegen/src/transform/function_compaction.rs
djolertrk: yes! nice catch!

crates/codegen/src/backend/evm/ir/passes/block_cse.rs
djolertrk: it makes sense

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -5362,8 +5968,11 @@ impl<'gcx> EvmCodegen<'gcx> {
             self.scheduler.instruction_executed(2, None);
         }
 
-        let retention_plan =
-            stack_mask.as_ref().and_then(|mask| self.plan_retained_stack_args(func, args, mask));
+        let retention_plan = (!carries_resident_stack)
```
djolertrk: Looks like a false positive. That branch only fires for a computed arg with no reloadable spill, but call args are already spilled by `spill_top_value_if_live` first, so it's a no-op (zero hit-count across the corpus), and the `debug_assert_eq!(stack.depth(), plan.retained.count())` after the drain holds.

crates/codegen/src/backend/evm/codegen.rs
```diff
@@ -4226,73 +4491,332 @@ impl<'gcx> EvmCodegen<'gcx> {
...
+                        continue;
+                    };
+                    let Some(mask) = candidates.get_mut(function) else { continue };
+                    seen.insert(*function);
+                    if args.len() != mask.domain_size() {
+                        excluded.insert(*function);
+                        continue;
+                    }
+                    for (index, &arg) in args.iter().enumerate() {
+                        if !Self::raw_arg_emittable(caller, raw_leaves_ok, arg)
```
djolertrk: false positive

djolertrk:
> > i would recommend getting into the habit of running a check over the entire pr for adherence to the agents.md rules, since they are easily missed, like `vec![bool] -> bitset`, `hashmap<index, _> -> indexvec<_>`
> 
> Hmmm, I agree. I am wondering if we can automate this somehow.

I will try to come up with some advanced lints for this, but in a separate PR.

DaniPopes:
cyclops audit

djolertrk:
cyclops audit

djolertrk:
cyclops audit

DaniPopes (approved):

PR #1157, opened 2026-08-08
perf(codegen): aggregate checked add carries

malik672:
Checked `uint256` addition trees currently lower each addition with its own immediate overflow branch. Besides duplicating control flow, that keeps intermediate values live and can force expensive stack arrangements or spills for large expressions.

This change recognizes builtin checked `uint256` addition trees whose leaves are distinct resolved identifiers. It evaluates those leaves in source order, accumulates the modular sums and carry bits in a stack-friendly order, then emits one arithmetic panic branch for the combined carry condition. Signed and narrow integers, unchecked expressions, user-defined operators, side-effecting leaves, and repeated operands retain the existing lowering.

The largest affected UI fixtures shrink from 431 to 272 runtime bytes for `params` under `-Ogas`, from 907 to 673 bytes for `locals` under `-Osize`, and from 1,013 to 794 bytes for `call_args` under `-Ogas`. Deployment gas falls by as much as 47,371, while affected runtime calls improve by as much as 432 gas. Runtime coverage exercises left- and right-associated trees plus overflow panics, and the affected EVM IR fixtures pin the aggregate carry chain.

### DIFF
--- a/crates/codegen/src/lower/expr.rs
+++ b/crates/codegen/src/lower/expr.rs
@@ -47,6 +47,76 @@ pub(super) enum DecodeStrategy<'gcx> {
 }
 
 impl<'gcx> Lowerer<'gcx> {
+    /// Collects the leaves of a checked `uint256` addition tree.
+    ///
+    /// This deliberately accepts only identifiers: reassociating an addition tree is
+    /// semantically valid for unsigned values, but lowering all leaves before checking the
+    /// aggregate carry must not move side effects across an arithmetic panic.
+    fn collect_checked_uint256_add_leaves<'hir>(
+        &self,
+        expr: &'hir hir::Expr<'hir>,
+        leaves: &mut Vec<&'hir hir::Expr<'hir>>,
+    ) -> bool {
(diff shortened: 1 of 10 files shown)


malik672:
@codspeedbot explain why this is faster

DaniPopes:
this is interesting, could we do this for all the other checked ops too recursively? like any combination of + - / etc. . the only problem is that this only works thru syntax so codegen behaves differently for `a + b + c` vs `d = a + b; d + c`

DaniPopes:
cyclops audit fast

malik672:
> this is interesting, could we do this for all the other checked ops too recursively? like any combination of + - / etc. . the only problem is that this only works thru syntax so codegen behaves differently for `a + b + c` vs `d = a + b; d + c`



Not sure but I don't think so to be honest
If addition overflows, we can do a final check to catch it(can't semantically dissapear);with subtraction, multiplication and division we probably have to panic

Will fix the audit issue and explore the MIR versions seperately

PR #1159, opened 2026-08-09
feat(lsp): enforce protocol compatibility

0xKarl98:
Towards OSS-517.

Improve server-side LSP protocol compatibility using pinned capability subsets modeled on vscode-languageclient 10.1.0, Neovim 0.12.4 across its Darwin/Windows and Linux/BSD profiles, Zed 1.14.2, and a minimal LSP 3.17 client. Enforce lifecycle ordering: ordinary requests are rejected and ordinary notifications are dropped until a successful `initialize` response is followed by `initialized`. `$/cancelRequest` remains active so failed or cancelled initialization can be retried. Once a `shutdown` request is received, subsequent ordinary requests are rejected and ordinary notifications are ignored; `exit` terminates successfully only after a prior `shutdown` request.

Use a project-local lifecycle middleware as a compatibility shim because [async-lsp 0.2.4's `LifecycleLayer`](https://github.com/oxalica/async-lsp/blob/da6ef291e5d9b8a7508548a02cb0d7b78e0cf4b4/src/server.rs#L107-L129) does not yet enforce these [LSP 3.17 lifecycle rules](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize). The shim stays at the service boundary and can be removed once async-lsp provides equivalent behavior upstream.

Profile-driven raw JSON-RPC sessions cover initialization, UTF-16 positions, document updates, workspace reloads, watcher registration, diagnostics, and shutdown. These are protocol-level capability-profile tests and do not launch editor processes.


crates/lsp/src/lifecycle.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+impl<S> LspService for Lifecycle<S>
+where
+    S: LspService,
+    S::Error: From<ResponseError> + Send + 'static,
+    S::Response: Send + 'static,
+    S::Future: Send + 'static,
+{
+    fn notify(&mut self, notification: AnyNotification) -> ControlFlow<Result<()>> {
```
DaniPopes: shouldn't this be handled by the lsp router crate?

0xKarl98: The 0.2.4 implementation we use does not enforce the required LSP 3.17 semantics: its catch-all notification arm forwards ordinary notifications to the inner service, and its `exit` arm always returns `Break(Ok(()))` without checking whether `shutdown` was received ([async-lsp implementation](https://github.com/oxalica/async-lsp/blob/da6ef291e5d9b8a7508548a02cb0d7b78e0cf4b4/src/server.rs#L107-L129)). 

LSP 3.17 requires ordinary notifications before initialization to be dropped except `exit` ([initialize specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize)), and requires process status 1 when `exit` arrives without a preceding `shutdown` ([exit specification](https://github.com/microsoft/language-server-protocol/blob/626d5708af4af4c5be7205c7cad1f861708cad6d/_includes/messages/3.17/exit.md#L1-L8)).

DaniPopes: let's submit an issue upstream and leave a TODO?

mattsse (approved):
lgtm

crates/lsp/src/lifecycle.rs
0xKarl98: Have submitted :
https://github.com/oxalica/async-lsp/issues/28

PR #1161, opened 2026-08-09
refactor(codegen): rewrite MIR codegen

DaniPopes:
Rewrites codegen as progressive MIR lowering: ABI wrappers, dispatch, memory objects, allocation, and EVM-shaped calls are explicit MIR passes. The backend now schedules and emits only EVM-shaped MIR. This removes backend-owned Solidity cases, makes call and memory lowering testable in MIR, and lets shared helpers remove repeated ABI and memory sequences.

## Performance

The latest codegen benchmark comment compares `dfc5c9d` with `4a00d8e`; it predates the current branch head, so these figures describe the measured comparison rather than the final commit.

| area | result against main | why |
| --- | --- | --- |
| runtime bytecode | smaller for all 15 runtime workloads, from -2.60% to -27.92% | Explicit ABI, dispatch, and memory lowering exposes repeated sequences to helper sharing and MIR optimization. |
| runtime gas | 12 of 15 workloads improve, from -0.01% to -10.30%; three regress by +0.02%, +0.13%, and +0.19% | The new control-flow and validation shapes remove duplicate work in most paths; the small losses come from conservative checks and different branch layout. |
| deployment bytecode | smaller for 13 of 15 workloads, from -4.77% to -21.47%; two grow by +0.83% and +4.04% | Shared runtime lowering reduces most constructor and dispatcher duplication, but a few contracts retain more explicit setup code. |
| project compile time | 14 of 24 workloads improve, up to -40.69%; ten regress | Full compilation now runs the progressive MIR and backend-shaping passes for each emitted artifact. The largest measured regression is seaport-1.6, +55.50% (2.39 s to 3.72 s). |

The CodSpeed `Solarray/codegen` result, 25.2 ms to 209.4 ms, is not a like-for-like regression. Although the benchmark set `codegen_all_functions` on main, main returned after lowering or analysis before EVM emission. The rewritten path compiles Solarray through the backend, so it measures work that main skipped. The separate `solarray-a547630-project` full-compile benchmark is -4.94%.

The full-project losses are real comparisons. A `-Ztime-passes` run on Seaport shows repeated MIR and EVM pass pipelines across its many contracts and helper artifacts, including no-op passes; that is the likely source of the remaining cost, rather than parsing or semantic analysis. The final-head benchmark must confirm each loss before merge. Output quality remains the priority, and the branch should not merge with a confirmed output or runtime regression.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

djolertrk:
@DaniPopes can you please rebase this?

djolertrk:
cyclops audit

djolertrk:
can we add lang-ref doc, similar to https://llvm.org/docs/LangRef.html? we can start with MIR, and I can extend it with EVM IR later

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

crates/codegen/src/lower/function/calls.rs
```diff
@@ -0,0 +1,1314 @@
...
+            // result.length = length
+            let object =
+                self.builder.alloc_object(size, layout, AllocationSemantics::SOLIDITY_ZEROED);
+            self.builder.set_memory_object_len(object, len, layout.kind());
+            if let TyKind::DynArray(element) = ty.peel_refs().kind
+                && self.types.memory_layout(element).is_some()
+            {
+                // for i in 0..length { object[i] = default(element) }
+                self.counted_loop(len, |this, index| {
+                    let value = this.default_object(element)?;
```
mattsse: `memory_layout(element).is_some()` also matches `bytes`, `string`, and nested dynamic arrays, so this loop allocates a fresh empty object for every default element. Solidity uses the shared zero-slot pointer for those defaults. For example, memory-safe assembly comparing the two element pointers in `new bytes[](2)` returns true under solc 0.8.30 but false at this commit, and this also advances the free-memory pointer once per element. Please keep dynamic-reference defaults on the zero slot and only allocate aggregates that require distinct backing objects. The added regression covers structs but not this case.

mattsse (changes requested):
The dynamic-reference case still needs correction.

PR #1162, opened 2026-08-10
perf(codegen): use stack-native static ABI

djolertrk:
Plan non-recursive static calls with one entry and exit ABI that selects resident or frame-backed arguments, preserves profitable caller prefixes, and returns complete multiword tuples on the EVM stack. Retain the existing memory convention whenever the physical stack or control-flow proof is not safe.

Initialize the free-memory pointer lazily per external selector and place it above only the static frames reachable from that entry. Keep ordinary phi-source lifetimes so loop-carried values are not duplicated into stale stack copies.


DaniPopes:
can you merge the two stacks or is that not allowed by github

djolertrk:
> can you merge the two stacks or is that not allowed by github

unfortunately, this cannot be merged with another stack PR

DaniPopes:
cyclops audit

djolertrk:
cyclops audit

DaniPopes (approved):

PR #1165, opened 2026-08-10
perf(codegen): retain and reuse stack-resident values

djolertrk:
Cost resident static-call layouts with a lifetime model, retain computed values across planned joins without their early spill stores, fuse consecutive constant-size allocations into one free-memory-pointer bump, bind stack-returned results directly to their consumers instead of round-tripping through the 0x20 buffer, and teach block CSE a known-words model that drops redundant constant-address stores and forwards them to later loads.


DaniPopes:
cyclops audit

DaniPopes (approved):

PR #1168, opened 2026-08-11
perf(codegen): plan store and copy operands

malik672:
Currently whenever we encounter a store or copy instruction we spill into memory even when there's a cheaper of a cheaper arrangement
This change allows the Operand plan to optimize against memory and store spilling then falling back when there's no available path
On the codegen UI corpus, `-Ogas` runtime bytecode drops from 110,319 to 107,231 bytes and deployment bytecode drops from 122,363 to 118,949 bytes. Under `-Osize`, runtime bytecode drops from 111,434 to 109,329 bytes and deployment bytecode drops from 123,756 to 121,397 bytes.


DaniPopes:
cyclops audit fast

malik672:
Damn and I formally verified it
@DaniPopes can anyone call _cyclops audit ?

malik672:
A specialized planner or a wrap around the current planner specifically for Memory mutating

DaniPopes:
> @DaniPopes can anyone call _cyclops audit ?

No but feel free to ask for one

malik672:
created a new operand plan specifically for memory writing/copy
and added runtime regression for cyclops finding
quite similar but with increase compile runtime by 0.03s

malik672:
> > @DaniPopes can anyone call _cyclops audit ?
> 
> No but feel free to ask for one

Requesting one

djolertrk:
cyclops audit

malik672:
Disabled the plan copy operands for later

PR #1169, opened 2026-08-11
feat(cli): warn instead of -Zcodegen

DaniPopes:
Codegen now runs whenever bytecode or IR output is requested, so solc-compatible callers no longer need an unstable opt-in. The compiler emits warning 2264 to mark codegen as experimental. This removes `-Zcodegen` from the compiler configuration and all callers, tests, benchmarks, and fuzzers; Standard JSON follows the same behavior, and Foundry can invoke Solar directly.


djolertrk (approved):
LGTM

PR #1170, opened 2026-08-11
fix(lsp): complete unterminated imports

mattsse:
Import completion only accepted paths present in a successfully parsed source AST. Unterminated string literals therefore produced no candidates, which breaks editors that do not automatically close quotes.

This recovers plain-string import prefixes only for completion by detecting lexically incomplete literals, closing a truncated prefix, and validating it through the parser AST. Completed imports and import-definition lookup retain their existing behavior, and replacement edits remain confined to the incomplete path. Regression coverage includes later unrelated strings, single-quoted named imports, and false-positive guards.

### DIFF
--- a/crates/lsp/src/import_resolution.rs
+++ b/crates/lsp/src/import_resolution.rs
@@ -9,9 +9,12 @@ use solar_interface::{
     source_map::{FileName, FileResolver, SourceMap},
 };
 use solar_parse::{
-    Parser,
+    Cursor, Parser,
     ast::{self, StrKind},
-    lexer::unescape::try_parse_string_literal,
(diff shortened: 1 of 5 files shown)


crates/lsp/src/import_resolution.rs
```diff
@@ -36,6 +39,19 @@ pub(crate) fn import_path_at(source: &str, cursor: usize) -> Option<ImportPathAt
         return None;
```
DaniPopes, line 39: 1. AST should be enough
2. why cant we get the AST from analysis instead of reparsing?

DaniPopes:
cyclops audit

PR #1171, opened 2026-08-12
fix(lsp): harden import completion

0xKarl98:
This supersedes #1170 and includes @mattsse's original parser-validated recovery for unterminated import strings. It is meant to replace that PR, not stack on top of it.

The follow-up addresses all four actionable findings from the [Cyclops review](https://github.com/paradigmxyz/solar/pull/1170#pullrequestreview-4912538282). Recovered edits now replace the rest of the original same-line token without crossing an unescaped line break, so they neither leave suffix text behind nor delete later source. String unescaping now always consumes input or exits after a continuation, and the LSP byte/position helpers share one CR-aware index, preventing the form-feed hang and lone-CR range mismatch.

Complete imports and go-to-definition still use the normal AST path. When the lexer already proves the import string is open, completion skips the parse that is guaranteed to fail and validates only the recovered prefix.

### DIFF
--- a/crates/lsp/src/handlers/reqs.rs
+++ b/crates/lsp/src/handlers/reqs.rs
@@ -4,7 +4,10 @@ use crate::{
     document_links::solidity_string_contents,
     formatter::{self, FormatterError},
     global_state::GlobalState,
-    import_resolution::{ImportCandidateKind, ImportResolver, decode_import_path, import_path_at},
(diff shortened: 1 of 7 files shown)


mattsse (approved):
lgtm

DaniPopes:
cyclops audit fast

PR #1179, opened 2026-08-13
fix(codegen): clean integer conversions

figtracer:
PR #1161 can carry the source word through integer conversions without cleaning it to the destination type. A negative int8 converted through uint8 could therefore remain an all-ones word, changing later widening, comparisons, and bitwise operations.

Normalize only conversions that change the 256-bit representation: narrowing conversions and same- or narrower-width signedness changes. Ordinary widening remains zero-cost, and the MIR tests pin both behaviors. Runtime coverage exercises int8/int24 boundaries, signed and unsigned directions, internal use, comparisons, and the full-width boundary across all optimizer modes.

The Solc-vs-Solar differential harness found this on #1161 in a stateful campaign, then reduced it to pure valid calldata. Independent EVM replay now matches Solc; a 1,024-case generated corpus and the full Foundry integration suite also pass.

### DIFF
--- a/crates/codegen/src/backend/evm/codegen.rs
+++ b/crates/codegen/src/backend/evm/codegen.rs
@@ -2846,8 +2846,16 @@ impl<'gcx> EvmCodegen<'gcx> {
         };
         let cfg = CfgInfo::new(func);
         let reachable = cfg.reachable();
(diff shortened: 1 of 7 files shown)


mattsse (approved):
lgtm

PR #1181, opened 2026-08-13
fix(codegen): hash storage bytes

figtracer:
The lowering rewrite leaves mapping-derived `bytes` and `string` expressions as storage references until a consumer requests a concrete value. `keccak256` and `erc7201` instead treated the raw storage slot as a memory object, which could expand memory at that slot and revert. This materializes those arguments through the existing typed memory boundary before hashing them.

The Solc-vs-Solar symbolic differential engine found the bug at the 33-byte short-to-long storage boundary. Its saved witness reproduced in Foundry and on a fresh Anvil: Solc returned normally while the rewrite reverted. With this patch, the same bounded stateful campaign completes without a mismatch. The regression covers lengths 0, 31, 32, 33, and 65 across direct, mapping, nested, struct, local storage-reference, internal storage-parameter, string, and ERC-7201 paths.

This is based on the active lowering rewrite in #1161 because the regression is not present before that rewrite.


mattsse (approved):
lgtm

PR #1182, opened 2026-08-14
fix(codegen): materialize storage revert reasons

figtracer:
The lowering rewrite passed nonliteral `Error(string)` reasons directly to the memory ABI encoder after handling only calldata materialization. A mapping-derived `string storage` therefore remained a raw storage slot, which could be interpreted as a memory object and produce an empty out-of-gas revert instead of the expected reason. This lowers the expression through the existing typed memory boundary so storage-backed reasons are loaded before encoding.

The Solc-vs-Solar symbolic differential engine found this with a clean-state single-call property that stores `"x"` in a mapping and reverts with it. Foundry replay and a fresh Anvil independently confirmed the same no-input witness: Solc returned exact `Error("x")` data while Solar returned `0x`. The same campaign reports no mismatch after this fix.

The regression pins exact short and long mapping payloads, local and nested storage references, `require` with a struct field, and calldata behavior across none, gas, and size optimization. The full UI suite, codegen unit tests, clippy, and formatting pass. This is based on #1161 because the regression is specific to the lowering rewrite.


mattsse (approved):
lgtm

PR #1184, opened 2026-08-14
perf(codegen): aggregate checked adds in MIR

malik672:
Replaces #1157.

@DaniPopes the MIR optimization feels cleaner than the HIR


DaniPopes:
Superseded by #1506, which adapts checked-operation aggregation to the current MIR pipeline.

PR #1187, opened 2026-08-16
fix(codegen): reload reserved spill values

DaniPopes:
`emit_value_fresh` skipped reserved spill slots until they were marked stored. That caused it to rematerialize an expression containing a stale free-memory-pointer load when block layout emitted the use before the defining block. Load any runtime-reloadable spill first, preserving the value across memory-pointer updates. Add a regression UI test for dynamic revert-data encoding followed by a self-call.


mattsse (approved):
lgtm

PR #1188, opened 2026-08-16
refactor(bench): share project corpora

DaniPopes:
Move every vendored project archive to testdata/projects so Criterion, CodSpeed, and the codegen runtime harness share one source of truth. Remove the extracted OpenZeppelin runtime archive by slicing the canonical 5.6.1 archive at runtime while preserving the focused optimizer/IR profile and project remappings. Add the small runtime projects to the parser and lowering benches, enable codegen for the cases that fit CodSpeed, and keep unsupported or large projects out of instrumented codegen. The runtime benchmark now also compares full archived Foundry Standard JSON inputs with solc and Solar in a separate compile-time-only projects suite; focused cases retain deployment and runtime checks. Document archive ownership, provenance, and the runtime fixture split.


mattsse (approved):
lgtm

PR #1193, opened 2026-08-17
fix(codegen): fix issues detected with seaport codegen

djolertrk:
Fix lowering of `addr.codehash`, `addr.code`, and `addr.code.length`.


crates/codegen/src/lower/expr.rs
```diff
@@ -1450,6 +1468,30 @@ impl<'gcx> Lowerer<'gcx> {
 
     /// Lowers `type(Contract).creationCode` or `type(Contract).runtimeCode`.
     /// Returns a `bytes memory` pointer with layout: [length (32 bytes)][bytecode data...]
+    /// Lowers `addr.code` into a fresh `bytes memory` copy of the account's
+    /// runtime code.
```
DaniPopes, line 1470: comment's messed up

crates/codegen/src/contract.rs
```diff
@@ -347,6 +347,20 @@ fn generate_contract_bytecode(
         .collect();
     let mir = capture_mir.then(|| built_mir.unwrap_or(module));
 
+    // EIP-170 rejects runtime code above this size at deployment, so an
+    // oversized artifact compiles but can never reach a mainnet-rules chain.
+    const DEPLOYABLE_CODE_SIZE: usize = 24576;
```
DaniPopes: let's submit this separately, i believe there's new limits in the next fork amsterdam too https://eips.ethereum.org/EIPS/eip-7954

djolertrk: ah, yes, I agree

DaniPopes (approved):

PR #1196, opened 2026-08-18
fix(codegen): lower fixed bytes indexing

figtracer:
Fixed-size bytes indexing currently falls through memory-object lowering and emits `INVALID`. This lowers it directly with `BYTE` after the Solidity bounds check and preserves the canonical `bytes1` representation.

`solsymdiff` found and replayed the discrepancy on `bytes7` index zero: Solc returned the selected byte while Solar faulted. The regression covers `bytes1`, `bytes7`, and `bytes32` at both endpoint and out-of-bounds indices. This is stacked on #1161.


mattsse (approved):
lgtm

DaniPopes:
Fixed-bytes indexing and its runtime coverage are integrated in dani/rewrite-codegen-lowering. Closing as superseded.

PR #1197, opened 2026-08-18
fix(codegen): preserve slice revert data

figtracer:
Invalid calldata slice bounds revert with empty data in Solc and the current Solar lowering. The rewrite emitted panic `0x32`, changing behavior observable through low-level calls and `try/catch`. This restores the empty revert while preserving panic `0x32` for an indexed access outside an otherwise valid slice.

`solsymdiff` found and replayed the discrepancy through an ERC-4337-style calldata struct field. Independent replay confirmed the same result for both `bytes` and typed array slices, with and without optimization and via IR. This is stacked on #1161.


DaniPopes:
Slice revert handling, snapshots, and runtime coverage are integrated in dani/rewrite-codegen-lowering. Closing as superseded.

PR #1198, opened 2026-08-18
fix(codegen): preserve getter bounds reverts

figtracer:
Compiler-generated public array getters revert with empty data on invalid indices in Solc, while explicit array indexing uses panic `0x32`. Solar used the explicit-index panic for both. This tracks getter lowering so fixed, dynamic, and nested public array getters preserve Solc behavior without changing user-written indexing.

`solsymdiff` found and replayed the discrepancy at index three of a three-element public array. It reproduces on current main with optimization and via IR both enabled and disabled; independent Anvil replay confirmed the exact revert-data difference.

### DIFF
--- a/crates/codegen/src/lower/checked_arith.rs
+++ b/crates/codegen/src/lower/checked_arith.rs
@@ -903,11 +903,19 @@ impl<'gcx> Lowerer<'gcx> {
                 return;
             }
             let always = builder.imm_bool(true);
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes:
I would rather file this to solc; we're consistent, solc isn't

PR #1200, opened 2026-08-19
fix(codegen): fold wide literal expressions

figtracer:
This fixes integer-literal expressions whose intermediate value exceeds one EVM word while the final value still fits its Solidity type. Solidity evaluates literal expressions with arbitrary precision, but our constant evaluator stopped at 256 bits and codegen then emitted checked runtime arithmetic, making expressions such as `2**256 - 1` revert.

The evaluator now uses the parser's existing 4096-bit numeric bound, codegen folds only syntactic integer-literal expressions whose final value fits an EVM word, and destination-type range checks remain unchanged. The storage-layout diagnostics continue to reject final values outside `uint256`.

`solsymdiff` found this in Solidity's `integer/basic.sol` and `integer/uint.sol` semantic cases. Both now produce bounded agreement with solc. This is based on #1161 because the affected lowering path lives in that rewrite.

### DIFF
--- a/crates/codegen/src/lower/function.rs
+++ b/crates/codegen/src/lower/function.rs
@@ -399,6 +399,12 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
     }
 
     fn lower_expr(&mut self, expr: &hir::Expr<'_>) -> Option<ValueId> {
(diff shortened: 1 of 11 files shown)


DaniPopes:
this is intentional

PR #1201, opened 2026-08-19
fix(codegen): clean exponent operands

figtracer:
This fixes high-level exponentiation when a narrow operand carries dirty upper bits from inline assembly. The raw value previously reached `EXP` or the checked power loop without masking or sign extension, so Solidity's declared operand width was ignored and valid operations returned the wrong value or panicked.

Lowering now tracks assembly-derived MIR values through assignments and control-flow merges, then applies the existing scalar normalizer only when such a value reaches `Pow`. Ordinary `uint32 ** uint8` code remains byte-for-byte identical to the baseline at both `-O gas` and `-O size`.

`solsymdiff` found this independently in Solidity's `exponentiation/signed_base.sol` and `exponentiation/small_exp.sol` semantic cases. Both now produce bounded agreement with solc. This is based on #1161 because the affected lowering path lives in that rewrite.

### DIFF
--- a/crates/codegen/src/lower/function.rs
+++ b/crates/codegen/src/lower/function.rs
@@ -204,6 +204,7 @@ struct FunctionLowerer<'gcx, 'ctx> {
     builder: FunctionBuilder<'ctx>,
     types: types::TypeLowerer<'gcx>,
     values: FxHashMap<VariableId, ValueId>,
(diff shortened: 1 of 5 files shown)


DaniPopes:
Exponent operand cleanup and its runtime coverage are integrated in dani/rewrite-codegen-lowering at 825aebacb. Closing as superseded.

PR #1204, opened 2026-08-19
fix(codegen): preserve loop continue edges

figtracer:
## Summary

- lower a `do while` condition as the loop update so explicit `continue` statements still evaluate it
- retain conditional `continue` states as loop-header phi inputs
- add upstream-based runtime coverage for do-while, nested loops, condition side effects, and ordinary while-loop controls

## Differential finding

`solsymdiff` found the upstream Solidity test `statements/do_while_loop_continue.sol` disagreeing at the current #1161 head (`825aebac`). With identical Standard JSON input (`sha256: 581631fa2802ea449dd8e58c80ffb9d8bd4b1c5670258eb47277c8789d714c41`), Solc 0.8.35 returned `42` while Solar returned `0` for calldata `0x26121ff0`. Foundry replay and an independent fresh-Anvil call confirmed the same result.

The mismatch reproduced with optimization on/off, optimizer runs 1/200, via-IR on/off, Cancun/Osaka, and BFS/DFS. Solar MIR showed the explicit `continue` jumping directly to the loop header and bypassing the condition. A conditional `continue` also exposed a missing phi predecessor; this change preserves that edge as well.

After the fix, the upstream test and the added boundary cases reach bounded symbolic agreement.

## Tests

- `cargo nextest run -p solar-codegen` (220 passed)
- `FILECHECK=/opt/homebrew/opt/llvm/bin/FileCheck TESTER_MODE=ui cargo nextest run -p solar-compiler --test tests` (1,269 passed)
- `cargo clippy -p solar-codegen -p solar-compiler --all-targets`
- `cargo +nightly fmt --all -- --check`

### DIFF
--- a/crates/codegen/src/lower/function/control_flow.rs
+++ b/crates/codegen/src/lower/function/control_flow.rs
@@ -757,13 +757,19 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
 
     pub(super) fn lower_loop(
         &mut self,
-        block: hir::Block<'_>,
+        mut block: hir::Block<'_>,
         source: LoopSource<'_>,
     ) -> Option<()> {
         self.materialize_default_bindings();
         let update_stmt = match source {
             LoopSource::For { update } => update,
-            LoopSource::While | LoopSource::DoWhile => None,
+            LoopSource::While => None,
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

PR #1207, opened 2026-08-20
fix(codegen): lower Yul literals as words

figtracer:
Yul string and hex-string literals are word values, but direct assignments and local declarations currently route them through Solidity literal lowering. Solar therefore returns a memory-object pointer where Solc returns the literal word.

This reuses the existing Yul word-expression lowering at the two binding boundaries while leaving Solidity literal materialization unchanged. The regression covers full and short hex literals, local declarations, string literals, and a builtin expression.

solsymdiff found and concretely replayed the mismatch for standard input be398f62cf9cdf3c4d7a31fa2f923cc697fc9ec757590f690b49ef761cef124e. The same input reaches bounded agreement after the fix, including unoptimized non-via-IR, Cancun DFS, and optimizer-runs=1 configurations.


mattsse (approved):
lgtm

PR #1208, opened 2026-08-20
feat(codegen): support nested tuple assignments

figtracer:
Nested tuple-literal assignments are currently lowered as a flat tuple. Solar tries to lower the nested right-hand side as one unsupported tuple object, emits INVALID, and reverts on Solc's tuple_assign_multi_slot_grow semantic test.

This recursively evaluates tuple-literal leaves, then performs every store only after evaluation is complete. That supports nested assignments while preserving simultaneous assignment semantics; the regression covers both multi-slot memory values and a three-level nested swap.

solsymdiff found and concretely replayed the mismatch for standard input 92f065a09d383d00d715ca07be4f1b6892d1939a0178d2797538b62b94a13e5a. The identical input reaches bounded agreement after the fix, including unoptimized non-via-IR, Cancun DFS, and optimizer-runs=1 configurations.


figtracer:
unsure we want this unpacking behavior

PR #1209, opened 2026-08-20
feat(lsp): add embeddable launch API

0xKarl98:
This adds a simple solar_lsp::launch(LaunchConfig) API so tools like Foundry can run the language server directly instead of spawning the Solar CLI .


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1210, opened 2026-08-20
fix(codegen): clean assembly scalar operands

figtracer:
Inline assembly may leave upper bits dirty even when a variable has a narrow Solidity type. This cleans those values when built-in operators, type conversions, events, custom errors, and storage writes consume them, while preserving the raw word across same-typed internal and user-defined operator calls as solc does.

The symbolic differential runner from #1084 produced replay-confirmed mismatches for arithmetic, comparisons, increments, decrements, negation, widening conversions, enum validation, indexed event topics, and boolean storage encoding. Exact upstream operator-cleanup cases also caught an earlier over-broad read-boundary implementation. The final replay matrix reaches bounded agreement under none, gas, and size optimization.


figtracer:
fixed: exact upstream operator tests caught the over-broad read cleanup, so the follow-up now preserves same-typed internal arguments and only cleans built-in operands/conversions. full ui, codegen, clippy, and fmt are green locally. @DaniPopes @onbjerg re-review please.

mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1211, opened 2026-08-20
fix(parse): preserve continued string whitespace

figtracer:
The symbolic differential runner from #1084 found and replayed a mismatch in the upstream Solidity semantic test `strings/unicode_string.sol:g()`. Solc returned the expected 20-byte string `😃, 😭, and 😈`, while the compiler returned 19 bytes after deleting the space following an escaped newline.

This preserves spaces and tabs after escaped LF and CRLF continuations, removes the non-Solidity whitespace-skipping diagnostic, and covers ordinary and Unicode strings under both line endings. The identical source reaches bounded agreement with default settings and with optimization and via-IR disabled.

### DIFF
--- a/crates/parse/src/lexer/unescape/errors.rs
+++ b/crates/parse/src/lexer/unescape/errors.rs
@@ -10,6 +10,7 @@ pub enum EscapeError {
     InvalidEscape,
     /// Raw '\r' encountered.
     BareCarriageReturn,
+
     /// Numeric character escape is too short (e.g. '\x1').
     HexEscapeTooShort,
     /// Invalid character in numeric escape (e.g. '\xz1').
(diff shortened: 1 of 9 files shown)


DaniPopes:
plz point to main

crates/parse/src/lexer/unescape/errors.rs
```diff
@@ -10,22 +10,6 @@ pub enum EscapeError {
...
-    /// "this is \
-    ///  \
-    ///  also ok" == "this is also ok";
-    ///
-    /// "this is \
-    ///  
-    ///  not ok"; // error: cannot skip multiple lines
-    /// ```
-    CannotSkipMultipleLines,
-
```
DaniPopes, line 28: keep empty line

figtracer: kept the grouping blank line. i also retargeted this parser-only fix to main.

DaniPopes (approved):

PR #1213, opened 2026-08-20
feat(test): add external foundry testing infra

djolertrk:
The idea is to add testing infra for external foundry projects. The command for it is:

```bash
cargo tq foundry-external [name]
```

Since it takes quite some time (it runs the tests from the external project itself, with using "solar"), it won't be integrated into CI.
While testing some famous projects (`seaport`, `uniswap` etc.) it triggered some missing pieces in the codegen. This branch contains the bug fixes as well.


djolertrk:
Running the `% cargo tq foundry-external` will run some pre-defined ones, as if we ran:

```bash
cargo tq foundry-external solady
cargo tq foundry-external morpho-blue
cargo tq foundry-external solmate
cargo tq foundry-external seaport
cargo tq foundry-external openzeppelin-contracts
cargo tq foundry-external uniswap-v4-core
```

Now, all those pass.

djolertrk:
cyclops audit

djolertrk:
cyclops audit fast

crates/parse/src/parser/mod.rs
```diff
@@ -790,6 +790,15 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
         self.prev_token = std::mem::replace(&mut self.token, next);
         while let Some((is_doc, kind, symbol)) = self.token.comment() {
```
DaniPopes, line 795: let's split the frontend fixes into a separate pr and stack?

crates/codegen/src/backend/evm/ir/passes/legalize_shifts.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn is_enabled(&self, gcx: Gcx<'_>, _module: &Module) -> bool {
+        !gcx.sess.opts.evm_version.has_bitwise_shifting()
+    }
+
+    fn run_pass(&self, gcx: Gcx<'_>, module: &mut Module) -> bool {
+        legalize_shifts(gcx, module)
+    }
+}
+
+pub(in crate::backend::evm) fn legalize_shifts(gcx: Gcx<'_>, module: &mut Module) -> bool {
```
DaniPopes: is this what solc does as well? maybe for stuff like this we shouldn't even support it unless solc does

crates/parse/src/parser/mod.rs
djolertrk: I agree

crates/codegen/src/backend/evm/ir/passes/legalize_shifts.rs
djolertrk: well, yes, solc legalizes Solidity shift expressions before `Constantinople` using the same EXP-based forms. It rejects explicit Yul `shl/shr/sar`, though. I’ll retain legalization for compiler-generated shifts and add EVM-version validation for explicit assembly builtins.

DaniPopes: right, let's do that in a separate pr, like https://github.com/paradigmxyz/solar/pull/1252 but there's a lot more yul builtins than these i believe

djolertrk: https://github.com/paradigmxyz/solar/pull/1267

DaniPopes:
cyclops audit

DaniPopes (approved):
most of these fixes are already present in #1161, i'm going to merge this and clean it up in there

PR #1215, opened 2026-08-21
fix(codegen): bound static call returns

figtracer:
The symbolic differential runner from #1084 found a replay-confirmed mismatch when an external callee returned a different byte count than its static return type.

This bounds static external return copying by the actual returndata size while retaining the expected ABI buffer shape for short data. The regression covers both short and oversized returndata.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1216, opened 2026-08-21
fix(codegen): hash implicit string literals

figtracer:
The symbolic differential runner from #1084 found a replay-confirmed mismatch while hashing an implicitly typed string literal.

This sends implicit string literals through the existing bytes materialization path so hash and precompile calls receive the same representation as explicitly typed strings. The minimized literal remains as a focused lowering regression.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1218, opened 2026-08-21
fix(codegen): preserve values across phi edges

figtracer:
The symbolic differential runner from #1084 found a replay-confirmed mismatch on a do-while continue path whose phi input was overwritten on a critical edge.

This splits critical incoming edges before lowering phi values to EVM-shaped MIR so each predecessor keeps its own value. MIR and runtime regressions retain the minimized control-flow shape.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1219, opened 2026-08-21
fix(codegen): type contract creation options

figtracer:
The symbolic differential runner from #1084 found a replay-confirmed mismatch for a typed contract creation option.

This lowers creation value and salt expressions with their expected word types so narrow and fixed-bytes options keep the correct representation. The minimized deployment case remains as a runtime regression.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1222, opened 2026-08-21
fix(codegen): handle recursive storage aggregates

figtracer:
The symbolic differential runner from #1084 found replay-confirmed mismatches in packed fixed-array sizing, packed aggregate deletion, empty array pushes, and recursive storage struct copy and clear operations.

This uses storage representation sizes consistently, preserves neighboring packed values, and traverses recursive aggregates without leaving stale slots. Each independently minimized storage case remains as a runtime regression.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1223, opened 2026-08-21
fix(codegen): convert calldata bytes directly

figtracer:
The symbolic differential runner from #1084 found a replay-confirmed mismatch when converting calldata bytes to fixed bytes.

This converts calldata bytes and strings directly from their slice representation instead of materializing memory first and changing the converted value. The minimized conversion and its boundaries remain as runtime regressions.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1225, opened 2026-08-21
test(lsp): avoid watched-file test race

DaniPopes:
Cancel the background analysis before waiting for the discovery registration. This keeps the test's synthetic analysis output from racing with the worker it schedules.


mattsse (approved):
lgtm

PR #1229, opened 2026-08-22
fix(codegen): return values on Yul leave

figtracer:
solsymdiff found a replay-confirmed mismatch where an inline Yul function that executed leave returned zero while solc returned the assigned output. Bare returns now use the existing return finalization path when the current function has return bindings. Runtime coverage includes immediate and conditional leave, multiple return values, and fallthrough across optimizer modes. This targets #1161 and was developed with AI assistance.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1230, opened 2026-08-22
fix(codegen): apply storage layout base slots

figtracer:
solsymdiff found a replay-confirmed mismatch where contracts using layout at still generated storage accesses and .slot values from slot zero. The persistent storage cursor now starts at the already-validated layout expression, while transient and struct-relative layouts remain zero-based. Runtime coverage includes packing, inheritance, direct storage access, transient storage, and an ERC-7201 expression across optimizer modes. This targets #1161 and was developed with AI assistance.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1232, opened 2026-08-22
fix(codegen): preserve qualified modifiers

figtracer:
solsymdiff found a replay-confirmed mismatch where Base.modifier on a derived function dispatched to the derived override. Explicitly qualified modifier targets are now preserved, while unqualified modifiers retain virtual dispatch in both call-graph construction and lowering. Runtime coverage checks both paths across optimizer modes. This targets #1161 and was developed with AI assistance.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1234, opened 2026-08-22
fix(bench): add CodSpeed context

DaniPopes:
Give the Unifap LSP benchmarks operation-specific CodSpeed groups so the dashboard keeps their LSP context instead of grouping them under a bare `lsp` name.


mattsse (approved):
lgtm

PR #1236, opened 2026-08-22
fix(codegen): handle storage reference expressions

figtracer:
Storage-reference assignment expressions updated the binding but then resolved the surrounding expression through the stale access. This returns the newly bound access from storage expression lowering so chained member and index writes use the assigned target.

`solsymdiff` replayed struct and mapping discrepancies under both optimized via-IR and unoptimized legacy compilation. Both now reach bounded agreement.

Developed with AI assistance.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1238, opened 2026-08-22
fix(codegen): coerce explicit returns

figtracer:
A single explicit return was lowered without its declared destination type, so a narrow value dirtied by inline assembly could be widened without cleanup. Route this path through the existing typed-expression coercion used by assignments and calls.

`solsymdiff` concretely replayed a `uint8` local holding the raw assembly word `0xffff` escaping as `uint256(0xffff)`, while solc returned `0xff`; optimized via-IR and unoptimized legacy compilation now reach bounded agreement. The pinned 18-project corpus remains byte-for-byte unchanged in every size-reporting case.

Developed with AI assistance.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1239, opened 2026-08-22
fix(codegen): discard static item expressions

figtracer:
Standalone library, library-member, and type references have no runtime effect, but expression-statement lowering tried to materialize them and replaced the containing function with `INVALID` after reporting an unsupported expression. Discard only statically resolved item references while preserving normal lowering for value expressions.

`solsymdiff` found this independently in the upstream library enum, struct, and stray-value semantic fixtures. All three concretely replayed under optimized via-IR and unoptimized legacy compilation; all six comparisons now reach bounded agreement. The pinned 18-case codegen corpus remains byte-for-byte unchanged.

Developed with AI assistance.


mattsse (approved):
lgtm

PR #1240, opened 2026-08-22
fix(codegen): reject calls to empty code

figtracer:
Solidity high-level calls to an address without code must revert, but raw CALL and DELEGATECALL succeed against empty accounts. Void external calls previously accepted that empty success because no returndata validation followed. Check code existence for direct calls, external function pointers, and linked libraries while leaving low-level calls and deployed self-calls unchanged.

`solsymdiff` found this independently in the upstream `bubble.sol` and `called_contract_has_code.sol` semantic fixtures. Both mismatched with replay-confirmed counterexamples under optimized via-IR and unoptimized legacy compilation; all four comparisons now reach bounded agreement. The full UI suite, 1,438-test workspace suite, nightly clippy, and the 18-case codegen corpus pass. Corpus compile status and errors are unchanged, with a net runtime-size change of 65 bytes.

Developed with AI assistance.


tests/ui/codegen/lowering/fmp_reload_computed_operand.sol
```diff
@@ -18,7 +18,10 @@ contract FmpReloadComputedOperand {
     }
 
     function run() external {
-        vm.expectRevert(stdError.arithmeticError);
+        (bool success,) = address(vm).call(
+            abi.encodeCall(Vm.expectRevert, (stdError.arithmeticError))
+        );
+        require(success);
         this.callWithFmp(2 ** 256 - 1, 2 ** 256 - 1);
```
DaniPopes:
```suggestion
        this.callWithFmp(2 ** 256 - 1, 2 ** 256 - 1);
```

DaniPopes (approved):

PR #1243, opened 2026-08-23
feat(lsp): accept resolved Foundry configs

0xKarl98:
This addresses the [inherited-profile review concern raised in foundry-rs/foundry#16254](https://github.com/foundry-rs/foundry/pull/16254#discussion_r3836421804). Forwarding only the selected profile left the embedded LSP inconsistent with Forge because the lightweight `foundry.toml` loader does not resolve `extends`: Forge could resolve `custom-src` while the LSP continued indexing`default-src`.

This adds a `LaunchConfig` API for embedding hosts to provide effective Foundry workspace values keyed by an exact normalized workspace root, including source and flycheck roots, library/include paths, final remappings, and the EVM version. 

An exact-root snapshot takes precedence during workspace discovery and loading, while unmatched manifests retain the existing local`foundry.toml` and selected-profile fallback. Relative paths are resolved against the workspace root, and snapshots are reused for rediscovery throughout the LSP session. 

This keeps Foundry-specific resolution in the embedding host while the LSP consumes only the final values it needs. The corresponding embedding-side wiring is tracked in [foundry-rs/foundry#16254](https://github.com/foundry-rs/foundry/pull/16254).

cc @figtracer @DaniPopes @mattsse

### DIFF
--- a/crates/lsp/README.md
+++ b/crates/lsp/README.md
@@ -24,25 +24,29 @@ When the host has already resolved Foundry workspace configuration (including `e
 provide the effective values directly:
 
 ```rust,no_run
-# use std::path::PathBuf;
+# fn build_config() -> Result<(), Box<dyn std::error::Error>> {
 # use solar_config::EvmVersion;
+let workspace = std::env::current_dir()?.join("workspace");
 let config = solar_lsp::LaunchConfig::default().with_foundry_workspace_config(
-    solar_lsp::FoundryWorkspaceConfig::new("/workspace")
-        .with_source_roots([PathBuf::from("/workspace/src")])
-        .with_flycheck_source_roots([PathBuf::from("/workspace/src")])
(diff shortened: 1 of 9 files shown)


crates/lsp/README.md
```diff
@@ -20,6 +20,35 @@ An embedding host that already selected a Foundry profile can pass it through wi
 `with_selected_profile("custom")`; the language server uses that profile when discovering workspace
 sources and when running its automatically configured Forge lint checks.
 
+When the host has already resolved Foundry workspace configuration (including `extends`), it can
```
DaniPopes: this is not necessary, please remove the paragraph above also, and stop editing this file

0xKarl98: Have removed

mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1245, opened 2026-08-24
fix(codegen): preserve tuple write order

figtracer:
`solsymdiff` replayed Solidity's upstream tuple-order cases and found two concrete divergences: repeated tuple targets kept the wrong value, and side-effectful left-hand targets were resolved in the wrong order. Solidity evaluates the right-hand side first, resolves left-hand places from left to right, then commits writes from right to left.

This prepares tuple values and places in that order across literal, call, low-level-call, and storage-reference sources, then applies the writes in reverse. It also reloads materialized storage bytes before aliased writes so a stale copy cannot restore an earlier value. Both upstream cases reach bounded agreement in all eight optimized and unoptimized legacy and via-IR runs, with unchanged aggregate hot gas and a nine-byte reduction in one benchmark artifact.

### DIFF
--- a/crates/codegen/src/lower/function/lvalues.rs
+++ b/crates/codegen/src/lower/function/lvalues.rs
@@ -117,7 +117,7 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
         self.store_lvalue_place_with_source(place, value, None)
     }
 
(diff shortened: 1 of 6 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1246, opened 2026-08-24
fix(codegen): clean mapping keys

figtracer:
`solsymdiff` replayed Solidity's upstream mapping-cleanup case and found that a `uint16` value dirtied by inline assembly selected a different storage slot from the equivalent clean key. Solc returned `true`, while Solar returned `false` for calldata `0x2b63198c` in optimized and unoptimized legacy and via-IR builds.

This normalizes fixed-width scalar mapping keys through the existing cleanup path before deriving their storage slot. The candidate reaches bounded agreement with solc in all four configurations, and the codegen size and hot-gas benchmark outputs remain identical.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1248, opened 2026-08-24
fix(codegen): emit empty contract runtime

figtracer:
solsymdiff replayed Solidity’s assign_type_info.sol fixture and found that solc emits non-empty, rejecting runtime code for an empty concrete contract while Solar emitted empty bytes. The mismatch reproduced with calldata 0x73c60a6b across optimized and unoptimized legacy and via-IR configurations.

This reuses the existing ABI and dispatch default-revert path for truly empty and constructor-only source modules while preserving interfaces and raw non-entry MIR behavior. The candidate reached bounded agreement in all four compiler configurations; the full 24-target benchmark suite was structurally identical and had identical hot-path deployment and call gas. The bug also exists on current main, while this patch targets the rewritten lowering architecture.


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1249, opened 2026-08-24
fix(fuzz): honor discovery EVM version

figtracer:
`solsymdiff` first asks Solc to discover the transitive source set before it materializes the shared standard-json input. That discovery compile used the default EVM version instead of the requested campaign version, so pre-Cancun campaigns could reject otherwise valid source before reaching differential comparison.

Pass the requested `evmVersion` through discovery as well. The regression test uses `mcopy` as a normal Paris identifier and verifies the setting in both discovery and the final shared input.

Found while extending symbolic Solc-versus-Solar campaigns across EVM versions.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1250, opened 2026-08-24
fix(codegen): materialize stack-phi sources

figtracer:
Stack-phi planning rejected an unstored value even when the scheduler had deliberately chosen to recompute it. The fallback then emitted copies for a physical layout that was never present, causing a stack underflow in unoptimized loop functions.

Allow the same recomputation predicate already used by `emit_operand` during stack-phi preflight. `solsymdiff` changed the reproducer from incomplete in both unoptimized pipelines to bounded agreement in all four modes; concrete boundary replays pass, and paired size, serialized-output, and hot-gas benchmarks are unchanged.


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1251, opened 2026-08-24
fix(codegen): preserve ABI encode allocation

figtracer:
Source-level `abi.encode` returns an owned bytes object at the current free-memory pointer. The rewrite encoded a raw payload there and materialized an object afterward, so the object header could overwrite the payload and the observable pointer was wrong.

Distinguish raw-slice encodes from owned bytes encodes in MIR and lower the latter directly into `[length][data]` at the free-memory pointer. Calldata scalar-array validation now also terminates at the encode boundary, since there is no later materialization copy to carry that requirement.

`solsymdiff` changes a minimized reproducer from mismatch to bounded agreement in all four compiler modes, and the exact upstream fixture agrees in both optimized pipelines. Full-corpus and hot-gas results are unchanged except a 441-byte reduction in the signature-checker target.


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1252, opened 2026-08-24
fix(yul): honor Cancun builtin versions

figtracer:
This keeps Cancun-only Yul builtin calls version-aware. When targeting a pre-Cancun EVM, `mcopy`, `blobhash`, `blobbasefee`, `tload`, and `tstore` now fail during name resolution instead of reaching codegen; the old `mcopy`-only codegen special case is removed.

Solar intentionally reserves these Cancun-only Yul builtin names in every EVM version, unlike solc's pre-Cancun identifier behavior. That divergence is now documented rather than changing parser policy.

The full UI suite, workspace clippy, formatting checks, and all 1,457 workspace tests pass. The existing CodSpeed run reported all 174 benchmarks untouched.


crates/parse/src/parser/mod.rs
```diff
@@ -1005,7 +1006,12 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
     #[track_caller]
     fn parse_ident_common(&mut self, recover: bool) -> PResult<'sess, Ident> {
         let ident = self.ident_or_err(recover)?;
-        if ident.is_reserved(self.in_yul) {
+        let is_reserved = if self.in_yul {
```
DaniPopes: please revert and add it as a divergence in the doc, it's intentional we allow *parsing* of this stuff, and then fail later. perhaps we can fail earlier than codegen, but in codegen is fine as well

DaniPopes (changes requested):

crates/parse/src/parser/mod.rs
figtracer: makes sense

DaniPopes (approved):

PR #1253, opened 2026-08-24
fix(codegen): clean memory bytes conversions

figtracer:
## summary

- mask bytes beyond the semantic length when converting `bytes memory` or `string memory` to fixed bytes
- handle memory objects, memory slices, and calldata slices through the same conversion path
- reuse the existing dynamic-bytes type classifier instead of keeping a calldata-only duplicate

## proof

Solsymdiff found the upstream `bytes_to_fixed_bytes_cleanup.sol` case: after inline assembly shortens a memory byte array, Solar leaked dirty tail bytes while solc cleared them. The baseline mismatched in all four optimizer/backend combinations; this change reaches bounded agreement in all four. The standard codegen matrix regression covers lengths below, at, and above the fixed-width boundary.

## validation

- full UI suite passed
- workspace: 1453 tests passed; the load-sensitive LSP timeout passed alone, and the Foundry integration passed directly in 166s
- workspace clippy with `-D warnings`, check, build, fmt, and source diff checks
- 24-project runtime corpus and hot-call gas unchanged
- measured correctness cost: OpenZeppelin Governor creation bytecode +26 bytes and deployment gas +552; deployed size and hot-call gas unchanged

### DIFF
--- a/crates/codegen/src/lower/function/abi_values.rs
+++ b/crates/codegen/src/lower/function/abi_values.rs
@@ -1347,29 +1347,6 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
         end_offset
     }
 
-    pub(super) fn is_calldata_dynamic_bytes_type(&self, ty: Ty<'gcx>) -> bool {
-        match ty.kind {
-            TyKind::Ref(inner, DataLocation::Calldata) => matches!(
-                inner.kind,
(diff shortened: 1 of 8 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1254, opened 2026-08-24
fix(codegen): lower fixed-bytes length

figtracer:
## summary

- lower fixed-bytes `.length` to the type's compile-time byte width
- preserve side effects when the receiver is an expression

## proof

Solsymdiff found the upstream `fixed_bytes_length_access.sol` case. Dani's rewrite branch rejected it in all four optimizer/backend combinations; this change reaches bounded agreement with solc in all four. The standard codegen matrix test covers `bytes32`, `bytes16`, storage `bytes1`, and a side-effecting `bytes4` receiver evaluated exactly once.

## validation

- focused codegen and full UI suites passed
- workspace passed; the load-sensitive Foundry wrapper passed separately with a 300-second timeout
- workspace clippy with `-D warnings`, check, and fmt passed
- 24-project runtime corpus, 175-call hot gas workload, and 258 common run-call outputs unchanged

### DIFF
--- a/crates/codegen/src/lower/function/expressions.rs
+++ b/crates/codegen/src/lower/function/expressions.rs
@@ -385,6 +385,20 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
                 self.context.gcx.event_selector(event_id).as_slice(),
             )));
         }
+        if builtin == Builtin::FixedBytesLength {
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1256, opened 2026-08-24
fix(sema): allow calldata string slices

figtracer:
## summary

- allow Solidity-compatible range slicing of `string calldata`
- keep direct `string[index]` access rejected

## proof

The differential campaign found Solar rejecting the upstream `string_concat_different_types.sol` slice case while solc 0.8.35 and 0.8.36 compile it. Concrete Solc/Solar replay now agrees exactly in all four optimizer/viaIR combinations across empty, valid, out-of-bounds, and maximum bounds. The symbolic run remains incomplete on dynamic string copying, so it is not used as proof here.

## validation

- focused UI, sema/codegen, and upstream Solidity target passed
- full workspace: 1,457 tests passed, one skipped
- clippy, nightly fmt, and diff checks passed
- focused analysis and warmed Solady benchmarks were neutral

### DIFF
--- a/crates/sema/src/typeck/checker/mod.rs
+++ b/crates/sema/src/typeck/checker/mod.rs
@@ -449,9 +449,7 @@ impl<'gcx> TypeChecker<'gcx> {
                 } else if !is_calldata_sliceable(ty) {
                     self.dcx().emit_err(expr.span, "can only slice dynamic calldata arrays");
                 }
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1257, opened 2026-08-24
fix(codegen): preserve isolated array pop access

figtracer:
## summary

- evaluate the receiver of an isolated storage `.pop` member expression
- preserve receiver side effects without mutating the array

## proof

The differential campaign found Solar rejecting Solidity's upstream isolated array and bytes `.pop` cases. The standard codegen matrix test proves that `values[index++].pop` evaluates the indexed receiver once while neither it nor `data.pop` removes an element. The candidate reaches bounded agreement with solc in all four optimizer/backend combinations.

## validation

- full UI suite passed
- workspace remainder passed; load-sensitive LSP tests and the 140-second Foundry wrapper passed separately
- workspace clippy with `-D warnings`, check, build, and fmt passed
- 24-project runtime corpus and hot-call gas workload unchanged
- source diff checks clean; the generated MIR snapshot retains the repository-required terminal blank line

### DIFF
--- a/crates/codegen/src/lower/function/expressions.rs
+++ b/crates/codegen/src/lower/function/expressions.rs
@@ -59,6 +59,10 @@ impl<'gcx, 'ctx> FunctionLowerer<'gcx, 'ctx> {
                 let receiver = self.lower_expr(receiver)?;
                 return Some(self.builder.balance(receiver));
             }
+            if builtin == Builtin::ArrayPop {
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1259, opened 2026-08-24
chore(deps): use Dependabot for Cargo updates

decofe:
This moves weekly Cargo dependency updates into Dependabot, matching Foundry's grouped `cargo-weekly` setup with a seven-day cooldown and five-PR limit. It removes the separate scheduled Cargo update workflow so dependency automation is configured in one place.

Prompted by: @DaniPopes


DaniPopes (approved):

mattsse (approved):
lgtm

PR #1262, opened 2026-08-24
feat: frontend fixes for external foundry tests

djolertrk:


crates/parse/src/parser/mod.rs
```diff
@@ -793,6 +793,15 @@ impl<'sess, 'ast, 'cb> Parser<'sess, 'ast, 'cb> {
         self.prev_token = std::mem::replace(&mut self.token, next);
         while let Some((is_doc, kind, symbol)) = self.token.comment() {
             if is_doc {
+                // Only the last documentation unit binds to the item, like
```
DaniPopes: i would instinctively rather not do this and document it as a divergence rather than doing this mega hack just for parity, wdyt? is this blocking tests because there's natspec hard errors?

DaniPopes:
cyclops audit

crates/parse/src/parser/mod.rs
djolertrk: well, wes, it blocks Solmate (https://github.com/transmissions11/solmate/blob/89365b880c4f3c786bdd453d4b8e8fe410344a69/src/utils/SignedWadMath.sol#L4-L9): detached `@author` tags get attached to the first free function and produce hard NatSpec errors. I agree the implementation is heavy, so I’ll simplify it, but documenting the divergence would leave Solmate uncompilable.

djolertrk:
cyclops audit

crates/parse/src/parser/mod.rs
DaniPopes: i made all natspec errors into warnings, so i'm going to revert natspec changes; we can revisit later

djolertrk: ok, it does make sense

djolertrk:
@DaniPopes so we do not want any from this?

DaniPopes:
it's all natspec so no

PR #1264, opened 2026-08-24
fix(sema): resolve overridden call targets

figtracer:
Valid external functions can be overridden by public functions while changing reference parameters from calldata to memory. Both declarations then match an unqualified internal call, but comparing their internal parameter types incorrectly leaves the call ambiguous.

This selects a target only when the validated override graph proves that it transitively overrides every other matching declaration. Unrelated overloads remain ambiguous.

The regression covers both the direct override and the internal call. `solsymdiff` found the upstream case, and the patched compiler reaches replay-confirmed bounded agreement with Solc 0.8.36 for dynamic-array lengths 0 through 3.

AI assistance: Codex helped investigate, implement, and validate this change.

### DIFF
--- a/crates/sema/src/ty/mod.rs
+++ b/crates/sema/src/ty/mod.rs
@@ -1339,6 +1339,18 @@ impl<'gcx> Gcx<'gcx> {
         self.virtual_function_target((contract, function))
     }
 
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1265, opened 2026-08-24
fix(sema): match solc fallback and receive overrides

figtracer:
Solar excluded nameless functions from the current contract's override signatures when checking inherited ambiguity. That omitted both `fallback()` and `receive()`, so explicit overrides could leave their inherited definitions in the ambiguity set and Solar rejected contracts accepted by solc.

Match solc's `OverrideChecker` by excluding only constructors. `OverrideSignature` already includes `FunctionKind`, so fallback and receive remain distinct despite both having no name.

Validation:

- Solar UI tests exercise multi-base fallback and receive overrides at runtime.
- solc 0.8.36 (`8a079791`, the revision tracked by the Solidity submodule) compiles both fixtures and returns the same runtime results: fallback `(true, 0)` and receive `true`.
- The negative fallback fixture matches solc diagnostics 9456 (missing `override`) and 4334 (overriding a non-virtual function).
- `cargo +nightly fmt --all --check`
- `cargo +nightly clippy --package solar-sema --all-targets -- -D warnings`

AI assistance was used for implementation and validation.

### DIFF
--- a/crates/sema/src/typeck/override_checker.rs
+++ b/crates/sema/src/typeck/override_checker.rs
@@ -801,7 +801,7 @@ impl<'gcx> OverrideChecker<'gcx> {
             .functions()
             .filter_map(|f_id| {
                 let f = self.gcx.hir.function(f_id);
-                if f.kind.is_constructor() || f.kind.is_receive() {
+                if f.kind.is_constructor() {
                     return None;
                 }
                 Some(self.signature(OverrideProxy::Function(f_id)))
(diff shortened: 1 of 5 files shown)


DaniPopes:
@figtracer can you check these changes? ^

figtracer:
right, current changes make more sense than the previous commit since it also includes `receive()`, think it's good to merge as-is @DaniPopes

DaniPopes (approved):

PR #1267, opened 2026-08-24
fix(codegen): legalize legacy EVM shifts

djolertrk:
Preserve high-level Solidity shift semantics on pre-Constantinople targets without emitting unavailable opcodes. Rewrite SHL, SHR, and SAR in block EVM IR after the configured optimization pipeline, keeping the assembler primitive and custom pipelines safe.

Reject explicit Yul shift calls earlier through the versioned builtin availability check.


DaniPopes:
cyclops audit

DaniPopes (approved):

PR #1268, opened 2026-08-24
fix(codegen): validate typed catch payloads

figtracer:
Solar entered `catch Error(string)` after matching only the selector. A malformed ABI payload could then revert during string decoding instead of falling through to the generic catch or rethrowing the original bytes.

This adds the same non-reverting offset and length guards used by Solc before typed dispatch. The shared path covers external calls, creation, and external function pointers while leaving Panic and custom-error routing unchanged. `solsymdiff` replayed the original mismatch on main, #1161, and #1213; the patched compiler reaches bounded agreement, including the relevant boundary and noncanonical payload cases across the exercised optimizer modes. This is stacked on #1161.

AI assistance was used for implementation and validation.

The generated helper follows Solc’s [`try_decode_error_message`](https://github.com/ethereum/solidity/blob/develop/libsolidity/codegen/YulUtilFunctions.cpp#L4676-L4714).


mattsse (approved):
lgtm

PR #1269, opened 2026-08-24
fix(sema): preserve call-option members

figtracer:
## summary

- preserve call options when an external function value is followed by `.address` or `.selector`
- evaluate the callee and option expressions in Solidity source order before extracting the member
- keep standalone option-bearing function values outside this scoped fix

## validation

- focused UI tests, including side-effect order and diagnostics
- 241 solar-sema/solar-codegen nextest tests
- clippy with warnings denied and nightly fmt
- replay-confirmed `solsymdiff` bounded agreement with Solc across legacy and via-IR settings

Stacked on #1161.

Implementation and test development were assisted by AI; the final change was manually reviewed and validated.

### DIFF
--- a/crates/sema/src/hir/mod.rs
+++ b/crates/sema/src/hir/mod.rs
@@ -1746,6 +1746,12 @@ impl Expr<'_> {
                     arg.visit(f)?;
                 }
             }
+            ExprKind::CallOptions(callee, options) => {
+                callee.visit(f)?;
+                for arg in options.args {
+                    arg.value.visit(f)?;
+                }
+            }
             ExprKind::Delete(expr)
             | ExprKind::Member(expr, _)
             | ExprKind::Payable(expr)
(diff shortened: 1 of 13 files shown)


mattsse (approved):
lgtm

crates/sema/src/hir/mod.rs
```diff
@@ -1811,6 +1817,9 @@ pub enum ExprKind<'hir> {
     /// A function call expression: `foo(42)`, `foo({ bar: 42 })`, `foo{ gas: 100_000 }(42)`.
     Call(&'hir Expr<'hir>, CallArgs<'hir>, Option<&'hir CallOptions<'hir>>),
 
+    /// Call options on a function value: `foo{ gas: 100_000 }`.
+    CallOptions(&'hir Expr<'hir>, &'hir CallOptions<'hir>),
```
DaniPopes, line 1821: intentionally not handled as a separate node. it's lowered in Call and should stay there.

PR #1271, opened 2026-08-24
fix(codegen): lower contract type names

figtracer:
## summary

- lower `type(T).name` from the receiver semantic contract meta-type
- reuse the existing byte-literal path, including names longer than one word
- cover concrete, abstract, interface, library, parenthesized, and constant forms

## validation

- focused runtime UI coverage ported from the Solc semantic suite and extended for edge cases
- 221 solar-codegen nextest tests
- clippy with warnings denied and nightly fmt
- `solsymdiff` bounded agreement with Solc in optimized via-IR and unoptimized legacy modes

Stacked on #1161.

Implementation and test development were assisted by AI; the final change was manually reviewed and validated.


mattsse (approved):
lgtm

PR #1273, opened 2026-08-24
fix(codegen): gate selfbalance optimization

figtracer:
The MIR simplifier rewrites `address(this).balance` to `SELFBALANCE` regardless of the selected EVM version. On Petersburg, Solc emits `ADDRESS; BALANCE`, while Solar emitted an unavailable opcode and failed at runtime. This keeps `BALANCE` before Istanbul and preserves the optimization from Istanbul onward.

A Solc-versus-Solar EVM-version replay exposed the runtime discrepancy. The revisioned MIR fixture proves both sides of the fork boundary.

### DIFF
--- a/crates/codegen/src/transform/inst_simplify.rs
+++ b/crates/codegen/src/transform/inst_simplify.rs
@@ -20,6 +20,7 @@ use crate::{
     utils::eval,
 };
 use alloy_primitives::U256;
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1275, opened 2026-08-24
fix(sema): gate high-level EVM features

figtracer:
Solar currently accepts several high-level constructs whose opcodes are unavailable on the selected target, while Solc rejects them. This adds semantic availability checks for `address.codehash`, CREATE2 salt, `block.chainid`, `block.basefee`, `blobhash`, `block.blobbasefee`, and transient storage at their exact fork boundaries.

The parser remains intentionally permissive, following #1252; rejection happens during type checking. A Solc-versus-Solar version matrix found the family and the revisioned fixtures are attributed to the corresponding upstream Solc syntax cases.

### DIFF
--- a/crates/sema/src/builtins/mod.rs
+++ b/crates/sema/src/builtins/mod.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use solar_ast::StateMutability as SM;
 use solar_data_structures::map::FxHashMap;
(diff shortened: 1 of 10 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1279, opened 2026-08-25
fix(codegen): dispatch compatible casts

figtracer:
OpenZeppelin's `Arrays.sort` assembly-casts `address` and `bytes32` comparators to a `uint256` comparator. The lowering rewrite matched internal-function targets only by exact MIR signature, so it omitted the real target and reverted with panic `0x51` where Solc executed normally.

This admits only assembly-compatible full-word argument shapes (`uint256`, `int256`, `bytes32`, and `address`), cleans address words inside the selected case, and keeps return shapes exact. The focused fixture covers normal, dirty-address, and signed behavior under `-O none`, `-O gas`, and `-O size`; `solsymdiff` reaches bounded agreement for all three minimized targets.

The branch also carries the minimal warning updates required by the latest rolling nightly so the rewrite remains CI-clean.

Developed with AI assistance.

### DIFF
--- a/crates/codegen/src/lower/call.rs
+++ b/crates/codegen/src/lower/call.rs
@@ -9,7 +9,7 @@ use crate::{
     },
 };
 use alloy_primitives::{U256, keccak256};
-use solar_ast::{DataLocation, LitKind, Span};
+use solar_ast::{DataLocation, LitKind, Span, TypeSize};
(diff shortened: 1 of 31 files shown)


DaniPopes:
not merging to the branch?

figtracer:
> not merging to the branch?

yeah that's wrong sec

DaniPopes (approved):

PR #1288, opened 2026-08-25
feat(codegen): implement Amsterdam EVM version

DaniPopes:
Implements [EIP-8024](https://eips.ethereum.org/EIPS/eip-8024) as target-independent logical `dup N`, `swap N`, and `exchange N, M` EVM IR operations. Assembly keeps shallow DUP/SWAP encodings, lowers `exchange` to three swaps on older targets, and selects Amsterdam opcode-plus-immediate forms when available. Parsing, verification, optimization, assembly, disassembly, and codegen all gate target-specific behavior by EVM version.

The stack scheduler uses the target's full reachable window, emits logical exchanges directly, and spills only beyond that window. The EVM IR tail-merge size model stays target-neutral until assembly so Amsterdam's compact `EXCHANGE` does not disable profitable merges. This also implements [EIP-7843](https://eips.ethereum.org/EIPS/eip-7843) `block.slotnum` and Yul `slotnum()` through semantic analysis, MIR, EVM IR, LSP completion, and bytecode; `slotnum` is rematerializable and requires Amsterdam.

The benchmark runner can now replace every archived Standard JSON `evmVersion` and run without solc, so the comparison forces Osaka or Amsterdam over the same selected corpus. Against `main` at `f063cc94d`, the branch at `3fab35ab6` has no matched size regressions:

| Corpus | Runtime bytes | Creation bytes | Wins / losses |
|---|---:|---:|---:|
| UI Ogas, 257 files | -6,196 (-4.82%) | -6,203 (-4.39%) | 50 / 0 |
| UI Osize, 257 files | -5,221 (-4.03%) | -5,228 (-3.67%) | 33 / 0 |
| Forced runtime corpus, 15 contracts | -198 (-0.18%) | -200 (-0.17%) | 7 / 0 |

The 100-local spill fixture falls from 7,118 to 1,071 runtime bytes under Ogas and from 6,270 to 1,112 under Osize. The ordinary corpus emits 170 `EXCHANGE` instructions but no deep `DUPN` or `SWAPN`, so Amsterdam saves another 143 bytes against the same branch forced to Osaka; most contracts do not exercise deep-stack spill removal.

On forced Osaka, both main and the branch deploy all 15 mutually compiled contracts and pass the same 175 hot calls and 139 runtime reads. Hot-call gas is identical, while branch deployment gas falls by 11,960. Main and the branch compile the same 22 of 25 full-corpus cases; both retain the same three unrelated corpus failures. Amsterdam gas is not reported because the local Anvil build cannot execute EIP-8024 bytecode.


DaniPopes:
cyclops audit

mattsse (approved):
lgtm

DaniPopes:
cyclops audit

PR #1291, opened 2026-08-25
fix(codegen): align functions in memory

figtracer:
External function values use a low-aligned 192-bit scalar representation in MIR and storage, but Solidity memory stores the same `bytes24` value left-aligned. This keeps the scalar form internally and converts only at typed-memory boundaries, including aggregate construction, calldata and constructor materialization, ABI encoding, return cleanup, storage copies, and indexed event encoding.

`solsymdiff` found the issue in Solidity's upstream `abicoder/cleanup/function_v2.sol` fixture. For `C.ffff(uint256)` with `1`, Solc returned `(0, 0)` while Solar returned `(2**64, 2**64)` because dirty memory padding entered the live function value. The mismatch replayed under optimized/unoptimized and via-IR/legacy Solc settings; all four configurations now report bounded agreement.

The regression coverage exercises dirty scalar, struct, and array cleanup, packed encoding, memory/storage copies, constructor arrays, and the exact indexed-event topic. The full UI suite, 221 `solar-codegen` tests, the focused Foundry stress-events project, and nightly workspace clippy pass.

AI assistance: Codex assisted with implementation, differential reproduction, and validation.


mattsse (approved):
lgtm

PR #1300, opened 2026-08-25
fix(codegen): use source array layout

figtracer:
Storage assignments materialized calldata using the destination type. For a `uint256[4][]` storage target and `uint256[2][]` calldata source, that made ABI validation require 128 bytes per source element instead of the canonical 64 bytes, so Solar reverted before entering the function.

Materialize the source memory shape first, then let the existing storage conversion widen each fixed array and zero-fill the added elements.

`solsymdiff` found the divergence in Solidity's `array_copy_nested_array.sol`: the current rewrite mismatched Solc under optimized/unoptimized and via-IR/legacy settings. A minimized one-element stateful reproducer now reports bounded agreement in all four configurations.

Validation: full UI suite; 223 `solar-codegen` tests; four-mode `solsymdiff` replay; nightly workspace clippy; formatting and diff checks.

This targets the codegen-lowering rewrite.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1301, opened 2026-08-25
fix(codegen): preserve storage tuple aliases

figtracer:
Storage tuple lowering had two related aliasing bugs: it materialized storage aggregate sources before assignment targets were resolved, and nested tuple leaves assigned to local storage references were copied through their old referents instead of rebinding them.

This keeps captured storage locations through RHS evaluation, defers storage-to-storage copies until the right-to-left commit phase, and treats nested storage-reference leaves as reference assignments. Storage-to-memory copies remain eager.

solsymdiff replayed six storage-copy cases and two nested-reference cases against Solc under legacy and via-IR codegen, with and without optimization. The rewrite base mismatched all 32 combinations; these two commits report bounded agreement for all 32, including concrete replay of the observable storage and return values.

Validation covered the full UI suite, 223 solar-codegen tests, 1,463 ordinary workspace tests, the full Foundry comparison corpus, nightly workspace clippy, formatting, diff checks, and the complete GitHub CI matrix.

This targets the codegen-lowering rewrite.


mattsse (approved):
lgtm

DaniPopes:
cyclops audit

DaniPopes (approved):

PR #1302, opened 2026-08-25
fix(codegen): rebind nested storage refs

figtracer:
Nested tuple lowering treated storage-typed leaves as value copies even when their targets were local storage references. The assignment therefore wrote through each variable's old referent instead of rebinding it.

This preserves captured storage locations as reference assignments for both nested tuple literals and nested multi-return calls, then reuses the existing right-to-left commit path.

`solsymdiff` replayed both minimized forms under legacy/via-IR and optimized/unoptimized settings. The #1301 parent mismatched all eight combinations; this commit reports bounded agreement for all eight. A concrete replay returns `(10, 31, 41, 7)` with Solc instead of the parent's `(31, 41, 0, 7)`.

Validation covered the full UI suite, 1,463 ordinary workspace tests, the full Foundry comparison corpus, nightly clippy, formatting, and diff checks.

This depends on #1301 and targets the codegen-lowering rewrite. The new reviewable commit is `ac376db9`.

### DIFF
--- a/crates/codegen/src/lower/function/values.rs
+++ b/crates/codegen/src/lower/function/values.rs
@@ -3,8 +3,16 @@
 use super::*;
 
 enum PreparedTupleAssignment<'gcx> {
-    Value { place: LValuePlace<'gcx>, value: ValueId, source_ty: Option<Ty<'gcx>> },
+    Value { place: LValuePlace<'gcx>, rhs: TupleAssignmentRhs<'gcx> },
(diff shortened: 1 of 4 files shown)


DaniPopes:
first commit is #1301, should it be one pr?

mattsse (approved):
lgtm

figtracer:
yes, codex is starting to get retarded

figtracer:
retargeting this one and bundling this commit into #1301

PR #1304, opened 2026-08-25
feat(codegen): add shared program data

DaniPopes:
This adds first-class relocatable program data to MIR and EVM IR so lowering emits bounded data copies instead of long constant, add, and MSTORE chains. References carry stable IDs, offsets, and semantic display names. Contract creation and runtime bytecode use names such as `Child_initcode_0` and `Child_runtime_code_1`, while anonymous constants use `literal_N`. Repeated contract creation, `type(C).creationCode`, and `type(C).runtimeCode` share exact data or safe subslices across constructor and runtime artifacts, including unoptimized output.

The pack-data pass handles exact reuse, contained slices, repeated store runs, dead-data removal, and final source-order layout. It removes unreachable blocks before pooling, validates complete copy ranges, limits substring work, and prices later-container reuse against possible PUSH-width cascades. Zero data uses bulk clearing, size-optimized word splats use logarithmic MCOPY expansion, and shared single words can use CODECOPY when that reduces size. Existing data is packed before structural transforms, store runs are materialized after global code sharing, and a required final pass removes dead data. High-fanout groups remain available to global sharing when the local data-copy cost cannot price the full layout effect. Assembly resolves data addresses with labels in one fixed-point layout and inserts a terminal guard when appended data would otherwise be reachable by physical fallthrough.

The design follows LLVM and lld where their choices fit the EVM: hash exact values, establish liveness before layout, preserve stable survivor order, merge addressable pieces only when relocation costs stay bounded, and stop general overlap searches on large pools. Vyper and Sonatina comparisons also informed zero fill, word splats, explicit copy lengths, and post-layout relocation.

Against a fresh `origin/main` build, Solady LibString runtime is 71,098 bytes instead of 71,291 and SignatureChecker is 22,100 instead of 24,120. Against the prior PR build, those results improve by 552 and 60 bytes. The measured LibString hot calls improve, while the measured SignatureChecker hot calls are unchanged from the prior PR build.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1311, opened 2026-08-26
fix(codegen): clean dirty external ABI args

figtracer:
`solsymdiff` found a replay-confirmed mismatch in Solc's `bytesx_v2.sol` fixture: inline assembly can dirty a narrow scalar that still shares an external-argument `ValueId`, causing the compiler to forward the dirty word without ABI cleanup. Solc cleaned the `bytes1`; the nested external call otherwise reverted.

This keeps the external-argument fast path only while the value remains clean, so dirty scalars are normalized to their declared ABI type. It also revalidates dirty enum arguments before normalization, preserving Solc's `Panic(0x21)` for out-of-range values.

The replay reaches bounded agreement in optimized and unoptimized IR and legacy configurations for the original `bytes1` case and representative bool, unsigned integer, signed integer, address, and enum paths.

Developed with AI assistance.


DaniPopes (approved):

DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

PR #1312, opened 2026-08-26
fix(codegen): preserve ABI allocation panics

figtracer:
solsymdiff found a replay-confirmed discrepancy while decoding malformed dynamic external return data: Solc panicked with `Panic(0x41)`, while the compiler reverted without data.

This preserves Solc's checked-allocation ordering for non-alias memory-backed `bytes` and dynamic-array decodes. Calldata and alias paths keep their existing fast paths, and ordinary truncated return data still reverts without a panic.

The replay now reaches bounded agreement in optimized and unoptimized IR and legacy configurations.

Developed with AI assistance.


DaniPopes:
cyclops audit fast

DaniPopes (approved):

PR #1321, opened 2026-08-27
fix(codegen): preserve internal calldata returns

figtracer:
solsymdiff found two related ways internal calls could lose calldata slice metadata:

- optimized inlining could preserve the pointer but reload a stale zero length;
- internal-function-pointer dispatch and multi-return staging kept one word per logical return, dropping pointer/length pairs for secondary slices.

this keeps slices logical until calls that remain are lowered, then stages and reconstructs every physical return word. solver replay reaches bounded agreement with solc in all four optimizer/via-ir combinations.

validation: both focused runtime/mir matrices pass, all 225 codegen tests pass, and workspace clippy passes. the wider workspace reached 1457/1458 tests before the existing 120-second foundry integration timeout; the unrelated lsp timeout passed in isolation.

targets the codegen lowering rewrite in #1161.

generated with ai assistance.


DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1322, opened 2026-08-27
fix(codegen): validate projected calldata aggregates

figtracer:
solsymdiff found that projecting a narrow fixed calldata array and forwarding it to another external function reverted in solar while solc returned the encoded values.

the validator compared an abi head size in bytes with a static reference length whose unit is not bytes: fixed arrays carry element counts, while assembly-created static struct references may retain zero. this removes that mixed-unit comparison while keeping the physical calldata range check and recursive scalar canonicality validation.

the regression extends the upstream fixed-array indexing cases and covers a physically bounded static struct return. the 19-shape matrix agrees after the fix, including uint8/uint16/int8/bool/address/fixed-bytes/enums, nested arrays, dynamic outers, struct fields, and internal/external forwarding. the minimal case agrees in all four optimizer/via-ir combinations.

validation: both focused runtime/mir matrices pass, all 225 codegen tests pass, and workspace clippy passes. the wider workspace reached 1457/1458 tests before the existing 120-second foundry integration timeout.

this is a follow-up to #1212: that fix skips validation for the original external argument; this handles derived static calldata references. targets the codegen lowering rewrite in #1161.

generated with ai assistance.


DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1323, opened 2026-08-27
fix(codegen): encode nested calldata events

figtracer:
solsymdiff found a replay-confirmed divergence while emitting a static calldata struct with a nested fixed array. The compiler encoded the nested values as zeroes and consequently produced both incorrect unindexed log data and an incorrect indexed topic hash.

The selectorless static slice path used the fixed return-buffer encoder, which interprets nested calldata aggregates as memory objects. Routing it through the typed tuple encoder preserves the calldata source location. The Foundry regression compares both observable event paths, and the minimized reproducer reaches bounded agreement across optimized and unoptimized IR and legacy configurations.

Developed with AI assistance.


DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1326, opened 2026-08-28
feat(lsp): add cross-server benchmark

0xKarl98:
Towards OSS-521


0xKarl98:
Run /bench lsp here in case it doesn't affect normal tip-base branch benchmarking

.github/workflows/lsp-bench-comment.yml
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+      - name: Publish sticky benchmark comment
+        if: >-
+          steps.pr.outcome == 'success' &&
+          steps.artifact.outcome == 'success' &&
+          steps.download.outcome == 'success' &&
+          steps.report.outcome == 'success' &&
+          steps.recheck.outcome == 'success'
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
```
mattsse, line 361: This privileged job posts Markdown produced by code from the fork. The artifact checks establish which run supplied the file, but they do not authenticate its claims or constrain links and mentions, so a contributor can publish arbitrary bot-attributed content. Please transfer strictly shaped raw data and render it with trusted default-branch code here; alternatively, quote and neutralize the entire contributor-controlled body and label it as untrusted.

.github/workflows/lsp-bench-comment.yml
```diff
@@ -0,0 +1,203 @@
...
+              pull.head?.repo?.full_name !== headRepository ||
+              pull.base?.repo?.full_name !== repository
+            ) {
+              throw new Error(`Pull request identity changed for workflow run ${run.id}`);
+            }
+            core.setOutput("number", String(pull.number));
+            core.setOutput("head_sha", pull.head.sha);
+            core.setOutput("head_ref", pull.head.ref);
+            core.setOutput("head_repo", pull.head.repo.full_name);
+            core.setOutput("base_sha", pull.base.sha);
```
mattsse: `pull.base.sha` is captured only after the producer run has finished, while the producer benchmarked GitHub's merge checkout. If the base advances during the run, this records the new base and the later recheck still passes, publishing a stale merge result as current. Please have the producer record the exact tested base and merge SHA in authenticated metadata, then compare both with the current PR before posting.

tools/lsp-bench/src/process.rs
```diff
@@ -0,0 +1,2346 @@
...
+        let config = environment.path().join("config");
+        let data = environment.path().join("data");
+        #[cfg(target_os = "linux")]
+        let (cgroup, cgroup_procs) = match CgroupHandle::create_linux() {
+            Ok((cgroup, procs)) => (Some(cgroup), Some(procs)),
+            Err(_) => (None, None),
+        };
+        #[cfg(not(target_os = "linux"))]
+        let cgroup = None;
+        let mut command = server_command(spec);
```
mattsse, line 684: The child inherits the complete parent environment except for two removed variables. That exposes unrelated credentials, proxy settings, and agent sockets to every third-party LSP process, and the checked-in profiles do not enforce network isolation. Please call `env_clear()` and restore a minimal documented allowlist, with a canary test proving ambient secret variables are absent.

tools/lsp-bench/src/report.rs
```diff
@@ -0,0 +1,1230 @@
...
+                status,
+                metrics: metric_values
+                    .into_iter()
+                    .map(|(name, values)| (name, Stats::new(&values)))
+                    .collect(),
+            }
+        })
+        .collect();
+
+    let (harness_git_revision, harness_git_dirty) = harness_git_provenance();
```
mattsse: This provenance describes runtime checkout state, not necessarily the binary that is executing: Git runs in the caller's current directory, while the contract digest reads mutable source files after compilation. A reused binary can therefore claim another checkout's revision and source digest. Please embed build-time revision/tree/contract identity, include the executable digest, and reject mismatches; at minimum run Git with `-C CARGO_MANIFEST_DIR`.

tools/lsp-bench/src/process.rs
```diff
@@ -0,0 +1,2346 @@
...
+        let id = Value::from(self.next_id);
+        self.next_id += 1;
+        let key = id_key(&id)?;
+        let mut message = json!({"jsonrpc": "2.0", "id": id, "method": method});
+        if let Some(params) = params {
+            message["params"] = params;
+        }
+        let process_tree_cpu_started_ms = measured.then(|| self.process_tree_cpu_ms()).flatten();
+        let started_at = Instant::now();
+        self.send(message)?;
```
mattsse: `send` performs blocking `write_all`/`flush` before the deadline is created. A server that stops reading stdin can fill the pipe and hang forever, bypassing every request and shutdown timeout. Please move writes to a bounded writer/nonblocking path governed by the same operation deadline, and add a fake server that never reads stdin.

tools/lsp-bench/src/process.rs
```diff
@@ -0,0 +1,2346 @@
...
+        if let Some(cgroup) = &self.cgroup {
+            forced_kill |= cgroup.kill_and_wait(self.timeout)?;
+        }
+        #[cfg(target_os = "linux")]
+        let peak_process_tree_rss_mib =
+            self.rss_sampler.take().and_then(ProcessTreeRssSampler::finish);
+        #[cfg(not(target_os = "linux"))]
+        let peak_process_tree_rss_mib = None;
+        if let Some(thread) = self.stdout_thread.take() {
+            let _ = thread.join();
```
mattsse: These joins are unbounded. When cgroup setup is unavailable, a descendant can enter a new process group, survive group cleanup, retain stdout/stderr, and keep these threads waiting for EOF forever. Please use an enforceable process-tree primitive on every supported platform (or fail closed for authoritative runs) and deadline-bound output-thread teardown.

tools/lsp-bench/src/lifecycle.rs
```diff
@@ -0,0 +1,2037 @@
...
+    mut process: Command,
+    command: &Path,
+    timeout: Duration,
+) -> Result<String> {
+    let mut child =
+        process.spawn().with_context(|| format!("failed to run `{}`", command.display()))?;
+    let deadline = Instant::now() + timeout;
+    loop {
+        if let Some(status) = child.try_wait()? {
+            let output = child.wait_with_output()?;
```
mattsse: After the direct child exits, `wait_with_output()` can still block indefinitely if a descendant inherited stdout or stderr. The timeout branch also kills only the direct child. Please run version probes through the same process-tree containment and bounded pipe-draining primitive used for benchmark servers, with a test where the version command forks a pipe-holding child.

tools/lsp-bench/src/runner.rs
```diff
@@ -0,0 +1,3395 @@
...
+}
+
+fn harness_error(error: anyhow::Error) -> WorkloadError {
+    WorkloadError::new(FailureKind::HarnessError, format!("{error:#}"))
+}
+
+fn classify_request_error(error: anyhow::Error) -> WorkloadError {
+    if let Some(remote) = error.downcast_ref::<RemoteError>() {
+        let kind = match remote.code {
+            Some(-32601) => FailureKind::Unsupported,
```
mattsse: Mapping every `-32601` to `Unsupported` is incorrect after a capability check has established support, and it is especially wrong for mandatory `initialize`. Because unsupported samples are excluded from `failed_runs`, a protocol-inconsistent server can make a strict run succeed. Please classify method-not-found as a failure once the method was advertised, using `Unsupported` only when the capability gate itself says the method is absent.

tools/lsp-bench/src/runner.rs
```diff
@@ -0,0 +1,3395 @@
...
+    }
+    Ok(())
+}
+
+fn validate_rename_operations(
+    operations: &[WorkspaceEditOperation],
+    new_name: &str,
+    expected: &[(Url, Range)],
+) -> std::result::Result<(), WorkloadError> {
+    for (expected_uri, expected_range) in expected {
```
mattsse, line 2394: This only proves that every expected edit appears somewhere in the response. Additional text edits, renames, or deletes are accepted and then applied, so a destructive rename can still pass when no follow-up probe is configured. Please compare the complete normalized operation set with the declared oracle, or explicitly declare and validate any allowed extras.

tools/lsp-bench/src/runner.rs
```diff
@@ -0,0 +1,3395 @@
...
+        measured: bool,
+        allow_unopened_target: bool,
+    ) -> std::result::Result<(), WorkloadError> {
+        match probe {
+            ProbeSpec::Definition { path, anchor, expected_path, expected_anchor } => {
+                let encoding = self.position_encoding()?;
+                if !allow_unopened_target {
+                    self.require_open_for_probe(path)?;
+                }
+                let source_anchor =
```
mattsse, line 1457: Anchor lookup reads the fixture from disk, but an earlier unsaved `replace` updates only `Session.documents`. Any length-changing edit before this anchor makes subsequent request and expected-result positions stale. Please resolve positions against the current open-document text and add a multiline/non-ASCII unsaved-edit probe test.

tools/lsp-bench/src/runner.rs
```diff
@@ -0,0 +1,3395 @@
...
+            json!({"files": [{
+                "oldUri": old_uri,
+                "newUri": new_uri
+            }]}),
+        )?;
+        if let Some(parent) = new_path.parent() {
+            fs::create_dir_all(parent)?;
+        }
+        fs::rename(&old_path, &new_path)?;
+        if let Some(document) = self.documents.remove(&old_path) {
```
mattsse: Moving the harness's document entry without sending `didClose` for the old URI and `didOpen` for the new URI leaves the server's text-sync state behind. The WorkspaceEdit path already handles this correctly. Please share that lifecycle transition here (and send `didClose` before direct deletion), then test open -> rename/delete -> probe/save.

tools/lsp-bench/src/config.rs
```diff
@@ -0,0 +1,1407 @@
...
+        #[serde(default)]
+        probe: Option<ProbeSpec>,
+    },
+    Warm {
+        name: String,
+        probe: ProbeSpec,
+        #[serde(default)]
+        warmup: Option<usize>,
+        #[serde(default)]
+        samples: Option<usize>,
```
mattsse, line 244: A per-step `samples: 0` bypasses the profile-level nonzero checks, so `warm` executes no measured request and still returns success; with `warmup: 0` it never exercises the probe at all. Please reject zero-valued step overrides during config validation and cover both zero combinations.

tools/lsp-bench/src/runner.rs
```diff
@@ -0,0 +1,3395 @@
...
+        })
+}
+
+fn validate_workspace_symbol(
+    value: Value,
+    expected_name: &str,
+    expected_uri: &Url,
+    present: bool,
+) -> std::result::Result<(), WorkloadError> {
+    let Some(items) = value.as_array() else {
```
mattsse: The LSP result type permits `null` for an empty workspace-symbol response, but this rejects it. Conversely, a malformed array such as `[{}]` passes a negative probe because only two JSON pointers are inspected. Please deserialize the typed optional workspace-symbol response, accept `None` only for `present: false`, and reject malformed elements.

tools/lsp-bench/src/fixture.rs
```diff
@@ -0,0 +1,703 @@
...
+        bail!("{kind} artifact digest mismatch: expected {expected}, found {actual}")
+    }
+    Ok(())
+}
+
+fn copy_tree(root: &Path, source: &Path, destination: &Path) -> Result<()> {
+    for entry in fs::read_dir(source)? {
+        let entry = entry?;
+        let name = entry.file_name();
+        if name.to_str().is_some_and(|name| IGNORED_DIRECTORIES.contains(&name)) {
```
mattsse, line 498: An explicitly configured source root under `cache`, `out`, `target`, or another ignored name is inventoried and validated during `open`, then silently omitted from the materialized fixture here. That lets doctor pass on a fixture different from the one actually executed. Please either reject source roots containing ignored components or apply identical inclusion rules to inventory, hashing, and copying.

mattsse (changes requested):
The benchmark framework has several correctness and trust-boundary gaps that can make published results misleading or leave runs outside their advertised resource bounds. I left focused suggestions on the affected paths.

DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

PR #1329, opened 2026-08-29
fix(codegen): materialize array literal elements

figtracer:
`solsymdiff` found that memory array literals stored dynamic calldata elements as raw calldata slices. Later memory consumers interpreted their data offsets as memory-object pointers, so returning a `bytes[2]` literal could truncate an element or exhaust memory depending on optimization.

This materializes each element at the point where the memory array is constructed, matching the existing assignment and struct-construction boundaries. The minimized replay now agrees with solc 0.8.36 across optimized and unoptimized legacy/via-IR settings.

Generated with AI assistance.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1335, opened 2026-08-30
fix(ci): repair main workflow failures

decofe:
Fix the main CI failures and remove automatic release-binary builds from pull requests:

- Resolve benchmark PRs across all states so a merge racing the trusted `workflow_run` follow-up does not fail the comment job. The exact benchmark head SHA is still required.
- Make the restored-source flycheck test force a deterministic file revision change instead of relying on sub-second filesystem timestamp resolution. This is the test failing in both the MSRV and nightly matrix jobs.
- Keep release-binary builds available to the release workflow and manual dispatch, but stop running them automatically for pull requests. The workflow had no push trigger.

Validation:
- Parsed the modified workflow YAML with PyYAML
- Confirmed failed benchmark head SHAs resolve when querying all PR states
- `cargo +nightly fmt --all --check`
- Targeted LSP test on stable and nightly

Prompted by: @DaniPopes

### DIFF
--- a/.github/workflows/bench-comment.yml
+++ b/.github/workflows/bench-comment.yml
@@ -29,7 +29,7 @@ jobs:
             const head = `${run.head_repository.owner.login}:${run.head_branch}`;
             const { data: pulls } = await github.rest.pulls.list({
               ...context.repo,
-              state: "open",
+              state: "all",
               head,
             });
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1336, opened 2026-08-30
fix(codegen): clear packed array copy tails

figtracer:
A fixed-prestate `solsymdiff` campaign replayed Solidity's upstream `cleanup_during_multi_element_per_slot_copy` case and found that a packed dynamic-array copy preserved a value outside its logical length. Solc returned zero after the next `push()`, while the rewrite returned the stale value `4`.

This masks unused logical elements in the final partial storage word after packed array copies. Fixed and dynamic sources are covered, and exact-slot-boundary copies retain the existing next-slot behavior. The adapted regression and boundary control reach bounded agreement in optimized and unoptimized legacy and via-IR builds.

The minimized optimized runtime grows from 498 to 518 bytes, while the failing transaction drops from 47,490 to 40,572 gas. Packed-copy-heavy UI fixtures grow roughly 5–10%; the 15-case hot benchmark suite is unchanged in bytecode, deploy/call gas, and runtime results.

This targets #1161.

### DIFF
--- a/crates/codegen/src/lower/function/storage_values.rs
+++ b/crates/codegen/src/lower/function/storage_values.rs
@@ -83,19 +83,7 @@ fn load_packed_storage_array_element(
     // storage_slot = data_slot + index / per_slot
     // shift = (index % per_slot) * bytes * 8
     // value = decode(sload(storage_slot), shift, encoding)
-    let per_slot = u64::from(32 / bytes);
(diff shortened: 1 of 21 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1337, opened 2026-08-30
fix(codegen): support bytes push lvalues

figtracer:
`solsymdiff` surfaced Solidity's upstream `push_no_args_bytes.sol` case as incomplete: solc accepts `bytes.push() = value`, while the rewrite rejected the push call as an unsupported lvalue. Semantic analysis has supported empty storage-array push lvalues since #808, and codegen already handled the same form for ordinary dynamic arrays.

This lowers an empty storage-bytes push as a storage-byte place and shares the existing bytes-growth path. Ordinary assignments persist the grown value once. Tuple target preparation commits each growth left-to-right before the existing reverse-order writes, preserving aliasing and evaluation order without duplicating stores on the common path.

The exact upstream fixture now compiles. Short and long layouts, the 31-byte transition, and repeated tuple pushes agree with solc; a symbolic transition target reaches bounded agreement in optimized and unoptimized legacy and via-IR modes. The existing 15-project codegen corpus is unchanged in generated bytecode.

This targets #1161.


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1339, opened 2026-08-30
fix(codegen): bound fixed-array ABI encoding

figtracer:
Stacks on #1161.

That rewrite currently expands every fixed-array element into MIR before ABI encoding. `bool[989]` consequently emits a 110,006-byte optimized runtime that overflows the EVM stack, while dynamic-element arrays such as `string[512]` reach 100,736 bytes and fail the same way.

This keeps the measured small-array unrolled paths, then encodes larger static arrays with a bounded cursor loop and reuses the existing memory-array loop for dynamic elements. Nullable/default aggregates retain their canonical zero encoding.

The `bool[989]` runtime falls to 13,920 bytes and `string[512]` to 4,801 bytes; the pathological compile is 19.1x faster with 56.8% lower peak RSS. Runtime hashes match Solc across legacy/viaIR and optimized/unoptimized settings, while the 15-contract hot corpus remains byte-for-byte and gas-identical.

Developed with AI assistance.

### DIFF
--- a/crates/codegen/src/transform/lower_abi_encode.rs
+++ b/crates/codegen/src/transform/lower_abi_encode.rs
@@ -14,6 +14,17 @@ use solar_data_structures::map::{FxHashMap, FxHashSet};
 use solar_interface::{Ident, sym};
 use solar_sema::Gcx;
 
+// Unrolling large fixed arrays leaves at least one value live per element until it is encoded,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1343, opened 2026-08-31
fix(bench): disable bytecode metadata

DaniPopes:
Solc appends metadata to generated bytecode while the compiler does not yet emit a matching trailer. Set `bytecodeHash` to `none` and `appendCBOR` to `false` in every compressed project input so benchmark bytecode size comparisons measure the same payload.


mattsse (approved):
lgtm

PR #1344, opened 2026-08-31
feat(cli): add bytecode metadata

DaniPopes:
This adds solc-compatible Standard JSON metadata and CBOR runtime trailers. Metadata is computed lazily from the parsed compilation input and emitted as an opaque runtime suffix. This PR is stacked on #1304.


DaniPopes:
cyclops audit fast

PR #1350, opened 2026-09-02
perf(lsp): cache workspace path matches

0xKarl98:
Towards OSS-738 , 

This caches normalized workspace-root matches once per `WorkspacePathQuery`, allowing ownership selection and overlay fan-out to reuse the same traversal. Common match sets stay inline in a `SmallVec`. It also leaves disk-source string capacity intact until `SourceFile::new` performs the final shrink after indexing, avoiding a redundant reallocation, and covers non-normalized import paths.

On the 16-workspace Criterion workload, 1024 multi-projection queries improve from 22.9 ms to 7.34 ms, while one multi-projection query improves from 34.0 us to 19.3 us.

### DIFF
--- a/crates/lsp/src/global_state.rs
+++ b/crates/lsp/src/global_state.rs
@@ -2456,11 +2456,12 @@ impl AnalysisBatch {
         batch
     }
 
-    fn push_file(&mut self, path: PathBuf, mut contents: String) {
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1351, opened 2026-09-02
perf(lsp): cache derived document data

0xKarl98:
Cache derived data for exact-content VFS entries. Open-document selection-range requests now reuse the flattened source, UTF-16 position index, and parsed range candidates; analysis batches reuse the flattened source. Content changes replace the cache, while version-only updates and renames retain it.

On the Optimism `Criterion` workload (5.38 MB source), repeated open-document selection-range latency fell from 58.825 ms to 153.092 µs per request (99.74% lower, 384× faster). This is the cached open-document path; cold/direct selection-range latency is unchanged.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1352, opened 2026-09-02
feat(web): add benchmark viewer

DaniPopes:
Adds a GitHub Pages benchmark viewer for comparing committed benchmark runs and inspecting compiler artifacts from PR links. It publishes static benchmark data from completed workflow artifacts, supports local data and an optional ClickHouse-backed Worker, and keeps website checks separate from the compiler CI.

## Website preview

Benchmark comparison overview

Artifact diff viewer

### Available on GitHub Pages

- Direct benchmark links from PR comments for published Benchmark runs.
- Metric comparison, benchmark details, local command, source link, and static run history.
- Artifact viewer with `vs base` and `vs solc` for MIR, EVM IR, disassembly, hex, compiler input/output, and Yul artifacts.
- Syntax highlighting and light/dark themes.

### Not part of the static Pages release

- A global dashboard or all-run performance charts.
- The ClickHouse API, continuous GitHub backfill, and Worker deployment.
- On-demand results for commits whose workflow artifacts were not published.


.github/workflows/web-pages.yml
```diff
@@ -0,0 +1,203 @@
...
+        run: vp run worker:check
+      - name: Build website
+        working-directory: web
+        run: vp build
+
+  publish:
+    if: >-
+      (github.event_name == 'push' && github.ref == 'refs/heads/dani/perf-site') ||
+      (github.event.workflow_run.conclusion == 'success' &&
+       github.event.workflow_run.event == 'push' &&
```
mattsse, line 51: PR report links encode `BENCHMARK_PR_HEAD_SHA`, but this gate only allows Benchmark `workflow_run`s produced by a `push` to `main`. The production frontend fetches `data/runs/<head>/run.json` from `gh-pages`, so PR heads are never imported; the current head's path is 404 even after the successful benchmark and preview deployment. Please provide a trusted way to publish the PR run's data, or avoid emitting links that cannot load.

web/scripts/ingest-run.mjs
```diff
@@ -0,0 +1,134 @@
...
+  pr: options.pr ? Number(options.pr) : null,
+  title: options.title || null,
+  started_at: options.timestamp || new Date().toISOString(),
+  workflow_name: options.workflow || 'Benchmark',
+  source_schema: 1,
+  raw_results: JSON.stringify(document),
+}
+const results = normalizeResults(document, run)
+const artifacts = await normalizeArtifacts(resolve(options.artifacts), { ...run, results })
+await insert('runs', [run])
```
mattsse: `runs` is the completion marker used by `importedRuns()`, but it is inserted before the result and artifact rows. If either later ClickHouse insert fails after this succeeds, every ordinary rerun sees the workflow ID as already imported and skips it, leaving a permanently incomplete run unless someone knows to force `--refresh`. Please insert the marker last or track an explicit completed state.

web/schema/clickhouse.sql
```diff
@@ -0,0 +1,52 @@
+CREATE DATABASE IF NOT EXISTS solar_perf;
```
mattsse: `CLICKHOUSE_DATABASE` is passed as the current database for every request, but this schema hard-codes `solar_perf`. With a non-default configured database, schema creation and later unqualified inserts and queries target different databases; selecting a not-yet-created database can also make the initial `CREATE` fail. Please parameterize the schema for the configured database or remove the override.

mattsse (changes requested):
I found three actionable issues, detailed inline.

DaniPopes:
cyclops audit fast

PR #1362, opened 2026-09-04
perf(lsp): speed up benchmark reporting

0xKarl98:
Towards OSS-738 , 

The LSP benchmark renderer currently walks every ordered five-session bootstrap resample, even when different orders produce the same estimate. This groups those resamples by replacement count and keeps their exact multinomial weights. If `Decimal` precision cannot preserve the grouped arithmetic, it uses the existing exhaustive path.

Here are the results from the same 10-session, same-binary control artifact:

| Measurement | Before | After | Change |
| --- | ---: | ---: | ---: |
| Bootstrap states per confidence interval | 3,125 | 126 | 95.97% fewer |
| Bootstrap states per full seven-metric report | 87,500 | 3,528 | 95.97% fewer |
| Renderer wall time, observed locally | 0.24 s (1 run) | 0.08–0.13 s (5 runs) | 46–67% lower |
| Generated Markdown and JSON | Baseline | Byte-for-byte identical | No change |

This only speeds up report generation. It does not change raw samples, measured LSP request times, confidence intervals, or verdicts. Benchmark collection itself took 49.58 seconds and does not run this code path.

### DIFF
--- a/benches/lsp/README.md
+++ b/benches/lsp/README.md
@@ -83,6 +83,9 @@ The displayed base and head values are means of those per-session percentiles;
 they are descriptive values, not percentiles from 100 pooled request samples.
 Within each order stratum, the renderer pairs base and head session summaries
 and computes an exact, deterministic paired-bootstrap 95% confidence interval.
+It represents each replacement count vector once with its multinomial weight, so
+the interval uses the same ordered-resample distribution without a random or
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1363, opened 2026-09-04
fix(sema): match solc library ABI and declaration checks

DaniPopes:
Part of the split of the solc-vs-solar differential audit branch (#1360) into independent PRs; this one is based on `main`. #1370 builds on it.

solc leaves every library function with a storage-reference parameter or return, and every library function above `view`, out of the JSON ABI while keeping them all in the selector list; we printed them, and asserted on a mapping in such a parameter type. solc also rejects `payable` library functions (7708) and five neighbouring declaration forms: constructors and `fallback` in libraries (7634, 5982), `virtual` library functions (7801), `internal`/`private` `payable` functions (5587), and `payable` free functions (9559). This adds those checks with solc's codes, fixes the ABI output, and teaches the typos check to ignore commit hashes and literal identifiers.

### DIFF
--- a/crates/sema/src/output/abi.rs
+++ b/crates/sema/src/output/abi.rs
@@ -31,7 +31,7 @@ impl<'gcx> Gcx<'gcx> {
         }
         let is_library = c.kind.is_library();
         for f in self.interface_functions(id) {
-            if is_library && stores_data_in_storage(f.ty) {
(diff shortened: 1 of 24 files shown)


mattsse (approved):
lgtm

DaniPopes:
cyclops audit fast

PR #1365, opened 2026-09-04
fix(sema): accept programs solc accepts in the type checker

DaniPopes:
Part of the split of the solc-vs-solar differential audit branch (#1360) into independent PRs; this one is based on `main` and nothing depends on it.

Five valid programs from the solc semantic tests were rejected by the type checker: pushing an array literal into a storage array of static arrays, array literals of calldata slices, integer constants of signed type as array lengths, and error or event parameters named after sibling parameter types. The reverse gap, `bytes transient` and `string transient` state variables, is now rejected with solc's 1834. The review of those fixes found and fixed three more: slices of static calldata arrays were accepted, the name-hiding change had made bodiless-modifier and `try` clause parameters invisible, and an implemented function's parameter did not shadow its own type. The remaining difference in this area, inline array literals taking the element type of their destination, is documented as the intentional `TYPECK-003`.


DaniPopes:
cyclops audit fast

crates/sema/src/typeck/checker/mod.rs
```diff
@@ -184,6 +184,19 @@ impl<'gcx> TypeChecker<'gcx> {
         self.check_expr_with(expr, Some(expected))
     }
 
+    /// Checks the length expression of a fixed-size array type.
+    ///
+    /// Any integer expression is accepted, whatever its declared signedness and width. Only the
+    /// evaluated value matters, and [`crate::eval::eval_array_len`] checks it.
+    #[must_use]
+    fn check_array_size(&mut self, size: &'gcx hir::Expr<'gcx>) -> Ty<'gcx> {
+        let ty = self.check_expr(size);
```
mattsse, line 253: Accepting every integer type here also admits signed arithmetic that overflows its declared width, because the current constant evaluator produces the mathematical result instead of rejecting the typed overflow. For example, `int8 constant MAX = 127; uint256[MAX + 1] x;` and `int8 constant MIN = -128; uint256[-MIN] x;` both compile on this head, while solc rejects each with an arithmetic error. The existing tests cover signed division and a negative final length but not boundary overflow. Please preserve checked typed arithmetic when allowing signed lengths, or make constant evaluation reject these overflows.

mattsse (dismissed):
One correctness issue in signed constant array-length validation.

crates/sema/src/typeck/checker/mod.rs
DaniPopes: Fixed in 9a236f10: constant values now carry the declared integer type they were computed in, so typed arithmetic stays checked instead of yielding the mathematical result. `int8 MAX = 127; uint256[MAX + 1]`, `int8 MIN = -128; uint256[-MIN]`, and `uint8 M = 255; uint256[M + 1]` now report an arithmetic overflow like solc (2643/3667), including intermediate overflow such as `(MAX + 1) - 1`, while literal-only expressions stay unbounded and in-range signed lengths still compile. Coverage: boundary cases in tests/ui/typeck/eval.sol and a port of solc's constantEvaluator overflow, underflow, underflow_unary, and unary_fine tests.

crates/sema/src/typeck/checker/mod.rs
```diff
@@ -184,6 +184,19 @@ impl<'gcx> TypeChecker<'gcx> {
...
+        if !ty.is_integer() && !ty.references_error() {
```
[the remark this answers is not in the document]
DaniPopes: Fixed in 9a236f10: constant values now carry the declared integer type they were computed in, so typed arithmetic stays checked instead of yielding the mathematical result. `int8 MAX = 127; uint256[MAX + 1]`, `int8 MIN = -128; uint256[-MIN]`, and `uint8 M = 255; uint256[M + 1]` now report an arithmetic overflow like solc (2643/3667), including intermediate overflow such as `(MAX + 1) - 1`, while literal-only expressions stay unbounded and in-range signed lengths still compile. Coverage: boundary cases in tests/ui/typeck/eval.sol and a port of solc's constantEvaluator overflow, underflow, underflow_unary, and unary_fine tests.

DaniPopes:
cyclops audit fast

PR #1368, opened 2026-09-04
fix(codegen): spill across branches and accept recursion

DaniPopes:
Part of the split of the solc-vs-solar differential audit branch (#1360). Based on #1366 because the re-blessed constructor snapshots embed its storage `bytes` code; #1367 continues the stack on top of it. The fixes themselves are independent.

Two backend fixes. In the MIR-to-EVM emission loop, a block whose terminator kept its stack alive across an edge left the frame-slot stores to its successor, and the successor computed its stores from an availability set to which a not yet emitted predecessor contributed nothing; with the `else` arm of a branch laid out after the join, nobody stored one loop-carried value and the `else` path read an unwritten slot, returning a wrong value at `-Ogas`. Every value live into an already emitted successor is now stored before the terminator, except across loop back edges. The EVM IR stack verifier re-queued blocks at every new concrete entry depth and never converged on a recursive call edge, rejecting any recursive function before constantinople; it now propagates depth bounds per block, checks underflow at the minimum and the 1024 limit at the maximum, and runs at every EVM version.


mattsse (approved):
lgtm

crates/codegen/src/backend/evm/ir/verify.rs
```diff
@@ -327,15 +359,24 @@ impl<'a> Verifier<'a> {
...
-                        pending.push((target, depth));
+                    Some(bounds) => {
+                        // A cycle edge may lower the minimum, which terminates because depths
+                        // are non-negative, but it may not raise the maximum: one round of a
+                        // recursive call leaves its return address behind, so following the
+                        // growth would never converge.
+                        if depth < bounds.min {
+                            bounds.min = depth;
+                            pending.push((target, depth));
+                        } else if depth > bounds.max && !cycle_edges.contains(&(block_id, target)) {
```
mattsse: This drops every larger depth that crosses the DFS cycle-closing edge. If the same cycle has a low-depth entry into the target and an acyclic path enters the source at depth 1024, the target is never checked at 1024. For example, with `bb2 -> bb3` classified as the cycle edge, enter `bb3` at depth 0 on one branch and `bb2` at depth 1024 on another; a `push 1` in `bb3` deterministically overflows on the latter path, but the `.evmir` input is accepted. This is a one-pass overflow, not growth accumulated across recursion. The irreducible-cycle fixture only uses small depths, so it does not catch this. Please preserve enough of each alternate ingress maximum to check one complete traversal before cutting recursive growth.

mattsse (dismissed):
One verifier correctness issue.

DaniPopes:
cyclops audit fast

crates/codegen/src/backend/evm/ir/verify.rs
DaniPopes: Good catch. The cutoff was keyed on `(source, target)` identity alone, so a cycle edge classified during a low-depth traversal suppressed every later maximum over that pair, including a first arrival on an acyclic path. Fixed in 02b56ac9. The new `multi_entry_cycle_stack_overflow.evmir` fixture matches this shape: `bb1` enters `bb2` at depth 0, `bb4` enters `bb3` at 1022 without going around the cycle, and `bb3 -> bb2` is the classified edge. Before the fix the file was accepted and assembled; now it reports `push` growing the stack to 1025 words.

The identity check is replaced by a maximum split by how many cycle-closing edges the path carrying it has crossed: each block keeps an ingress maximum over paths that cross none and a post-cycle maximum over paths that cross exactly one, and crossing a cycle-closing edge promotes the former into the latter while dropping the latter. Every new ingress maximum is therefore propagated through one complete traversal of the cycle and checked there, and only a second traversal around the same cycle is cut. Both halves are least fixed points over the DAG the cycle edges leave behind, so termination is structural and the result no longer depends on worklist order or on which edge of a multi-entry cycle the DFS picked. The module documentation records the argument. Every existing expectation, including the recursion and irreducible-cycle fixtures, is unchanged, and bytecode and diagnostics are identical to the parent commit across the single-file contracts and the project archives.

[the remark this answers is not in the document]
DaniPopes: Agreed, and fixed in 02b56ac9 together with the thread above. The suppression is no longer by `(source, target)` identity. Maxima are tracked separately for paths that have crossed no cycle-closing edge and paths that have crossed exactly one, so a first high-depth arrival into a multi-entry cycle is always propagated once around the cycle and checked, while only repeated growth around the same cycle is cut. A path can only be cut at its second cycle-closing edge if the depth it carries was itself produced by going around a cycle, which is exactly the runtime-dependent quantity the verifier declines to bound, so recursion still compiles and nothing else stays unchecked. The multi-entry-cycle regression fixture reproduces your PoC and now fails verification at 1025 words.

DaniPopes:
cyclops audit fast

PR #1370, opened 2026-09-04
fix(sema): add solc's interface and free function checks

DaniPopes:
Part of the split of the solc-vs-solar differential audit branch (#1360). Based on #1363, next to whose library checks these live. The `try` clause checks moved to their own PR (#1372) because they reuse the pre-byzantium discard tracking.

Declaration checks solc applies that we lacked: non-`external` interface functions (1560), `virtual` free functions (4493), interface constructors, implemented interface functions, interface modifiers and variables (6482, 4726, 6408, 8274), the `virtual` interface function warning (5815), `virtual` library modifiers (3275 instead of the function message), bodiless modifiers (8063), bare `override` on free functions (1750), free functions with modifiers (5811), and the missing 4549 on the library `receive` error. solc's syntax tests for these are ported where they exist.

### DIFF
--- a/crates/sema/src/typeck/mod.rs
+++ b/crates/sema/src/typeck/mod.rs
@@ -335,11 +335,30 @@ fn same_external_params<'gcx>(gcx: Gcx<'gcx>, a: Ty<'gcx>, b: Ty<'gcx>) -> bool 
     key(a) == key(b)
 }
 
+/// Checks that every declaration without a body is allowed to have none.
+///
+/// References:
+/// - functions: <https://github.com/argotorg/solidity/blob/8a079791d9cca7a6c03fd6a8429b93aa3bddefed/libsolidity/analysis/TypeChecker.cpp#L448-L460>
+/// - `virtual`: <https://github.com/argotorg/solidity/blob/8a079791d9cca7a6c03fd6a8429b93aa3bddefed/libsolidity/analysis/TypeChecker.cpp#L306-L318>
+/// - modifiers: <https://github.com/argotorg/solidity/blob/8a079791d9cca7a6c03fd6a8429b93aa3bddefed/libsolidity/analysis/TypeChecker.cpp#L300-L301>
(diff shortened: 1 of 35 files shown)


DaniPopes:
cyclops audit fast

crates/sema/src/typeck/mod.rs
```diff
@@ -457,6 +499,109 @@ fn check_library_functions(gcx: Gcx<'_>, contract_id: hir::ContractId) {
...
+fn check_interface_members(gcx: Gcx<'_>, contract_id: hir::ContractId) {
+    let contract = gcx.hir.contract(contract_id);
+    if !contract.kind.is_interface() {
+        return;
+    }
+
+    // Iterate over the items in declaration order to report them in the same order solc does.
+    for &item in contract.items {
+        match item {
+            ItemId::Function(f_id) => {
```
mattsse, line 518: `contract.items` also contains HIR entries for Yul functions collected from inline assembly. An implemented interface function containing `assembly { function inner() {} }` therefore reaches this branch for `inner`: Solar emits 4726 and 1560 on the Yul helper, while solc only reports 4726 on the outer Solidity function. Please skip `f.is_yul` here and add an inline-assembly regression test.

mattsse (changes requested):

crates/sema/src/typeck/mod.rs
DaniPopes: same for generated functions, handled below

mattsse (approved):

crates/sema/src/typeck/mod.rs
DaniPopes: Fixed in 0fd44bed: Yul functions are skipped like getters (they are items of the contract but belong to the Solidity function that contains them), and the interface implementation test now has an implemented function with an inline-assembly helper, reporting 4726 once like solc.

PR #1372, opened 2026-09-04
fix(sema): check the try returns clause against the callee

DaniPopes:
Part of the split of the solc-vs-solar differential audit branch (#1360). Based on #1369 because the check reuses the pre-byzantium discard tracking and the slot-encoding rule introduced there.

solc compares a `try` statement's `returns` clause with the callee's return types and reports 2800 for a count mismatch and 6509 for a type mismatch, requiring exact types including data location; we bound the variables without any comparison, so a `bool` binding on a `uint256` return or a two-variable clause on a one-value callee compiled. The check runs after the statement is walked, applies to external, library, and creation callees only (solc rejects the rest with its own codes), and avoids double reports next to the declaration-location and pre-byzantium checks. solc's `invalid_returns.sol` and `returns_mismatch.sol` are ported 1-1.

### DIFF
--- a/crates/sema/src/typeck/checker/mod.rs
+++ b/crates/sema/src/typeck/checker/mod.rs
@@ -1334,6 +1334,70 @@ impl<'gcx> TypeChecker<'gcx> {
         Ok(())
     }
 
+    /// Checks a `try` statement's `returns` clause against the callee's return values.
+    ///
+    /// solc requires the clause to repeat the callee's return types exactly, without implicit
+    /// conversions and with the same data location (`TypeChecker::endVisit(TryStatement)`): the
+    /// decoder writes the callee's values straight into the clause's variables, so a differing
+    /// declaration would reinterpret them.
+    fn check_try_returns_clause(&self, try_: &'gcx hir::StmtTry<'gcx>) {
+        let clause = &try_.clauses[0];
+        if clause.args.is_empty() {
+            return;
+        }
+        let hir::ExprKind::Call(callee, ..) = try_.expr.peel_parens().kind else { return };
+        let Some(&callee_ty) = self.results.expr_types.get(&callee.id) else { return };
+        let TyKind::Fn(f) = callee_ty.kind else { return };
+
+        if f.returns.len() != clause.args.len() {
+            self.dcx()
+                .err(format!(
+                    "function returns {} value{}, but the `returns` clause has {} variable{}",
+                    f.returns.len(),
+                    pluralize!(f.returns.len()),
+                    clause.args.len(),
+                    pluralize!(clause.args.len())
+                ))
+                .code(error_code!(2800))
+                .span(clause.span)
+                .emit();
+        }
+
+        // solc still compares the pairs it has, so a count mismatch can carry a type mismatch.
+        for (&id, &expected) in std::iter::zip(clause.args, f.returns) {
+            let var = self.gcx.hir.variable(id);
+            let actual = self.gcx.type_of_item(id.into());
+            if actual == expected || actual.references_error() || expected.references_error() {
+                continue;
+            }
+            // A clause variable's only valid data location is `memory`; a rejected one has
+            // already been reported and rewritten to `memory`, so its type says nothing.
+            if matches!(actual.kind, TyKind::Ref(..))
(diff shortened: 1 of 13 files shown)


mattsse (approved):
lgtm

DaniPopes:
cyclops audit fast

crates/sema/src/typeck/checker/mod.rs
```diff
@@ -1290,6 +1290,77 @@ impl<'gcx> TypeChecker<'gcx> {
...
+        // solc still compares the pairs it has, so a count mismatch can carry a type mismatch.
+        for (&id, &expected) in std::iter::zip(clause.args, f.returns) {
+            let var = self.gcx.hir.variable(id);
+            let actual = self.gcx.type_of_item(id.into());
+            if actual == expected || actual.references_error() || expected.references_error() {
+                continue;
+            }
+            // A clause variable's only valid data location is `memory`; a rejected one has
+            // already been reported and rewritten to `memory`, so its type says nothing.
+            if matches!(actual.kind, TyKind::Ref(..))
```
mattsse: This recovery guard only skips reference types. For `try i.f() returns (bool memory b)` when `f` returns `uint256`, `var_type` already reports the illegal `memory`, but `actual` is an elementary type, so this falls through and adds a 6509 mismatch. solc 0.8.36 emits only the data-location error. Please also skip comparison when a non-reference clause variable has any explicit data location; the current test covers only the reference/`storage` branch.

mattsse (dismissed):
One invalid-declaration recovery case still diverges from solc.

crates/sema/src/typeck/checker/mod.rs
DaniPopes: Good catch, fixed in 7071e17b. The guard now compares the declared location against the single location the variable's type accepts at all, `memory` for a reference or mapping type and none for anything else, and skips the 6509 comparison whenever they differ, instead of only for the reference case. Either way `var_type` has already reported the declaration and rewritten the location, so comparing the rewritten type would invent a second error. The UI test gained `bool memory b`, `uint256 calldata x` and `bool storage c` against a `uint256` callee; each now reports exactly one diagnostic, the same single data-location error solc 0.8.36 reports for all three.

crates/sema/src/typeck/checker/mod.rs
```diff
@@ -1290,6 +1290,77 @@ impl<'gcx> TypeChecker<'gcx> {
...
+            return;
+        }
+        let hir::ExprKind::Call(callee, ..) = try_.expr.peel_parens().kind else { return };
+        let Some(&callee_ty) = self.results.expr_types.get(&callee.id) else { return };
+        let TyKind::Fn(f) = callee_ty.kind else { return };
+        // solc reaches the clause only for the callee kinds `try` accepts, returning after 2536
+        // for any other one. The return values of the rest are not what the clause would bind:
+        // a builtin's are not even always a type, so comparing them says nothing.
+        if !matches!(f.kind, TyFnKind::External | TyFnKind::DelegateCall | TyFnKind::Creation) {
+            return;
```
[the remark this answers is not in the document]
DaniPopes: Fixed in 1ed2609d, and the remaining clause checks came along with it. The statement is now validated as a whole after the called expression is checked, mirroring solc's `TypeChecker::endVisit(TryStatement)`: the target must be a call whose callee is external, a delegatecall, or a creation; every other target is reported with 2536 and stops before any clause is looked at, and solc's 5347 case (a target that is not a function call, such as a type conversion) lands on the same message. On top of that a repeated clause is rejected with 5320, 1036 and 6732, each pointing at the first one, and a parameter list that is not the decoded error's own signature with 6231, 2943 and 1271. The new `tests/ui/typeck/try_catch_clause_checks.sol` covers all of these plus the pruned internal-call case, and solc 0.8.36 reports the same 15 errors on the same clauses. Two test updates fell out of it: the previously accepted callees in `try_returns_clause.sol` now carry 2536 annotations, and `evm_version/byzantium.sol` had a bare `catch` following `catch (bytes memory)`, which solc rejects with 5320, so that trailing clause is gone.

DaniPopes:
cyclops audit fast

PR #1373, opened 2026-09-04
rewrite(codegen): use isle, egraph and decl MIR/EVMIR

djolertrk:
Use declarative way to define MIR and EVM IR ops. Besides that, add ISLE and egraph into pipeline.


.github/workflows/ci.yml
```diff
@@ -15,6 +15,26 @@ env:
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      - run: uv run scripts/test_evm_rules.py
+      - name: Verify compiled ISLE word rules
+        run: uv run scripts/verify_evm_rules.py verify --output target/evm-rules/proofs.json --artifacts target/evm-rules/smt
+      - name: Preserve proof queries and coverage
+        if: always()
```
DaniPopes: why is this preserved

crates/codegen/isle/egraph.isle
DaniPopes, line 1: https://github.com/bytecodealliance/wasmtime/blob/668016926adfd1b8a79dbce894f1e203d8892599/.gitattributes#L9-L10
let's add this to gitattributes

crates/codegen/isle/egraph.isle
```diff
@@ -0,0 +1,438 @@
...
+(decl pure partial field_offset (MemoryObjectLayout u64) u64)
+(extern constructor field_offset field_offset)
+(decl pure layout_kind (MemoryObjectLayout) MemoryObjectKind)
+(extern constructor layout_kind layout_kind)
+
+;;;; In-place rewrites.
+
+(decl multi rewrite (Op) Op)
+
+;; Fold a constant into an existing offset of the other operand, or move the
```
DaniPopes: please rewrite the comments to be showing the pattern in pseudocode like in mir lowering and for all patterns to be documented

crates/codegen/isle/egraph.isle
```diff
@@ -0,0 +1,438 @@
...
+(rule 8 (simplify (Op.Or (band _ b) b)) b)
+(rule 7 (simplify (Op.Or a a)) a)
+(rule 6 (simplify (Op.Or (and a (all_ones)) _)) a)
+(rule 5 (simplify (Op.Or a (zero))) a)
+(rule 4 (simplify (Op.Or _ (and b (all_ones)))) b)
+(rule 3 (simplify (Op.Or (zero) b)) b)
+(rule 2 (simplify (Op.Or (bnot b) b)) (imm (u256_max)))
+(rule 1 (simplify (Op.Or a (bnot a))) (imm (u256_max)))
+
+(rule 9 (simplify (Op.Xor a (bxor a b))) b)
```
DaniPopes, line 490: why is there a number in 'rule'?

crates/codegen/isle/word_sequence.isle
```diff
@@ -0,0 +1,82 @@
+;; Proved multi-operation recipes; materialize only after target cost selection.
+
+;; (and (not x) (not y)) => (not (or x y))
```
DaniPopes: please rewrite these comments in human readable pseudocode (not lisp)

.github/workflows/ci.yml
djolertrk: to reproduce solver failures using the exact SMT queries and coverage report. I’ve changed it to upload only on failure, with seven-day retention, I guess it works?

crates/codegen/isle/egraph.isle
djolertrk: yep, added *.isle linguist-language=lisp to .gitattributes, matching Wasmtime.

crates/codegen/isle/egraph.isle
djolertrk: It’s ISLE’s match priority, higher numbers are tried first when rules overlap. Here it controls which simplification wins. I’ve added a comment explaining it.

crates/codegen/src/backend/evm/codegen/stack/scheduler.rs
```diff
@@ -342,6 +312,9 @@ impl OperandCostModel {
     }
 }
 
+/// One very-low instruction without an immediate.
+const VERY_LOW: Cost = Cost::new(GasTier::VeryLow.fixed_gas(), 1);
```
DaniPopes: should have a specific module or struct with this info?

benches/runtime/Encoding.sol
DaniPopes, line 1: shouldnt these be in testdata?

benches/runtime/cases.py
```diff
@@ -22,6 +22,84 @@
 MIXED_BYTES32 = "0x" + "ff" * 30 + "0000"
 SIGNED_HASH = "0x7d768af957ef8cbf6219a37e743d5546d911dae3e46449d8a5810522db2ef65e"
 
+# Exercise both array iteration and bytes tails around the ABI word boundary.
+GOVERNOR_PROPOSALS = tuple(
```
DaniPopes, line 26: https://www.getfoundry.sh/perf/solar/?base=4495c3e9&head=f9a57a2d&benchmark=solady-lib-string#benchmarks

I guess this is why this reports so much more gas, we're just running more benches right? can we add them in a separate pr so we can see the results on this branch?

crates/codegen/isle/extractors.isle
DaniPopes, line 1: just making sure, AGENTS.md has enough info for ISLE right? DSL reference, how to write rewrites etc

scripts/evm-rules/replay.py
DaniPopes, line 1: move all to subdir?

crates/codegen/isle/egraph.isle
DaniPopes, line 1: do these supersede any existing passes? should we move all rewrites into ISLE, and what benefit would this bring

DaniPopes:
proofs CI failing

benches/runtime/cases.py
djolertrk: yeah, good idea, https://github.com/paradigmxyz/solar/pull/1438

crates/codegen/src/backend/evm/codegen/stack/scheduler.rs
djolertrk: Agreed! I moved the fixed stack and frame costs into the shared target cost model

benches/runtime/Encoding.sol
djolertrk: yeah

crates/codegen/isle/extractors.isle
djolertrk: yes yes, it is there, added

djolertrk:
cyclops audit

djolertrk:
cyclops audit

djolertrk:
cyclops audit

DaniPopes:
reviewing manually a small MIR diff from the benchmarks: https://www.getfoundry.sh/perf/solar/?base=123626e7564606956bad6644c1d977b39f723824&head=f41e47bc97211443816357af9b4926e49fe47f8b&benchmark=seeded-words&metric=total_gas&view=files&baseRevision=7e553244df8298af651337a3e69032e7c227010159bcd22762bb9130ea18981a&headRevision=cec143e5f30c4359088217d6c6019b1909cee39e95816ecf8036847346c9caaf&file=mir.mir

1. looks like there's a leftover dead jump, bb0->bb1, bb5 should not be necessary.
```
fn @difference() [selector=0x214636e1, pure] {
  bb0:
    jump bb5
  bb5:
    jump bb1
  bb1:
```

2. similarly, `bb4: invalid` is unreachable so it shouldn't even exist.

I think this should be a more streamlined optimization: after lowering dispatcher the internal MIR functions that actually implement the solidity functions will all have some logic shared, in the case of this benchmark its everything except the loop body. So it should be recognized after inlining these MIR functions into the switch of the dispatcher, and then normal passes should recognize this. EVMIR recognizes this after lowering and creates an outlined function to share the input loading but it should be done at MIR.

djolertrk:
> reviewing manually a small MIR diff from the benchmarks: https://www.getfoundry.sh/perf/solar/?base=123626e7564606956bad6644c1d977b39f723824&head=f41e47bc97211443816357af9b4926e49fe47f8b&benchmark=seeded-words&metric=total_gas&view=files&baseRevision=7e553244df8298af651337a3e69032e7c227010159bcd22762bb9130ea18981a&headRevision=cec143e5f30c4359088217d6c6019b1909cee39e95816ecf8036847346c9caaf&file=mir.mir
> 
> 1. looks like there's a leftover dead jump, bb0->bb1, bb5 should not be necessary.
> 
> ```
> fn @difference() [selector=0x214636e1, pure] {
>   bb0:
>     jump bb5
>   bb5:
>     jump bb1
>   bb1:
> ```
> 
> 2. similarly, `bb4: invalid` is unreachable so it shouldn't even exist.
> 

hmm, yeah, I find CFG cleanup missed from time to time, thank you! the fix was easy

> I think this should be a more streamlined optimization: after lowering dispatcher the internal MIR functions that actually implement the solidity functions will all have some logic shared, in the case of this benchmark its everything except the loop body. So it should be recognized after inlining these MIR functions into the switch of the dispatcher, and then normal passes should recognize this. EVMIR recognizes this after lowering and creates an outlined function to share the input loading but it should be done at MIR.

yes, that is true. I tried it at some point at MIR level, but passes down there in the pipeline did not optimise it, and it introduced some regressions, but let me try again now.

djolertrk:
cyclops audit

djolertrk:
cyclops audit

djolertrk:
cyclops audit

djolertrk:
cyclops audit

crates/codegen/isle/egraph.isle
djolertrk: Well, not a full replacement for sure. They only replace the local pure expression part of CSE for example. I’d move simple local rewrites to ISLE, but keep things like dataflow, alias analysis, CFG, and stack optimizations in Rust, since it is natural to stay that way. ISLE mainly gives us typed rules, overlap checks, cost-based alternatives, and easier verification.

PR #1377, opened 2026-09-04
refactor(codegen): rename `internal_call` to `icall`

DaniPopes:
Renames the MIR `InternalCall` instruction to `ICall` and its text-format mnemonic from `internal_call` to `icall`, so that MIR dumps read `v1 = icall @f, 1, v0` instead of the longer form. The `EffectKind::InternalCall` variant, the `sym::internal_call` symbol, and the builder helpers `internal_call`/`internal_call_void` become `ICall`, `sym::icall`, and `icall`/`icall_void`; pass-local and backend helper names derived from the old name (`has_internal_call`, `emit_internal_call`, `InternalCallStackEdge`, and similar) follow along for consistency.

This is a pure rename with no behavioral change. UI fixtures, FileCheck patterns, and snapshots are updated to the new mnemonic, while test file names such as `internal_call_fallbacks.sol` are kept as-is since they describe the Solidity feature rather than the MIR spelling.


mattsse (approved):
lgtm

PR #1384, opened 2026-09-05
fix(sema): resolve `length` on calldata slices

DaniPopes:
Member lookup returned an empty member list for every slice type, so `data[start:end].length` on a `bytes calldata` or `T[] calldata` parameter was rejected with "member `length` not found on type `bytes calldata slice`", while the same length read through a `bytes calldata` local bound to the slice compiled fine. Solc types index range access on a dynamically sized calldata array as a slice of that array and resolves members against the array itself, which gives `length` to every slice except a slice of a `string`, mirroring `ArrayType::nativeMembers`.

The slice arm of `native_members` now returns the array member list unless the sliced array is a `string`. Indexing into a slice was already accepted by the type checker, and codegen already lowers the length of a slice value through `slice_len`, so the member table was the only missing piece. Assigning to a slice's `length` is still rejected as a non-lvalue, as in solc.

The new lowering test runs `bytes` and `uint256[]` slice lengths, open-ended slices, indexing into a slice, chained slices, and a `string` slice read through `bytes` under the standard codegen matrix, and the typeck test pins the accepted forms alongside the `string` slice and lvalue rejections.


DaniPopes:
cyclops audit fast

PR #1388, opened 2026-09-05
refactor(codegen): rebuild the EVM backend

DaniPopes:
Rebuild the EVM backend after the agreed 40-file deletion, with focused stack scheduling, physical block IR, deployment construction and a primitive assembler. Stack scheduling retains private SSA identities at the MIR-to-EVM boundary. Target optimizations operate on blocks; the assembler uses a byte buffer with sparse relocations and resolves label widths to a fixed point. No deleted implementation or legacy fallback was recovered.

Bounded shared-capture contraction prevents a spill restore from overwriting source memory at address 576. The latest late-DCE change retains a stored word across a literal operand, removing a redundant reload with explicit stack-capacity and metadata guards. It reduces UI creation/runtime totals by 700 bytes under Gas and 682 under Size, and the heavy corpus by 149,352 bytes, without individual object growth. All 175 hot-gas labels and 139 runtime observations remain identical in each mode. These corpora overlap. Broader trials with individual growth were rejected.

The backend has 17,661 physical Rust lines across 48 files, versus the 34,638-line deletion inventory. The shared-capture milestone adds 352 MIR lines; the latest change adds 31 backend lines. Counts include comments, blanks and local tests. Main afdd3a1c is merged. The latest Full/Size compiler-time comparisons are -0.718%/+1.741%; a separate quiet Size ABBA repeat is +0.020%. These establish no causal speed claim. Exact comparisons and reviewed snapshot changes are recorded in docs/evm-store-reload-milestone.md. The current CodSpeed report flags 18 compiler benchmarks against main afdd3a1c with differing-runtime-environment warnings; these concerns remain open.

This remains a draft: general source-memory ownership and existing performance debts remain open. Fifteen of 175 calls still use more gas than retained main 25b0c078. The alias fixture retains 49/10 additional Gas/Size runtime bytes versus the sealed baseline. Positive sealed size debt is 30,130,696 bytes across 1,007 heavy objects, including embedded-child amplification; this sums individual increases rather than net corpus growth. The full ledger is in docs/evm-rewrite-progress.md.


DaniPopes:
: (

PR #1389, opened 2026-09-06
perf(lsp): reuse unchanged analysis snapshots

0xKarl98:
Towards OSS-738 , 

This change adds a conservative analysis snapshot fast path for unchanged in-memory LSP workspaces. After a successful analysis, the server retains the published analysis output and reuses it when the VFS content revision and configuration are unchanged. Edits, disk-backed analysis, cache invalidation, cancellation, stale analysis versions, configuration changes, and failed analyses continue through the existing rebuild path.

The implementation also adds a production-path Criterion benchmark and a regression test covering snapshot reuse and invalidation. The fast path publishes the same analysis output and diagnostics, so LSP providers continue to observe the existing symbol and diagnostic state.

## Benchmark

Measured with the new `lsp/incremental-analysis` Criterion workload in the same checkout. These are two execution modes in the same benchmark group, rather than separate main-versus-PR runs: cold analysis creates a fresh production analysis state for each iteration, while unchanged analysis repeats one state after its first successful run. The reduction therefore measures the repeated-analysis fast path against the cold path in the candidate build.

| Run | Execution mode | Median |
| --- | --- | ---: |
| 10 samples, 1 s measurement, 1 s warm-up | Cold analysis | 7.37 ms |
| 10 samples, 1 s measurement, 1 s warm-up | Unchanged snapshot | 1.02 ms |
| 20 samples, 1 s measurement, 1 s warm-up | Cold analysis | 7.46 ms |
| 20 samples, 1 s measurement, 1 s warm-up | Unchanged snapshot | 0.99 ms |

The unchanged snapshot path is 86.1% lower than cold analysis in the 10-sample run and 86.8% lower in the 20-sample run.

The benchmark demonstrates the intended repeated-analysis workload; it does not claim that first-run analysis or analyses invalidated by edits become faster, and it is not a main-versus-PR regression measurement.

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -5,9 +5,9 @@ use lsp_types::{GotoDefinitionResponse, HoverContents, OneOf, Position, Url};
 use solar_config::CompileOpts;
 use solar_lsp::{
     BenchmarkAnalysis, BenchmarkDocumentUpdate, BenchmarkOpenDocuments, BenchmarkProject,
-    BenchmarkRequest, BenchmarkResponse, BenchmarkSelectionRangeRequests,
-    BenchmarkWorkspaceDiscovery, BenchmarkWorkspacePathQueries, BenchmarkWorkspaceReports,
-    benchmark_selection_ranges,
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

PR #1390, opened 2026-09-06
test: integrate local debug-info differentials

djolertrk:
Add an opt-in cargo tq debug-diff suite that compiles fixtures with solc and this compiler. Execute them locally through soldb and compare ETHDebug and legacy source-map stops across optimization modes.

Require executable checkpoints for cross-compiler comparisons. Retain inputs, artifacts, traces, and reports. Keep the newly exposed require metadata gaps as separate failing reproducers. The baseline covers 30 passing comparisons without accepting known differences.

Depends on the `soldb` (main branch for now; the next soldb release will contain this feature).

```
% SOLDB=/path/to/target/debug/soldb cargo tq debug-diff
```


djolertrk:
cyclops audit

DaniPopes (approved):

PR #1393, opened 2026-09-06
perf(lsp): reduce indexing time and memory

DaniPopes:
Repeated editing duplicated large symbol indexes and rebuilt identical member completion lists, while position conversion and import detection repeatedly scanned or parsed whole files. Share cached indexes and context-specific completion lists, convert Unicode positions within the current line, reject non-import completion positions lexically, and scan rope newlines by chunk. Keep batch indexes owned through aggregation and share only the finished index. Publish immutable indexes through ArcSwap so active readers do not block replacement, and use the entry API for member completion caching. Grouped LSP benchmark labels include their workflow in CodSpeed.

Before the ArcSwap follow-up, in three alternating profiling-build sessions on an M4 Max, each creating three test files and making 12 edits, peak RSS falls from 1,097 to 489 MiB in Uniswap v4 periphery and from 1,038 to 415 MiB in v4 core. Solady edit responses fall from 480 to 366 ms, including the 250 ms debounce, and new-test responses from 231 to 114 ms. The original and candidate use identical repository configurations, and all 21 paired symbol-response streams match. Regression coverage preserves Unicode and CRLF positions, import recovery, completion scope isolation, shared cache ownership, and release of retired snapshots after readers finish.


mattsse (approved):
lgtm

PR #1396, opened 2026-09-07
perf(lsp): reduce query work and hierarchy memory

DaniPopes:
Reduce allocation work and retained memory in VS Code LSP workflows. Type and call hierarchy indexes share source URIs across keys, internal items, and URI-indexed maps, constructing owned LSP items only for responses. Against 1bf4408f, this removes 14.5 MiB of live analysis allocations on Solady and 26.9 MiB on OpenZeppelin after querying call hierarchy. Repeated analysis and drop returns to the same allocation baseline.

Opaque hierarchy data uses versioned flat arrays instead of nested JSON object maps. The 128-result Criterion cases decrease from 58.4 to 11.7 µs for type hierarchy and from 71.4 to 24.0 µs for call hierarchy compared with the object payloads. OpenZeppelin returns the same hierarchy results apart from the opaque payload encoding. Foundry profile and fixture marker lookups use hash maps; import completion retains its tree to bound the alphabetically first candidates.

CodeLens reference counts avoid an intermediate allocation and sort, reducing the 256-function Criterion case from 402 to 382 µs. Folding reuses the open document's rope, avoiding 21.4 MiB of transient allocations on the large Optimism fixture with similar runtime. These changes complement #1395 and keep the existing analysis cache design.


DaniPopes:
cyclops audit fast

mattsse (approved):
lgtm

DaniPopes:
cyclops audit fast

PR #1412, opened 2026-09-07
feat(bench): add solx reference compiler

DaniPopes:
Add NomicFoundation/solx as a separate reference compiler in the runtime benchmarks, with its own compilation, gas, runtime checks, and artifacts. The CI Markdown report includes its gas, bytecode size, compilation time, and peak memory measurements alongside solc. Both reference compilers request Yul outputs and save `ir.yul` and `optimized-ir.yul` when returned.

CI pins solx 0.1.8 and runs it only on pushes to main. PR jobs reuse matching results from the exact-base artifact and never run solx, including when rebuilding a missing baseline.


mattsse (approved):
lgtm

PR #1413, opened 2026-09-08
feat: expose rooted struct storage layouts

decofe:
Expose `Gcx::storage_layout_for_struct` to compute a struct's fields at an arbitrary storage root using the existing packing and type generation. Nested members keep relative slots, and root addressing wraps modulo 2^256.

This is the generic compiler primitive proposed in [the Foundry ERC-7201 plan](https://github.com/foundry-rs/foundry/pull/15299#issuecomment-5317766666), based on Mablr's draft. ERC-7201 discovery and policy remain in Foundry.

Code and tests were written by Centaur with AI assistance.

Prompted by: @DaniPopes

PR #1414, opened 2026-09-08
fix(bench): ignore reference failure noise

DaniPopes:
Reference compiler failures were marking cross-compiler checks as failed and posting benchmark comments on unrelated PRs even when our compiler's output was unchanged, as in #1413. Solx rejects several archived EVM targets, so those failures persist across runs.

Ignore reference failures in CI warnings and in the comparison used to trigger PR comments. Keep their raw results and successful measurements, and continue reporting failures from our compiler and result mismatches involving it.


mattsse (approved):
lgtm

PR #1419, opened 2026-09-08
chore(bench): shorten codegen PR comments

DaniPopes:
Codegen benchmark comments currently include the full report. Show a compact gas and bytecode size overview without counts, followed by changed benchmarks compared with the base branch. Flag incomplete results and add a button that opens the benchmark overview for the PR's base and head commits.

Use the compact comment for both same-repository and fork PRs. Keep detailed Markdown in local output, CI job summaries, and artifacts, with counts removed from its overview too. Baseline reports show changed benchmarks, per-call gas changes, and artifact details without the solc/solx comparison tables.


mattsse (approved):
lgtm

PR #1425, opened 2026-09-09
perf(lsp): defer import path parsing

0xKarl98:
Towards OSS-738 , 

Import-literal go-to-definition currently parses the full source buffer before checking the analysis index. For clean open documents, the analyzed `DocumentLinkIndex` already contains the import range and resolved target, so that parsing work is unnecessary.

- Classify clean, fully analyzed open documents separately and query the existing import index first.
- Defer source scanning, `Rope` to `String` conversion, VFS overlay collection, and import resolution to the fallback path.
- Keep dirty buffers, closed files, stale analysis, malformed literals, and unowned workspace requests on their existing behavior.
- Preserve ordinary code-symbol go-to-definition and require an import resolution context before using an indexed import target.

Measured with the production stdio probe on a 5.38 MB source file:

| Case | Baseline p50 | Candidate p50 | Change |
| --- | ---: | ---: | ---: |
| Import-literal go-to-definition | 0.139 ms | 0.038 ms | 72.5% lower |

The response locations matched in both runs. The focused and full `solar-lsp` test suites pass.

### DIFF
--- a/crates/lsp/src/handlers/reqs.rs
+++ b/crates/lsp/src/handlers/reqs.rs
@@ -561,37 +561,85 @@ pub(crate) fn goto_definition(
         &analysis_revision,
     );
     let config = state.config.clone();
+    let vfs = state.vfs.clone();
     async move {
         let Some(latest_analysis) = latest_analysis else { return Ok(None) };
         let symbol_tables = latest_analysis.await?;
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1427, opened 2026-09-09
fix: preserve library address relocations

DaniPopes:
Library addresses must remain unresolved symbols until linking. Represent them as IDs into a module-owned table of source-qualified names, preserving the same IDs through MIR, EVM IR, and the compact assembler. Keep relocations through constant-data packing and embedded creation and runtime bytecode, remapping child IDs into the parent table. Support `address(Library)` and keep same-named libraries from different source files distinct.

Downstream trigger: Aave v3 Origin could not compile its Pool getters for `address(FlashLoanLogic)`, `address(BorrowLogic)`, `address(LiquidationLogic)`, `address(PoolLogic)`, and `address(SupplyLogic)`, blocking its deployment and Pool test suites. This removes that compilation blocker.


DaniPopes:
cyclops audit

PR #1410, opened 2026-09-07
fix: correct ABI data locations and lowering

DaniPopes:
External Foundry suites exposed corrupted ABI values and failures in constructor-reachable lowering. Encode memory arguments using their actual location, including `abi.encodeCall`, and materialize storage structs before placing them in memory structs.

Keep constructor-reachable array encoders inline so call frames cannot overwrite their unfinished output. Remove detached ABI-body blocks before lowering orphaned slice projections. Cover these cases with reduced runtime and MIR regressions.

Downstream tests fixed by these ABI/lowering changes:

| Project | Affected tests |
| --- | --- |
| forge-std, `Config.t.sol` | `test_configExists`, `test_loadConfig`, `test_loadConfigWithoutRpcEndpoint`, `test_writeConfig`, `test_writeUpdatesBackToFile`, `testRevert_ChainNotInitialized`, and `testRevert_WriteToFileInForbiddenCtxt`; constructor string-array encoding previously reverted. |
| v4-periphery, PositionDescriptor | `test_tokenURI_succeeds` and `test_native_tokenURI_succeeds`; generated JSON was corrupt. |
| v4-periphery, ReservesLens | `test_pagedEqualsSingleShot`, `test_pagedMinimumBudgetMakesProgress`, `test_fullRangePositionMaximumTickSpacing`, `test_pagedFinalPageProbesHookStats`, and `test_fuzz_matchesIndependentAggregate`; cursor bytes were encoded from the wrong data location. |
| v4-periphery, Execute | `test_execute_rebalance_perfect`; the call previously reverted with `PoolNotInitialized`. |

Reduced regressions include `testConstructorReturnedStrings`, `testPagedReturn`, `testStorageStructConstructorArgument`, `test_InternalStringConstructor`, and the `abi.encodeCall` memory-layout and detached-ABI-body UI fixtures. The standalone branch passes all 15 abi-encoding and 5 constructor-args tests at `-Onone`, `-Ogas`, and `-Osize`, with and without debug output; debug output leaves bytecode unchanged. The full downstream replay predates the split. Foundry block-number/chain-ID cheatcode differences are outside these fixes.

PR #1428, opened 2026-09-09
fix: enforce compiler memory ownership

DaniPopes:
Unrestricted assembly can address memory used by compiler spills, call frames, and tuple buffers. Track that ownership contract through MIR and internal call contexts, keep compiler state on the stack, and reject any remaining private-memory fallback. Skip storage promotion in these contexts and preserve ownership metadata on promoted temporaries and internal tuple projections. Closed stack recovery clears its temporary words and remains disabled when source `msize()` can observe the expansion. Recognize the exact legacy `@solidity memory-safe-assembly` annotation using solc’s comment and tag boundaries.

Keep constructor immutables in SSA until source execution ends, merging assignments across branches and loops from their default-zero values. Pass internal, decoded, and library-call tuples directly when their uses permit it. Scope constructor argument analysis to its reachable callers and preserve immutable payloads across panic paths. Retry constructor emission when a stack ABI is abandoned, reset each attempt’s state, and reject fallbacks that cannot recover. Keep annotated recursive Yul tuples on the required stack convention, and preserve Yul identity through function merging and MIR round trips. Preserve transaction-halting assembly returns through internal calls, and support wide stack layouts with each EVM version’s instruction limits. Debug output leaves executable bytecode unchanged.

Regression coverage includes annotation boundaries, argument entry loads, constructor retry state, immutable branch joins and loops, constructor helper state and OpenZeppelin Checkpoints lookup, overwritten frame prefixes, tuple lifetime and packed-storage assignments, dynamic ABI tuples, wide branches and calls, recursive helpers, assembly halts, and observable `msize()`. The reduced Foundry projects cover constructor-args, structs, multi-return, low-level-calls, stack-deep, libraries, library-storage-pointers, and both Unifap variants.

Solady matches the solc 0.8.36 baseline in all six optimization/debug configurations: 1,571 passes and the same existing SafeTransfer failure. Debug output leaves bytecode identical for all 256 contracts in each optimization mode, and debug and profiling compiler builds emit the same bytecode. The full local corpus passes all 15 runtime cases and 9 compilation-only projects, including Seaport, V4, Morpho, OpenZeppelin, and Solady.

Stack-owned state increases Nitro’s runtime bytecode from 11,183 to 16,692 bytes. The full Seaport project also takes 322 seconds to compile in this debug build versus 53 seconds in the preserved baseline; both compile all 432 contracts.

Addresses the [Cyclops frame-aliasing finding](https://github.com/paradigmxyz/solar/pull/1410#discussion_r3953531270).


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1431, opened 2026-09-10
perf(lsp): cache completion positions

0xKarl98:
Towards OSS-738 , 

Large Solidity files currently pay the full document-position and source-materialization cost on every completion request, even when the open document has not changed. This adds repeated line-index scans and Rope-to-String copies to the warm completion path.

- Add an exact-content `CompletionSource` backed by the existing immutable `Arc<VfsFile>`.
- Lazily cache one `LspPositionIndex<Rope>` per VFS file content and reuse the existing contiguous analysis source.
- Compute the checked cursor offset once per completion request and share it between NatSpec and import completion.
- Preserve cache safety through the existing VFS lifecycle: content edits replace the file object; unchanged version updates and renames retain it.
- Add regression coverage for UTF-16 boundaries, CR/LF/CRLF, trailing lines, content replacement, version-only updates, rename, and removal.

| Workload | p50 baseline → candidate (ms) | p50 change (95% CI) | p95 baseline → candidate (ms) | p95 change (95% CI) |
| --- | ---: | ---: | ---: | ---: |
| Uniswap member completion | 0.098958 → 0.048129 | −51.36% [−51.74%, −50.99%] | 0.129650 → 0.059806 | −53.87% [−55.99%, −51.43%] |
| Solady ordinary completion | 0.096827 → 0.050719 | −47.62% [−48.24%, −47.02%] | 0.127067 → 0.064388 | −49.33% [−52.71%, −45.42%] |

PR #1432, opened 2026-09-10
perf(lsp): speed up signature help

0xKarl98:
Towards OSS-738  

Depends on #1431.

Signature help still rebuilds the document's position index and copies the source prefix on every request. It also converts positions for callsites before checking whether their function names match.

- Check the call form and callee name first, so calls to unrelated functions skip the position conversions.
- Reuse the VFS position index introduced in #1431. Completion and signature help share one index for the same document contents.
- Borrow the prefix from the cached contiguous source instead of copying it from the Rope on every request. Content changes replace the document snapshot and its caches together; the existing stale-callee checks still run.
- Add handler benchmarks for 64, 256, and 1,024 callsites, a real router, and its first request after an edit. Add full-profile stdio workloads for an imported library call and the final argument of a v4-core call following a nested call. Regression tests cover warmed requests across receiver edits, changed lines, and UTF-16 positions while analysis is still stale.

The baseline is `afdd3a1c` with #1431 (`9c9dd9134`) applied, so the table measures the additional gains from this PR. Both binaries used the same release build flags on the same macOS arm64 host. Times are warm request latencies over stdio, in milliseconds; lower is better.

| Request | p50 before → after | Change [95% CI] | p95 before → after | Change [95% CI] |
| --- | ---: | ---: | ---: | ---: |
| Uniswap V3 `deploy` signature | 0.315604 → 0.172369 | -45.38% [-47.67, -43.95] | 0.437425 → 0.215419 | -50.75% [-52.74, -48.79] |
| Uniswap V3 `mulDiv` signature | 0.289190 → 0.161867 | -44.03% [-44.47, -43.59] | 0.406015 → 0.200356 | -50.65% [-52.35, -48.98] |
| v4-core `_accountDelta` signature | 0.091179 → 0.061469 | -32.58% [-33.57, -31.51] | 0.118469 → 0.077787 | -34.34% [-38.13, -30.33] |
| Synthetic `double` signature | 0.057411 → 0.058521 | +1.93% [-2.60, +7.92] | 0.071769 → 0.073231 | +2.04% [-3.34, +8.74] |
| Synthetic hover (control) | 0.056000 → 0.057833 | +3.27% [-1.03, +7.41] | 0.070133 → 0.070413 | +0.40% [-5.00, +6.37] |
| Uniswap V3 member completion (control) | 0.049437 → 0.049508 | +0.14% [-0.90, +1.20] | 0.062479 → 0.062427 | -0.08% [-4.16, +4.91] |
| Synthetic import completion (control) | 0.113264 → 0.115098 | +1.62% [-1.20, +4.54] | 0.139965 → 0.139788 | -0.13% [-4.41, +4.30] |

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -9,9 +9,10 @@ use solar_config::CompileOpts;
 use solar_lsp::{
     BenchmarkAnalysis, BenchmarkDocumentUpdate, BenchmarkFoldingRangeRequests,
     BenchmarkOpenDocuments, BenchmarkProject, BenchmarkRepeatedAnalysis, BenchmarkRequest,
-    BenchmarkResponse, BenchmarkSelectionRangeRequests, BenchmarkWorkspaceDiscovery,
-    BenchmarkWorkspacePathQueries, BenchmarkWorkspaceReports, benchmark_folding_ranges,
-    benchmark_folding_ranges_from_rope, benchmark_import_path_at, benchmark_selection_ranges,
+    BenchmarkResponse, BenchmarkSelectionRangeRequests, BenchmarkSignatureHelpRequests,
+    BenchmarkWorkspaceDiscovery, BenchmarkWorkspacePathQueries, BenchmarkWorkspaceReports,
+    benchmark_folding_ranges, benchmark_folding_ranges_from_rope, benchmark_import_path_at,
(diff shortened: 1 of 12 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1435, opened 2026-09-11
perf(lsp): speed up document highlights

0xKarl98:
Towards OSS-738 , 

Document highlights scan every reference in the current file even when the requested symbol has a smaller reverse index.

- Reuse the existing symbol-to-reference index for single-target queries when it is cheaper than scanning the file index.
- Filter indexed references by the requested URI; keep the file-scan fallback for multi-target and common symbols.
- Preserve stable range ordering and declaration `WRITE` precedence.
- Add regression coverage so references from other documents cannot leak into the response.

| Workload | p50 before → after | Change | p95 before → after | Change |
| --- | ---: | ---: | ---: | ---: |
| Optimism (5.38 MB) | 188.76 → 22.93 µs | -87.85% | 264.76 → 31.56 µs | -88.08% |
| Uniswap | 27.92 → 24.31 µs | -12.91% | 34.12 → 30.25 µs | -11.33% |
| Solady | 23.89 → 21.73 µs | -9.04% | 28.82 → 27.03 µs | -6.21% |
| Synthetic small file | 13.92 → 13.85 µs | -0.48% | 20.44 → 17.85 µs | -12.63% |

```text
Warm documentHighlight p50 latency (lower is better)

Optimism  before  ████████████████████ 188.76 µs
          after   ██                    22.93 µs
Uniswap   before  ███                    27.92 µs
          after   ██                     24.31 µs
Solady    before  ███                    23.89 µs
          after   ██                     21.73 µs
Synthetic before  ██                     13.92 µs
          after   ██                     13.85 µs
```

### DIFF
--- a/crates/lsp/src/symbols.rs
+++ b/crates/lsp/src/symbols.rs
@@ -939,15 +939,33 @@ impl SymbolTables {
             .collect::<Vec<_>>();
 
         if let Some(references) = self.file_references.get(uri) {
-            highlights.extend(references.iter().filter_map(|&index| {
-                let reference = &self.references[index];
-                reference.targets.iter().any(|target| targets.contains(target)).then_some(
-                    DocumentHighlight {
+            let target_references = match targets.as_slice() {
+                [target] => Some(self.symbol_references.get(target).map_or(&[][..], Vec::as_slice)),
+                _ => None,
+            };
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

PR #1436, opened 2026-09-11
perf(lsp): cache workspace path index

0xKarl98:
Towards OSS-738  ,  

Workspace path ownership rebuilt the root lookup map for every request, even though workspace configuration changed much less often than requests arrive.

- Cache the immutable import entries and root lookup map together.
- Reuse the cached index for import resolution, source ownership, and flycheck ownership queries.
- Keep the existing cache invalidation points for workspace discovery and root changes.
- Preserve path precedence and tie handling by reusing the same index construction and matching code.
- Assert that cached and uncached benchmark paths return the same result.

The benchmark uses 16 workspaces and measures one workspace ownership query. The cached path avoids rebuilding the root map on every request.

| Benchmark | Before | After | Change |
| --- | ---: | ---: | ---: |
| Workspace path single query | 18.87 µs | 2.01 µs | -89.3% |
| Previous entries-only cache | 5.95 µs | 2.01 µs | -66.2% |

```text
Workspace path query latency (lower is better)

before  ███████████████████  18.87 µs
after   ██                   2.01 µs
```

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -665,6 +665,7 @@ fn workspace_path_queries(c: &mut Criterion) {
     let queries =
         BenchmarkWorkspacePathQueries::new(PATH_INDEX_WORKSPACE_COUNT, PATH_INDEX_QUERY_COUNT);
     assert_ne!(queries.run(), 0);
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

PR #1437, opened 2026-09-11
perf(lsp): cut update and completion overhead

0xKarl98:
Towards OSS-738 , 

Full-document updates rebuilt a Rope even when the text was unchanged, and completion filtering ran Unicode case folding over ASCII names.

- Compare identical full-document updates against the existing Rope's chunks and update the document version directly. The `Optimism.sol` update benchmark drops from 1.257 ms to 0.266 ms (78.8% less time).
- Match ASCII completion names byte by byte, keeping the existing Unicode path for other labels. Selective completion drops from 5.803 µs to 3.508 µs (39.6%); no-match completion drops from 4.082 µs to 2.629 µs (35.6%).
- Reuse the workspace path index from #1436 when building analysis batches. This removes another index rebuild per analysis epoch; the unconfigured workspace still constructs its own index.

The chart compares official main `2c47185cc` with `c016afaf5`, run sequentially in the same checkout on macOS arm64 using the Cargo bench profile: 30 samples, 1 s warmup and 1 s requested measurement time. These are local Criterion update/query timings; error bars show 95% confidence intervals. Empty-prefix completion is included as a control. Analysis-batch cache reuse is not separately timed here.

[Raw samples, estimates and run logs](https://github.com/0xKarl98/solar/tree/d6c14b0777ccf8e625ba32ce94594b20924a74a4).

LSP update and completion benchmark comparison

### DIFF
--- a/crates/lsp/src/config.rs
+++ b/crates/lsp/src/config.rs
@@ -350,7 +350,7 @@ impl Config {
         self.workspace_path_cache = Arc::new(OnceLock::new());
     }
 
-    fn workspace_path_index(&self) -> WorkspacePathIndex<'_> {
+    pub(crate) fn workspace_path_index(&self) -> WorkspacePathIndex<'_> {
         let cache = Arc::clone(
             self.workspace_path_cache
                 .get_or_init(|| Arc::new(WorkspacePathIndex::cache(&self.workspaces))),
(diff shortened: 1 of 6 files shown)


mattsse (approved):
lgtm

PR #1440, opened 2026-09-11
perf(lsp): speed up member and position queries

0xKarl98:
Towards OSS-738 , 

- Index member completion ranges so a query can skip unrelated accesses in the file. Keep inclusive endpoints and the existing ordering for overlapping ranges.
- Use the rope's UTF-16 counts for position conversion. ASCII lines use byte offsets directly, and empty ranges convert the cursor once.
- Add member-access scaling benchmarks and ASCII/Unicode long-line fixtures, with regression coverage for range boundaries and split surrogates.

[Raw samples, confidence intervals and run logs](https://github.com/0xKarl98/solar/tree/4431dcfc446136e76e4ab87b8d02cb0efa9c4718).

LSP query benchmarks

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -214,6 +214,37 @@ fn completion_queries(c: &mut Criterion) {
     group.finish();
 }
 
+fn member_completion_queries(c: &mut Criterion) {
+    let mut group = c.benchmark_group("lsp/member-completion");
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

DaniPopes (approved):

PR #1442, opened 2026-09-11
perf(lsp): index repeated query work

0xKarl98:
Towards OSS-738 ,  

LSP requests that repeatedly inspect the same source now use small indexes instead of rescanning the full collection.

- Signature help indexes call sites by the opening parenthesis's LSP position. Requests retain cached member-call signatures when edits on earlier lines shift byte offsets before reanalysis, while keeping the existing callee and stale-source checks.
- Workspace diagnostic pulls keep the sorted URI list between mutations. Large workspaces merge only stale client IDs; workspaces up to 512 documents retain the simpler path to avoid paying for an extra index.
- Selection ranges order candidate blocks by source offset and track prefix end bounds. Point queries use binary search and stop once earlier blocks cannot contain the cursor.
- The Criterion suite includes a repeated-call signature-help workload and continues to cover workspace diagnostics and open-document selection ranges.

Benchmark evidence from local Criterion runs:

| Workload | Before | After | Change |
| --- | ---: | ---: | ---: |
| Signature help, 1,024 repeated calls | 356 µs | 1.66 µs | 214× faster |
| Workspace diagnostics, 256 documents | 159 µs | 128 µs | 19.5% faster |
| Workspace diagnostics, 4,096 documents | 4.14 ms | 2.38 ms | 42.6% faster |
| Selection range, middle of Optimism source | 5.16 µs | 0.90 µs | 82.6% faster |

LSP query benchmark results


crates/lsp/src/signature_help.rs
```diff
@@ -118,7 +132,8 @@ impl SignatureHelpIndex {
         let cursor = positions.text_range(Range::new(position, position)).start;
         let context = call_context(&source[..cursor])?;
         let call = self.calls.get(uri).and_then(|calls| {
-            calls.iter().find(|call| {
+            let indices = self.calls_by_open.get(uri)?.get(&context.open)?;
```
mattsse: Can we key this by the opening delimiter's LSP position instead? Lengthening a comment on an earlier line preserves a later `target.set(...)` call's line/column coordinates but shifts its byte offset. Before reanalysis completes, this lookup misses the cached call and member-call signature help returns `None` because it has no fallback. The previous implementation converted the cached LSP position against the current text and still returned the signature. I reproduced this with a test that passes with the original implementation and fails here; please cover the earlier-line edit without reanalysis.

mattsse (changes requested):

crates/lsp/src/signature_help.rs
0xKarl98: Have fixed

mattsse (approved):
lgtm

DaniPopes:
btw you can attach images with --attach on gh cli now, so you dont have to upload any files to the repo itself.

DaniPopes (approved):

DaniPopes (approved):

PR #1443, opened 2026-09-11
perf(lsp): cache CodeLens reference counts

0xKarl98:
Towards OSS-738 , 

CodeLens reference counts were recomputed from the full reference index on every request. For immutable analysis snapshots, that repeated work dominated requests with many references.

Changes:

- Cache one reference-count array per source file with `OnceLock`, so repeated requests reuse counts while still invalidating the cache when merged analysis tables are rebuilt.
- Preserve conflict handling and location deduplication, including conflicting source snapshots and references merged from multiple analysis batches.
- Avoid allocating a hash set for declarations with zero or one distinct reference location.
- Pre-size the response vector from the enabled lens kinds to reduce growth reallocations.
- Add scaling benchmarks for repeated references, first requests, and the Unifap V2 pair, plus regression coverage for warmed caches and merge order.

The implementation is independent of PR #1442. It does not change the CodeLens response contents or the behavior of selector and inheritance lenses. `cargo nextest run --locked -p solar-lsp`, Clippy, formatting, and diff checks pass.

Benchmark evidence from the matched 14-case run (20 Criterion samples per case):

| --- | ---: | ---: | ---: |
| CodeLens, 1,024 references | 24.362 us | 0.467 us | -98.08% |
| CodeLens, 16,384 references | 428.423 us | 0.463 us | -99.89% |
| CodeLens, 256 functions | 409.559 us | 345.828 us | -15.56% |
| CodeLens, Unifap V2 Pair | 19.750 us | 13.177 us | -33.28% |
| First request, 1,024 references | 24.589 us | 25.602 us | +4.12% |
| First request, 16,384 references | 482.326 us | 517.647 us | +7.32% |

### DIFF
--- a/crates/lsp/README.md
+++ b/crates/lsp/README.md
@@ -57,6 +57,10 @@ The benchmark groups intentionally keep separate timing boundaries:
 - `project-analysis` and `project-analysis-after-edit` measure compiler and symbol-table rebuilds.
 - `project-edit-application` measures UTF-16 document edit application without analysis.
 - `symbol-table-queries` measures synchronous query kernels, not complete LSP request latency.
+- `code-lens` measures repeated queries, including response destruction. The separate
+  `code-lens-first-request` group clones an unqueried analysis snapshot outside timing and
+  includes the first query's reference-count initialization; snapshot and response destruction
+  stay outside timing. Compare each group's results against its own baseline.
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1445, opened 2026-09-11
perf(lsp): cache diagnostic fingerprints

0xKarl98:
Towards OSS-738 ,

## Summary

Compiler diagnostics currently recompute the complete source fingerprint for every diagnostic produced from a file. On large files with many diagnostics, this makes conversion cost grow with both source size and diagnostic count.

- Cache each file's fingerprint for the duration of one analysis snapshot, keyed by its `SourceFile` start position.
- Reuse the cached value while constructing diagnostic code-action data; the LSP payload and fingerprint format remain unchanged.
- Keep the cache local to the conversion pass so it cannot outlive the source map that produced it.
- Add a separate Criterion target for repeated diagnostic conversion and include the measured result as a chart.

## Benchmark

| Diagnostics in one file | Before | After | Change |
| ---: | ---: | ---: | ---: |
| 1 | 9.33 ms | 9.38 ms | ~0% |
| 16 | 98.26 ms | 9.34 ms | 90.5% faster |
| 64 | 381.37 ms | 9.32 ms | 97.6% faster |


mattsse (approved):
lgtm

benchmarks/lsp-diagnostic-fingerprint.png
DaniPopes, line 1: please remove

0xKarl98: removed

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1448, opened 2026-09-12
fix: exclude runtime IR from creation dumps

DaniPopes:
`-Zdump=evm-ir` now prints only deployment EVM IR, so benchmark `creation.evmir` artifacts no longer repeat the runtime module. Runtime IR remains available through `-Zdump=evm-ir-runtime`. Update UI tests and snapshots to request runtime output explicitly where needed.


mattsse (approved):
lgtm

PR #1450, opened 2026-09-12
perf(lsp): reuse validated aggregate indexes

0xKarl98:
Towards OSS-738 , 

- Reuse the combined workspace index when every batch still matches its source inputs and filesystem dependencies. This keeps initialized query indexes alive across analysis epochs instead of cloning each batch and rebuilding the aggregate.

- Refresh document versions on reuse, including closed overlays, while keeping the existing rebuild path for changed sources, missing imports, dependency changes, and workspaces that become empty.

- Add lifecycle benchmarks that separate initial indexing from incremental first requests. 

Incremental LSP first-request latency


mattsse (approved):
lgtm

PR #1452, opened 2026-09-12
perf(lsp): narrow lexical scope lookups

0xKarl98:
Towards OSS-738 

`scope_at_position` walked every lexical scope in a file and recomputed scope depth for each candidate. Completion and signature-help requests hit this path repeatedly on the same source position.

- Use the existing start-sorted scope list with `partition_point` to find the last scope that starts before the cursor.
- Walk parent links from that scope until finding the innermost containing range.
- Keep the parent check for scopes sharing a start position so nested constructs retain the existing result.
- Remove the per-query depth walk; lexical scope construction already gives us the parent chain needed for lookup.

Benchmark results from the LSP completion group:

| --- | ---: | ---: | ---: |
| selective prefix | 3.45–3.54 µs | 2.75–2.84 µs | −22.7% |
| no-match prefix | 2.88–2.99 µs | 1.80–1.92 µs | −37.5% |

### DIFF
--- a/crates/lsp/src/symbols.rs
+++ b/crates/lsp/src/symbols.rs
@@ -25,6 +25,7 @@ use solar_sema::{
     ty::{CallableParamSource, Ty, TyKind},
 };
 use std::{
+    cmp::Reverse,
     fmt::Write as _,
     ops::ControlFlow,
     path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


crates/lsp/src/symbols.rs
```diff
@@ -1563,15 +1563,27 @@ impl SymbolTables {
...
-            })
+        let scopes = self.file_scopes.get(uri)?;
+        let mut index =
+            scopes.partition_point(|&scope_id| self.scopes[scope_id].range.start <= position);
+        while index > 0 {
+            index -= 1;
+            let mut scope_id = scopes[index];
+            loop {
+                if proto::range_contains(self.scopes[scope_id].range, position) {
+                    return Some(scope_id);
```
mattsse, line 1575: This can return the file scope instead of the contract scope when both start at byte zero. With `contract C { uint stateValue; uint other = stateValue; }` followed by a trailing comment, the file range ends later, so the ascending end-position sort puts it after the contract. This return then skips the contract entirely, and completion at `state|Value` produces no suggestions. I checked a focused completion test: it passes on the base and fails on this head. Sorting equal-start scopes by descending end position should put the smallest range last while preserving construction order for identical ranges. Please add coverage with trailing content; the fixture helper strips the final newline, which otherwise hides this case.

mattsse (changes requested):

crates/lsp/src/symbols.rs
0xKarl98: Have fixed this case

mattsse (approved):
lgtm

PR #1453, opened 2026-09-12
perf(lsp): accelerate repeated query context

0xKarl98:
Towards OSS-738 , 

Signature help now reuses the statement boundary for repeated requests at the same position in an unchanged document snapshot. This avoids rescanning strings, comments, and earlier statements while retaining the existing lexer path for cold requests and edits.

Selection range lookup now uses subtree end bounds to skip blocks that cannot contain the cursor. The range construction and ordering behavior are unchanged, with tests covering broad outer ranges, equal starts, and boundary positions.

Performance improvement

```text
Repeated signature help     21.95% faster  |█████████████████████|
Unchanged router signature  20.90% faster  |████████████████████ |
Selection range queries       3–8% faster  |███–████             |
```


mattsse (approved):
lgtm

PR #1454, opened 2026-09-12
perf(lsp): reduce folding-range latency

0xKarl98:
Towards OSS-738 , 
Speed up rebuilding folding ranges after document edits.

- Reuse the parser's contiguous source and Unicode metadata for line and UTF-16 positions, avoiding repeated rope lookups and extra source copies.
- Check spans directly against the parsed file and group imports using source slices.
- Scan only comment and string delimiters after a successful parse. The existing lexer still determines their boundaries, and incomplete input keeps the full recovery scan.

### DIFF
--- a/crates/lsp/src/folding_range.rs
+++ b/crates/lsp/src/folding_range.rs
@@ -1,13 +1,12 @@
 //! Syntax-based folding-range construction.
 
-use crate::proto;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

PR #1457, opened 2026-09-13
perf(lsp): defer workspace search indexing

0xKarl98:
Towards oss-738

Reduce LSP request and analysis cost by delaying expensive workspace-symbol indexing until a large workspace actually receives a symbol query. Completion responses now sort compact `(label, symbol)` candidates before materializing full `CompletionItem` values, and workspace-symbol queries use a packed lowercase corpus with one-time `OnceLock` initialization and invalidation after index rebuilds. Small workspaces retain the direct scan path.

The broad completion gain comes from avoiding construction of candidates that sorting and deduplication discard. Workspace-symbol queries scan contiguous lowercase data, while `OnceLock` prevents that corpus from being built during analyses that never request workspace symbols. The existing LSP test suite and lint checks pass.

### DIFF
--- a/crates/lsp/src/symbols.rs
+++ b/crates/lsp/src/symbols.rs
@@ -29,7 +29,7 @@ use std::{
     fmt::Write as _,
     ops::ControlFlow,
     path::{Path, PathBuf},
-    sync::Arc,
+    sync::{Arc, OnceLock},
 };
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

PR #1458, opened 2026-09-13
perf(lsp): reuse rename position indexes

0xKarl98:
Rename validation rebuilt the source line-start index once for every occurrence. This change groups the existing URI-sorted locations and reuses one `LspPositionIndex` per affected document, preserving source, UTF-16, version, occurrence, and `WorkspaceEdit` behavior while keeping peak index storage bounded to one document.


mattsse (approved):
lgtm

PR #1459, opened 2026-09-13
perf(lsp): batch descending document edits

0xKarl98:
Towards OSS-738

Applying a rename or Replace All can send hundreds of edits in one `didChange` notification. We currently rebuild the whole document's line-position index for every edit, blocking notification handling before analysis even starts.

- **Scan the document once per batch.** For separate edits ordered from the end of the file toward the beginning, resolve all UTF-16 positions using one index. Later edits leave earlier positions intact, so rebuilding the index adds no useful work.
- **Preserve sequential editing behavior.** Check ordering before building the index, require exact position conversions, and use the existing path for overlapping, adjacent, or otherwise unsuitable edits. Regression coverage includes oversized columns, Unicode, mixed line endings, and multiline replacements.
- **Measure realistic replacement batches.** Add Optimism and Uniswap identifier replacements, alongside single-edit and small-file controls.

The large gain comes from removing repeated full-file scans: a 377-edit batch in the 5.4 MB Optimism fixture needs one scan instead of 377.

| Edit-application workload | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Optimism, 377 replacements | 497.81 ms | 1.72 ms | 99.65% |
| Uniswap TickMath, 20 replacements | 384.18 µs | 30.24 µs | 92.13% |

### DIFF
--- a/crates/lsp/benches/lsp.rs
+++ b/crates/lsp/benches/lsp.rs
@@ -4,16 +4,20 @@
 
 use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 use crop::Rope;
-use lsp_types::{GotoDefinitionResponse, HoverContents, OneOf, Position, Url};
+use lsp_types::{
+    GotoDefinitionResponse, HoverContents, OneOf, Position, Range, TextDocumentContentChangeEvent,
+    Url,
+};
 use solar_config::CompileOpts;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1460, opened 2026-09-13
perf(lsp): prioritize interactive analysis

0xKarl98:
Towards oss-738

1. **Product optimizations:** Faster hover, navigation, reference lookup, and rename after edits, while still waiting for fresh results. Background typing keeps its debounce, and analysis remains limited to one worker.
2. **Why it helps:** Interactive requests wake queued analysis immediately, removing up to 250 ms of repeated waiting. Across 48 local scripted sessions with real source edits, startup and typing pauses, total time fell **33.4–52.0%** with matching responses.

The tradeoff is extra analysis during continuous overlapping edits and hover requests: **54–153% more CPU**, with **13–24% shorter sessions** in that stress workload.


mattsse (approved):
lgtm

PR #1461, opened 2026-09-13
perf(lsp): reduce repeated analysis work

0xKarl98:
Towards OSS-738 , 

Frequent edits repeat semantic analysis and rebuild LSP indexes. Normalize overload parameters once per declaration within each duplicate-check group, reuse per-file declaration order for workspace symbols, and compare signature ranges before longer URIs. This removes repeated conversion and sorting while preserving diagnostic and symbol ordering, complementing the scheduling change in #1460.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1465, opened 2026-09-13
perf(lsp): reuse parsing across quick fixes

0xKarl98:
Towards OSS-738 , 

Reuse one parsed AST and UTF-16 position index across quick fixes within each request, avoiding repeated parsing and indexing. Include diagnostic and quick-fix benchmarks in CodSpeed.

Local release build on macOS arm64, compared with `2849c26`. Full-document request latency, measured as the median of session p50s:

| Fixes per request | Before | After | Latency reduction |
| --- | ---: | ---: | ---: |
| 64 | 2.75 ms | 0.40 ms | 85.6% |
| 256 | 28.20 ms | 1.68 ms | 94.0% |


mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1466, opened 2026-09-14
perf(lsp): reduce rename request overhead

0xKarl98:
Towards OSS-738 , 

- Reuse ordered occurrence indices and group snapshot lookups, validation, and edit construction by file.
- Share rope slices for same-line UTF-16 endpoints while preserving range and document version checks.


mattsse (approved):
lgtm

PR #1467, opened 2026-09-14
perf(lsp): speed up indexing and locations

0xKarl98:
LSP indexing spent substantial time repeating work for the same source locations and inherited contracts. This change snapshots source files with sparse UTF-16 prefix adjustments and recent file/line locality, then reuses the compiler's resolved source scopes for NatSpec names instead of recursively enumerating every import path.


mattsse (approved):
lgtm

DaniPopes:
i dont like the redoing of the sourcemap

PR #1470, opened 2026-09-14
perf(lsp): reuse signature statement boundaries

0xKarl98:
Towards OSS-738
Signature help scans the text before the cursor to find the previous statement boundary. The existing cache only remembers the first cursor position, so moving between arguments repeats that scan.

This PR caches statement boundaries for each unchanged document. Requests within the indexed region use binary search, while requests farther ahead extend the index by scanning only the remaining text. This reduces repeated scanning as the cursor moves, especially in large files.

The index is built only when the cursor moves away from the first cached position, so a single request after an edit avoids its construction cost.


mattsse (approved):
lgtm

PR #1471, opened 2026-09-14
perf(lsp): reuse rope indexes for document edits

0xKarl98:
- Reuse Rope’s line and UTF-16 indexes for single edits and large-document batches, avoiding a full line-offset vector for LF/CRLF documents.
- Unchanged local benchmarks show single-edit time down 94.9% on Optimism, 88.2% on Uniswap, and 79.5% on Unifap. Optimism’s 377-edit batch takes 73.9% less time.
- Warm completion measured 4.8% slower (~1.7 µs); signature help was effectively unchanged in the control confirmation.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1473, opened 2026-09-14
perf(lsp): reduce completion and location overhead

0xKarl98:
Towards OSS-738

Completion filtering skips unnecessary subsequence scans for ASCII names and avoids rechecking the prefix for every candidate. Local fuzzy, missing-name and overlong-prefix benchmarks take about 9%, 13% and 70% less time.

Location conversion snapshots loaded files once per analysis, removing three read-lock acquisitions, two temporary Arc clones and one binary search per span. Analysis timings remain mixed, with a workspace-edit control averaging about 3% slower. Benchmark logic is unchanged.


mattsse (approved):
lgtm

crates/lsp/src/proto.rs
```diff
@@ -432,18 +432,18 @@ pub(crate) fn benchmark_diagnostic_conversion(
...
 pub(crate) struct LocationConverter {
-    source_map: Arc<SourceMap>,
+    files: Vec<Arc<SourceFile>>,
     uris: FxHashMap<BytePos, lsp_types::Url>,
 }
 
 impl LocationConverter {
     pub(crate) fn new(source_map: Arc<SourceMap>) -> Self {
-        let files = source_map.files();
+        let files = source_map.files().to_vec();
```
DaniPopes, line 446: why?

0xKarl98: We’re not copying the source contents here—only cloning the Arc handles into an owned snapshot. SourceMap::files() holds a read lock, so this releases the lock immediately and lets LocationConverter reuse the sorted file list without two locked lookups per span. The upfront cost is amortized because the converter serves many location conversions.

DaniPopes (approved):

PR #1474, opened 2026-09-14
perf(lsp): reduce signature-help request overhead

0xKarl98:
Towards OSS-738

Signature help was re-tokenizing argument text and allocating temporary buffers on each request. This trims that work:

- Reuse the tokens collected while finding the call to identify the active positional or named argument.
- Borrow the current callee text from the document snapshot instead of copying it into a temporary string.
- Check both endpoints on the same line with one rope-line lookup, and use direct offsets for ASCII positions and parameter labels. Non-ASCII text keeps the UTF-16 boundary checks.
- Skip signature deduplication when there are fewer than two candidates.

This also adds Unifap call-hierarchy benchmarks covering prepare, incoming and outgoing calls, expansion, and the first request after analysis. They check callable identities and grouped source ranges, including agreement between first and repeated requests.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1475, opened 2026-09-14
fix(lsp): complete symbols in incomplete code

DaniPopes:
Complete visible symbols, import aliases, and namespaces, including their documentation where available. Member completion uses compiler resolution types for every declaration kind and reads inherited declarations from the compiler's existing scopes. Resolve members in the cursor's source and contract scope so visibility and `using for` attachments follow the use site, including while analysis is pending.

Keep dot completion restricted to receiver members instead of globals such as `abi`, including when local names shadow builtins. Preserve completion for trailing dots and partial member names without semicolons or closing braces. In incomplete-input recovery mode, skip malformed statements at semicolons and continue parsing so nearby errors do not discard later completion context.

NatSpec references are handled separately in #1486.


mattsse (approved):

mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1477, opened 2026-09-14
fix(lsp): share the analysis thread pool

DaniPopes:
Each uncached LSP analysis batch creates a fresh worker pool. Share one server-owned Rayon pool across revisions and workspace batches while keeping the normal parallel worker count. Bind fresh session globals on every worker for one analysis at a time, clear them on return or panic, and shut down workers when the last pool owner drops. Source maps, interners, and diagnostics remain separate for each revision.

Give LSP analysis microbenchmarks a reusable one-worker pool to limit scheduling noise. Pin session microbenchmarks to two workers and exclude non-reentrant dispatch from CodSpeed, while retaining it for wall-time runs.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

PR #1478, opened 2026-09-14
perf(lsp): speed up selection and type hierarchy

0xKarl98:
Towards OSS-738
- Validate selection spans against the parsed source file, removing repeated source-map lookups and rope traversals. Cold expand-selection on Optimism drops from 47.93 to 31.57 ms (34%); warm requests are effectively unchanged.
- Skip rebuilding type-hierarchy responses when references have identical targets. This removes duplicate allocation and serialization during the agreement check; Unifap reference preparation takes 15–17% less time.
- Reuse the request’s parsed URI and compare echoed JSON fields directly, avoiding URI reparsing and temporary JSON allocation. Expanding Unifap Pair’s three supertypes drops from 1.066 to 0.712 µs (33%), with the same strict item validation.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1482, opened 2026-09-15
perf(lsp): speed up call hierarchy expansion

0xKarl98:
Towards OSS-738
- Order callers and callees when building the index, so expanding the hierarchy no longer sorts them on every request.

- Store call relations in flat lists keyed by symbol ID. This removes nested maps and repeated cloning and hashing of URI-based keys.

- Reserve space for each distinct caller or callee up front to avoid growing the result vector while filling it.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1483, opened 2026-09-15
perf(lsp): avoid scanning symbols for rename

0xKarl98:
Towards OSS-738 ,  

Rename currently scans every renamable symbol to find related declarations, even for a helper with no overrides.

- Index override-family members during analysis so rename can look up the group directly.
- Borrow the cached group instead of allocating a temporary list. Symbols with no related declarations need no cache entry or list allocation.


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1484, opened 2026-09-15
perf(lsp): speed up completion after edits

0xKarl98:
Completion after an edit rebuilt a full document line index and materialized contiguous source before the handler knew whether the cursor could complete an import. This made the first request after a large edit pay for work that ordinary member completion does not need.

This change resolves a single completion cursor through the rope's existing line tree, keeps explicit indexing for standalone CR, and checks rope chunks for a possible import string before materializing source. Requests that need repeated conversions promote the same cached position object to a dense line index. It also adds production-handler benchmarks and regressions for Unicode, CRLF chunk boundaries, invalid cursors, edits, and incomplete imports.

On the preserved local workload, first completion after editing an 8,192-access file went from 68.573 µs to 6.270 µs, and the Unifap router case went from 3.025 µs to 1.919 µs. Warm completion on the large fixture increased from 1.516 µs to 1.696 µs; this tradeoff is limited to repeated requests where the previous dense index is already available. The measurements are local and include the pre-existing local LSP baseline.


mattsse (approved):
lgtm

crates/lsp/src/import_resolution.rs
```diff
@@ -74,23 +81,52 @@ pub(crate) fn import_path_at_for_completion(source: &str, cursor: usize) -> Opti
...
+
 /// Rejects code lines that cannot contain a completable import string.
-fn may_complete_string(source: &str, cursor: usize) -> bool {
-    let prefix = &source.as_bytes()[..cursor];
-    let line_break = memchr::memrchr2(b'\r', b'\n', prefix);
-    let line_start = line_break.map_or(0, |offset| offset + 1);
-    if matches!(source.as_bytes().get(cursor), Some(b'\'' | b'"'))
-        || memchr::memchr2(b'\'', b'"', &prefix[line_start..]).is_some()
-    {
+fn may_complete_string<'a>(
```
DaniPopes, line 96: why does this not use the solar Cursor

0xKarl98: This prefilter scans Rope chunks directly, avoiding source copying and tokenization for ordinary completions. Using the current Cursor here would require contiguous text and tokenization before we could skip that work

PR #1485, opened 2026-09-15
perf(lsp): reuse identical workspace analysis

0xKarl98:
Towards OSS-738 , 

Repeated analysis of an unchanged single workspace was rebuilding the full index even when an edit was reverted or a document was reopened with identical text. This keeps the existing aggregate and validates its filesystem inputs before reusing it.

- Reuse the cached symbol tables after replaying recorded disk reads and resolver probes, so changed imports, missing files, and retargeted symlinks still invalidate correctly.
- Refresh open-document versions on reuse, covering overlay open/close transitions without rerunning semantic analysis.
- Avoid the second symbol-table copy used by multi-workspace caching while retaining the existing invalidation behavior.
- Add regression coverage for equal-length dependency edits, missing imports, symlink retargeting, reverted overlays, and version refreshes.
- Add cold, changed, unchanged, reverted-edit, and reopen benchmarks for generated callers and the Unifap router import closure.


DaniPopes:
please fix all bench names in codspeed

mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1488, opened 2026-09-15
perf(lsp): filter quick-fix diagnostics early

0xKarl98:
Towards OSS-738  , 

Cursor quick fixes were copying every diagnostic in the file before checking the requested range.

- Filter stored diagnostics before copying their JSON suggestions.
- Skip byte-offset conversion for unrelated ranges, keeping exact position and edit validation for matching diagnostics.
- Preserve cursor boundaries, diagnostic order, and server-owned fix data, with regression coverage for Unicode and line endings.

Unchanged production-handler benchmarks show 37-39% less time with 64 warnings and 50-52% less with 256. One-warning controls were 2-14% slower; larger whole-document controls were 1-7% slower across two local runs. Benchmark logic is unchanged.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1492, opened 2026-09-15
perf(lsp): cache URI path conversions

0xKarl98:
Towards OSS-738 , 

Repeated LSP requests decoded the same URI and normalized the same path before each VFS lookup. This change memoizes the most recent `Url` to `VfsPath` conversion in `GlobalState`, then reuses it for signature-help and completion requests while keeping document lookup uncached so edits, closes, reopens, and file moves observe the current VFS snapshot.


mattsse (approved):
lgtm

mattsse (approved):
lgtm

DaniPopes (approved):

PR #1493, opened 2026-09-15
perf(mir): simplify select and base-two EXP

DaniPopes:
Stacked on #1444, above #1373, #1489, and #1392. Add costed scalar Select recipes to the existing MIR word-sequence pass for zero, all-ones, and constant arms. General constants, including one, run only under gas optimization. Also simplify `exp(2, x)` to `shl(x, 1)` where SHL is available. The parent already handles small pure branches and power-of-two multiplication after memory lowering. The stack scheduler and optional backends in #1481 are unchanged.

Preserve zero/nonzero conditions without trusting nominal bool types. A dirty word of two previously returned signature `0x6a0ca452` instead of `0x26121ff0`. Required EVM shaping now inverts unknown conditions and swaps Select arms, requiring one ISZERO; proved canonical values need none. The cost model includes that lowering. MIR and final EVM IR FileChecks cover the recipes, normalization, fork guards, and size-mode exclusions; runtime regressions cover dirty callers, exponent boundaries, and checked overflow.

The new dynamic EXP workloads save 3,598–5,248 gas per 64-iteration call under both gas and size optimization, with unchanged bytecode lengths. These are synthetic workload gains. The condition-inversion follow-up independently saves three runtime and creation bytes across 843 common size-mode UI fixtures, with unchanged gas-mode bytecode and no size regressions. The original Select correctness fix makes one fixture 49 bytes larger than #1444; that tradeoff remains.


mattsse (approved):
lgtm

PR #1496, opened 2026-09-16
fix(lsp): reject invalidated analysis requests

DaniPopes:
An analysis can finish, wake a pending request, and then become stale when another edit arrives before the request resumes. Keep each waiting request tied to its input epoch and return `ContentModified` when processed changes invalidate it, even if newer analysis has already finished. An invalidation notification wakes pending requests without waiting for cancelled workers to publish. This also prevents rename from applying its original cursor position to a different identifier after edits or disk changes.

Document and workspace diagnostic pulls translate invalidation to `ServerCancelled`, whose [protocol default asks the client to retry](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.17/language/pullDiagnostics.md#L310-L322). This follows the snapshot-cancellation approach used by [rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/fa88768e772857f332bc8383e3a1c5a4212f9a6e/crates/ide/src/lib.rs#L957-L969). Invalidation is conservative across the workspace; content-identical edits preserve pending requests.

Regressions cover publication and invalidation ordering, wakeups without publication, disk-backed rename target changes, diagnostic retry, and 200 rapid edit/save cycles in a synthetic 257-file workspace. VS Code computer-use reproduction remains open because this environment has no editor, display server, or computer-use tool; the stress coverage uses the existing LSP handler harness.


DaniPopes:
cyclops audit fast

PR #1505, opened 2026-09-17
refactor(mir): enforce explicit type contracts

DaniPopes:
Enforce exact MIR operand, result, aggregate, call, and terminator types, with full type and SSA checks at phase and backend boundaries. Generated integers use `i1`, `i160`, and `i256`; `memptr`, structs, slices, and memory-object references retain their types. Booleans are zero or one, addresses fit in 160 bits, and branches require `i1`. Preserve raw source bits as `i256` where assembly can observe them, including immutable storage.

Use explicit LLVM integer and pointer conversions and canonical equality/inequality zero tests, leaving `ISZERO` to EVM IR. Other `iN` widths remain syntax-only. Keep cast folding, e-graph rewrites, lowering, and proofs consistent with these contracts. ABI and call-site proofs remove redundant cleanup without changing SSA types.

Pointer casts establish no allocation ownership. Memory optimizations require bounded, disjoint ranges and preserve observable memory expansion; return cleanup requires proven initialization, layout, and alias safety. Narrow immutable encodings only when every assignment preserves all stored bits.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1512, opened 2026-09-17
refactor(mir): canonicalize ISLE constant operands

DaniPopes:
Keep constants on the right throughout e-graph matching, stored nodes, and emitted MIR, reversing comparison predicates when operands swap. Remove redundant mirror rules and add scalar reductions for selects and bit operations. Group ISLE sources by MIR, backend selection, and EVM IR, with matching build inputs, generated preludes, and proof paths.

Select branch conditions by zero/nonzero sense after layout chooses the taken edge. Share the conditional-jump emitter across branches and switches, and fold comparison inversions across independent operand computations while preserving bounds, stack peaks, effects, and metadata guards. Keep cold revert branches cheap to invert and price switch selection from the emitted miss tests.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1514, opened 2026-09-17
perf(codegen): add loop and value cleanup passes

DaniPopes:
Add bounded loop unswitching, closed-form loop recurrence folding, constant-memory hash folding, pure-expression sinking, and dependency-based EVM word cleanup. These passes remove repeated work and redundant masks while preserving typed MIR, memory effects, and the scheduled EVM stack layout. Hash folds must recover their added deployment cost over the configured optimizer runs.

Gas optimization does not preserve exact out-of-gas outcomes: a finite, pure loop can finish after folding even when executing every iteration would exhaust gas. Explicit arithmetic failures remain, and emitted storage writes still obey the EVM gas sentry. The rewrite guide documents this contract.

Build on the scalar improvements merged in #1523. Include pass-specific regression coverage and the compiler comparison prompted by the [backend investigation](https://gist.github.com/DaniPopes/1a6b03da14e5805fead5f2490fd65573).


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

legion2002 (commented):
:eye: **Cyclops Review**

Recovered review for `58b4912ab9edd19e2d66e7e931badd7584f4ae2c` from the four saved worker reports in failed run `cyclops-pr-review-kx2k2`. The workers reported no concrete miscompile. One low-severity defense-in-depth candidate survives consolidation; the inline comment covers its opcode-width contract recommendation, with the related raw-stack-op hardening callout below. This is a partial-run recovery, not a completed audit or a verified vulnerability. Earlier review comments are not repeated.

<details>
<summary>Reviewer Callouts</summary>

- **Raw stack-op fallback (`word_cleanup.rs`, `Graph::run`):** Raw `DUPN`/`SWAPN`/`EXCHANGE` instructions without a logical `StackOp` fall through to the opcode table's generic stack effects. Those effects do not model duplication/permutation dependencies. Compiler-generated Solidity/Yul uses the logical stack-op representation, and the IR verifier rejects these raw forms, so the report found no source-reachable miscompile. As hardening, barrier on these raw opcodes before generic `operation()` handling; preserve the existing precise handling of logical stack ops.
- **Lowering-pipeline coverage:** The reports flag the new post-physical-lowering `MemoryDse` invocation and lowering-pipeline `LoopCanonicalize` invocations for targeted review. In particular, cover materialized free-memory-pointer stores, deferred allocations, and memory read by external-call argument windows. No defect was reproduced in these paths.
- **Validation boundary:** Saved reports describe passing compiler unit tests and differential checks, but the UI suite could not complete FileCheck comparisons because FileCheck was unavailable. These are the workers' recorded results; tests were not rerun during recovery. Consolidation failed, and planned worker iterations were lost.

</details>

crates/codegen/src/backend/evm/ir/passes/word_cleanup.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut masks = self.masks.iter_mut().rev().peekable();
+        for id in (0..self.nodes.len()).rev() {
+            let node = &self.nodes[id];
+            let demanded = node.demanded;
+            let opcode = node.opcode;
+            let mut removed = false;
+            if let Some(mask) = masks.peek_mut()
+                && mask.node == id
+            {
+                removed = demanded <= mask.bits || self.nodes[node.inputs[1]].bits <= mask.bits;
```
legion2002, line 192: 🛡️ **[DEFENSE-IN-DEPTH]** Protect the opcode-width contract used to delete cleanup masks

**Severity: Low — latent/future-change risk; no current miscompile reproduced.**

Mask removal here trusts producer widths from `op::definition().result_bits`, while the backward walk trusts `input_bits` to decide which high bits consumers ignore. An understated annotation can therefore remove a required `PUSH <mask>; AND` pair. The archived worker checked the currently annotated rows and found them sound, but the opcode-table test checks numeric bounds and a snapshot rather than the semantic width guarantees.

**Recommended Fix:** Document that these annotations must conservatively bound result widths and observed input bits for every supported fork, and add semantic checks to `scripts/evm-rules/` or equivalent differential regression tests. Include boundary cases such as `CLZ(0) == 256` (nine result bits), address truncation, and `MSTORE8` truncation. This would make future annotation changes exercise the assumption that now drives instruction deletion.

Recovered from `pr-1514-w2_iter_2_report.md`; canonical candidate `rc-f41f91e290c0bc19` was recorded as `not-required` for verification, not as a verified bug.

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1525, opened 2026-09-21
feat(mir): optimize native Solidity integers

DaniPopes:
Solidity integers now retain their widths in MIR: `uint128` and `int128` become `i128`, with signedness carried by operations. MIR supports `i1` and byte widths from `i8` through `i256`, requires explicit conversions, and keeps every integer clean above its declared width. Source bindings whose raw bits cross inline assembly use explicit `i256` carriers.

Constant folding, shared scalar rewrites, range analysis, and loop optimizations account for native wrapping and sign bits. The `lower-integers` pass materializes masks and sign extensions inside MIR, where scalar optimization removes redundant conversions before EVM IR. Division by zero and oversized shifts retain EVM semantics; checked Solidity arithmetic retains its panic behavior.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1526, opened 2026-09-21
fix(codegen): handle Sourcify compilation failures

DaniPopes:
Sourcify contracts exposed failures in storage-reference lowering, ABI pointer tracking, stack scheduling, and loop splitting. This fixes those paths, uses full-width storage array layouts and target-specific stack reach, and extends existing MIR and runtime UI regressions. Heap-helper proofs account for caller memory resets, oversized aggregate copies report unsupported materialization, and discarded library method references preserve receiver effects. The corpus default now matches the tracked Solidity 0.8.37 release.

The Sourcify campaign covered 173,545 inputs through Solidity 0.8.34 with IR verification enabled. Front-end fixes for tuple initializer scope and abstract interface IDs are in #1537, which also targets main independently.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1529, opened 2026-09-21
perf(codegen): refine pass scheduling

DaniPopes:
Memory lowering and late EVM CFG rewrites expose simplifications after their cleanup passes have run. Repeat SCCP/CFG cleanup after memory lowering, share exposed terminal tails before constant packing, and finish with block CSE and peepholes. Refresh final layout for the size objective while preserving gas-mode loop layout. Terminal sharing skips modules containing `PC` to preserve distinct instruction-position observations. Final CFG cleanup follows all sharing, and sharing retains custom instruction stack contracts. All optimization passes already exist.

| Measurement | Gas objective | Size objective |
| --- | ---: | ---: |
| Runtime bytes, geometric mean over 23 cases | 🟢 -0.037% | 🟢 -0.076% |
| Runtime gas, geometric mean over 23 cases | unchanged | 🟢 -0.096% |
| Project runtime bytes, total reduction | 🟢 6,801 | 🟢 24,419 |

The project comparison covers 1,171 nonempty runtime objects in gas mode and 938 in size mode, including tests and helpers. Eight gas-mode objects grow by one or two bytes. In size mode, SeaportValidatorHelper grows 37 bytes and two V4 test helpers each grow two bytes. The existing Uniswap syntax failure and forced-size OpenZeppelin forwarding-buffer failure remain excluded.

Matching debug builds take 1.50% longer by geometric mean across the nine projects, using three alternating samples per project; Solmate is the largest increase at 4.20%. This is a compile-time cost for the generated-code improvements.

[The investigation](https://github.com/paradigmxyz/solar/blob/dani/pass-scheduling/docs/PASS_SCHEDULING.md) compares LLVM, solx LLVM, Cranelift, rustc MIR, Vyper Venom, Sonatina/Fe, and solc schedules, and records bounded repetition experiments. Whole-pipeline repetition can keep reporting layout changes after bytecode stabilizes; targeted cleanup gives the retained improvements.


DaniPopes:
cyclops audit

DaniPopes:
cyclops audit

PR #1536, opened 2026-09-22
fix(lsp): reject conflicting point query contexts

0xKarl98:
Towards CYCLOPS-1465 , 

- Return no definition, declaration, type-definition, hover, or highlight result when indexed workspace contexts disagree.
- Compare physical bindings and query results while preserving compatible results and single-context behavior.
- Cover conflicting remappings, inferred types, inherited documentation, and source snapshots in both context orders.


DaniPopes (approved):

DaniPopes (approved):

DaniPopes:
cyclops audit fast

DaniPopes:
cyclops audit fast

